Issue #323, opened 2026-05-06
asyncio event loop wedges silently — accept queue fills, taOS becomes unreachable


jaylfc:
## Symptom

User opens the taOS web UI and gets a grey screen / endless spinner. systemd reports the unit is `active (running)` and uvicorn is alive, but no requests get answered.

## Diagnosis on production Pi (2026-05-06)

Service had been up 3 days. Listening socket was healthy:

```
LISTEN 2049   2048   0.0.0.0:6969   users:(("python",pid=...,fd=48))
```

The `2049/2048` is the kernel's TCP accept queue: connections waiting for the application to call `accept()`. **Full queue == event loop has stopped accepting connections.** New SYNs arrive, get queued, eventually get rejected. Browser SYN never gets `accept()`'d → grey screen.

Last journal entry was 3 days old, then nothing. `/api/cluster/heartbeat` had been working fine until the moment it wasn't. Process never crashed, never logged anything, just stopped servicing the loop. systemd has no idea anything's wrong because the process is technically alive.

Only fix was `systemctl restart tinyagentos.service`.

## Likely cause

A background task (or scheduled coroutine) makes an outbound `httpx` / `aiohttp` call **without a timeout**. When the destination becomes unreachable mid-flight (worker reboots, DNS blip, network change), the call blocks the loop indefinitely. The listen socket queue piles up; eventually the queue is full and connections start failing.

This pattern is consistent with what we know:
- Last journal lines before silence were `Backend health check failed for local-X: Unknown backend type: ''` (a periodic loop)
- POST /api/cluster/heartbeat from worker 192.168.6.108 was the most recent successful op
- After that — nothing for 3 days

## Hot suspects worth auditing

- `backend_adapters.check_backend_health` (the loop printing "Unknown backend type")
- Cluster heartbeat / reverse-poll
- LiteLLM / Ollama / rkllama probes
- Scheduler that reaches out to workers
- Anything calling `httpx.AsyncClient()` without an explicit `timeout=`

## Proposed fix (separate PR)

1. Audit every `httpx.AsyncClient(...)` and `aiohttp.ClientSession(...)` instantiation. Default to `timeout=httpx.Timeout(10.0, connect=5.0)` or similar.
2. Wrap every periodic outbound call in `asyncio.wait_for(...)` with an explicit timeout.
3. Add a watchdog: a 30s-period asyncio task that hits `/api/health` from inside the loop and logs (or restarts) if its own request hangs.
[...]

PR #324, opened 2026-05-06
fix(settings): update-flow hardening — fail-loud + manual Rebuild Frontend

jaylfc:
## Why

Two real failure modes hit my Pi today and would hit any user updating from Settings:

**1. Silent dep install failures.** `POST /api/settings/update` ran `pip install -e .` with stdout+stderr piped to `DEVNULL` and never checked the return code. A failed pip install (network blip, missing system lib for a wheel that needs to build, partial venv state) left no breadcrumbs and the endpoint still returned `status: \"updated\"`. User clicks Restart, service tries to import new code, hits ModuleNotFoundError, crashes — grey screen.

**2. Stale frontend bundle after source-only PRs.** `static/desktop/assets/*` is committed to the repo. When a PR lands source-only (e.g. PR #319 added the Store filter components but didn't include the rebuilt bundle), `git pull` ends up with new source AND old committed bundle touched at roughly the same instant. The mtime-based staleness check in `rebuild_desktop_bundle_if_stale` picks the wrong winner and skips the rebuild — user restarts into old UI without the new components.

Both bit me on the production Pi today.

## What this PR does

**Fail-loud on dep install:**
- Captures pip stdout/stderr into the response. On non-zero exit, returns 500 with the captured output and aborts the update.
- Adds a post-install **import smoke test** in a fresh interpreter that imports `tinyagentos.app` plus the desktop_browser modules pulling in BrowserApp v2 deps (lxml, sqlcipher, argon2, py_vapid, pywebpush, cryptography). Catches the partial-install case where pip exits 0 but a wheel was quietly skipped.
- Both failure modes return structured output (`git_output`, `pip_output`, `import_error`).
- Extracts a small `_run_capture` helper for subprocess calls, with two unit tests.

**Force rebuild on Update + manual button:**
- `apply_update` now calls `rebuild_desktop_bundle_if_stale(..., force=True)`. The cost is one ~30s npm build per Update click; the benefit is the bundle is always current.
- New `POST /api/settings/rebuild-frontend` endpoint that always rebuilds unconditionally.
- Updates section in SettingsApp grows a **Rebuild Frontend** button — manual escape hatch when the UI looks stale.

## Out of scope

The asyncio-loop wedge that caused my Pi's accept queue to fill (independent of update flow) is tracked separately as #323 — different root cause (uncancelled outbound httpx call), separate audit pass.

## Test plan

- [x] `pytest tests/test_routes_settings.py` — 16 passed
- [ ] On Pi: click Install Update — verify rebuild happens, logs visible
- [ ] On Pi: click Rebuild Frontend after a source-only state — verify bundle rebuilt, hard-refresh shows new components
- [ ] On Pi: simulate failed pip install (typo in pyproject.toml) → /api/settings/update should return 500 with pip_output

Refs #323

### DIFF
--- a/desktop/src/apps/SettingsApp.tsx
+++ b/desktop/src/apps/SettingsApp.tsx
@@ -633,6 +633,7 @@ function RestartProgressModal({
 function UpdatesSection() {
   const [checking, setChecking] = useState(false);
   const [applying, setApplying] = useState(false);
+  const [rebuilding, setRebuilding] = useState(false);
   const [info, setInfo] = useState<UpdateInfo | null>(null);
   const [status, setStatus] = useState<string | null>(null);
   const [prefs, setPrefs] = useState<AutoUpdatePrefs>({ check_enabled: true, auto_apply: false, auto_restart: false });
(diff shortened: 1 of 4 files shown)


jaylfc:
@coderabbitai full review

This PR merged before your initial review completed (rate-limit hit on the original push). Could you take another pass now and surface anything that might have slipped through? A manual audit found three potential issues worth your second opinion — see #327. If you see those plus more, even better.

PR #364, opened 2026-05-07
fix(cluster): install-targets dedupe + canonical RAM bucket + unknown-hardware UX

jaylfc:
## What this fixes

Three bugs surfaced by the live `/api/cluster/install-targets` response on the Pi:

```json
[
  { "name": "local", "type": "local", "tier_id": "arm-npu-15gb", "targets": ["cpu"] },
  { "name": "fedora-worker", "type": "remote", "tier_id": "", "targets": ["cpu"] },
  { "name": "local (current)", "type": "remote", "addr": "unix://", "tier_id": "", "targets": ["cpu"] }
]
```

### Bug 1 — duplicate "local (current)" entry (`routes/cluster.py`)

Incus exposes the local daemon as a remote named `"local (current)"` when it's the active context. That string isn't in `_BUILTIN_REMOTES`, so it fell through and was added as a second controller entry. Fix: skip any remote whose `addr` starts with `"unix://"` — the durable signal for the local incus socket regardless of display name.

### Bug 2 — tier_id 15gb doesn't match catalog buckets (`cluster/capabilities.py`)

A 16 GB Pi reports ~15 GB after kernel reservation. `ram_mb // 1024` produced `15`, giving `arm-npu-15gb`. Catalog manifests only declare canonical buckets (2/4/8/16/32/64/128 GB), so the Store filter went empty when the controller NPU was selected. Fix: `_snap_to_bucket()` rounds raw GB **up** to the next canonical bucket. 15 → 16, 17 → 32, 33 → 64, 200 → 128. Applied to all five `gb = max(1, ...)` call sites in `worker_tier_id`.

### Bug 3 — incus-only remotes silently break the Store filter (multi-file)

`fedora-worker` is registered as an incus remote but not as a taOS cluster worker. `worker_tiers.get(name)` returned `""` and `worker_hw` was `{}`, so `hardware_to_targets({})` returned `["cpu"]` and `tier_id` was `""`. Selecting this device in the Store matched nothing and emptied the catalog with no explanation.

Fixes:
- **Backend**: empty-hardware remotes now emit `tier_id: "unknown"` and `hardware_known: false`.
- **DevicePillBar**: chips for `hardware_known=false` devices show a muted `?` badge with a tooltip.
- **Filter banner**: when an unknown-hardware device is selected, a banner appears above the catalog: "Hardware unknown — register it via the Cluster app to filter accurately. Showing all models for now."
- **filter.ts**: unknown-hardware devices don't contribute a tier to the filter set. If all selected devices are unknown, `requireDeviceMatch` is false — the catalog shows everything instead of nothing.

## Tests

- `tests/cluster/test_worker_tier_id.py` — 14 new Python tests: `_snap_to_bucket` boundary cases + `worker_tier_id` Pi regression.
- `filter.test.ts` — 6 new cases for `hasUnknownHardwareDevice` and unknown-device filter passthrough (22 total).
- All 27 Python cluster tests pass. All 34 frontend tests pass. `tsc --noEmit` clean. Build succeeds.

## Out of scope — known follow-up

The controller shows `targets: ["cpu"]` instead of `["rockchip", "cpu"]`. Root cause is in `hardware_to_targets` — the controller hardware dict's `npu.type` may not match `"rk3588"` or `"rknpu"` at runtime. A TODO comment is added near `hardware_to_targets` pointing at this disconnect. Will be addressed in a separate PR.

### DIFF
--- a/desktop/src/apps/StoreApp/DevicePillBar.tsx
+++ b/desktop/src/apps/StoreApp/DevicePillBar.tsx
@@ -3,6 +3,31 @@ import { useMemo } from "react";
 import { X } from "lucide-react";
 import type { InstallTarget } from "./types";
 
+/** Banner rendered above the catalog when one or more selected devices have unknown hardware. */
+export function UnknownHardwareBanner({ devices }: { devices: InstallTarget[] }) {
+  const unknownNames = devices
+    .filter((d) => d.hardware_known === false)
+    .map((d) => d.friendly_name ?? d.label);
+  if (unknownNames.length === 0) return null;
+  const names = unknownNames.join(", ");
+  return (
+    <div
+      role="status"
+      aria-live="polite"
+      className="flex items-center gap-2 rounded-lg border border-amber-500/30 bg-amber-500/10 px-3 py-2 text-xs text-amber-300 mb-2"
+    >
+      <span aria-hidden="true">?</span>
+      <span>
+        Selected {unknownNames.length === 1 ? "device" : "devices"}{" "}
+        <strong>{names}</strong>{" "}
+        {unknownNames.length === 1 ? "has" : "have"} unknown hardware — register{" "}
+        {unknownNames.length === 1 ? "it" : "them"} via the Cluster app to filter
+        accurately. Showing all models for now.
+      </span>
+    </div>
+  );
+}
+
 interface Props {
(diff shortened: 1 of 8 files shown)


jaylfc:
@coderabbitai review

PR #365, opened 2026-05-07
fix(pwa): shell pre-login accessible so SW can actually precache it

jaylfc:
## Bug

The service worker added in #348 precaches `/desktop/`, `/desktop/index.html`, and `/chat-pwa` on install. All three URLs were auth-gated and returned 401 on a fresh visit without a session cookie. With `Promise.allSettled` the SW install didn't abort, but those entries silently never made it into the cache. When the backend goes down after an Install/Update click, the SW has no cached shell to serve, falls through to network, network is down, user sees the browser offline error. The whole point of the SW is defeated.

Observed on the Orange Pi (192.168.6.123) — fresh install sequence always ends with a blank offline screen after the backend restarts.

## Fix

Drop the auth gate on the SPA shell HTML. The SPA already enforces auth client-side: it calls `/auth/status` on boot and redirects to `/auth/login` if there is no valid session. Server-side 401 on the HTML is redundant.

**Stale-bundle trade-off:** the original gate comment cited stale-bundle risk. That risk is now mitigated by `__TAOS_VERSION__`-namespaced caches: the SW's activate step deletes any cache whose name doesn't match the current build token, so stale index.html entries are evicted automatically on every deploy.

### `auth_middleware.py`

- Added `/desktop`, `/desktop/index.html`, `/chat-pwa` to `EXEMPT_PATHS`.
- Widened `EXEMPT_PREFIXES` entry from `/desktop/assets/` → `/desktop/` and from `/chat-pwa/assets/` → `/chat-pwa/` so all SPA subpaths (client-side routes under `/desktop/*` and `/chat-pwa/*`) are reachable pre-login. The `serve_spa` and `serve_chat_pwa_assets` routes both fall back to `index.html` for unknown paths, so those subpaths serve the shell — they need to be exempt for the same reason as the root.
- Updated the misleading comment block to explain the new design.

### Tests (`test_routes_desktop.py`)

Five new tests:
- `/desktop` returns 200 or 404 (never 401) without a cookie.
- `/desktop` returns 404 (not 401) when not built.
- `/chat-pwa` returns 200 or 404 without a cookie.
- `/chat-pwa/<subpath>` returns 200 or 404 without a cookie.
- `/api/agents` still returns 401 without a cookie (sanity: we didn't open everything).

### Docs

- `docs/getting-started.md` — added a callout explaining the install-time auth flow: install works without being logged in, first launch shows the SPA login screen, shell survives backend restarts.
- `docs/design/desktop-shell.md` — expanded the one-line PWA table entry with a follow-up paragraph covering precache behaviour, version-namespaced cache eviction, and the version-mismatch toast.

## Test results

```
62 passed (test_routes_desktop.py + test_routes_auth.py + test_auth.py), 0 failures
```

### DIFF
--- a/docs/design/desktop-shell.md
+++ b/docs/design/desktop-shell.md
@@ -462,6 +462,8 @@ Import templates from a community marketplace or shared repository. This extends
 | API client | fetch / SWR | REST + WebSocket to FastAPI |
 | Icons | Lucide React | Clean, consistent icon set |
 | PWA | Service worker | Offline shell, installable |
+
+The service worker precaches the shell HTML (`/desktop`, `/desktop/index.html`, `/chat-pwa`) and all hashed JS/CSS assets on install. Caches are namespaced with `__TAOS_VERSION__` so the activate step evicts any stale cache from a previous build. When a new build is deployed the SW detects the version mismatch on next load and shows a toast prompting the user to reload — they never silently get stale assets. The shell HTML is intentionally pre-login accessible (no auth gate) so the SW can cache it on first visit; the SPA enforces auth client-side via `/auth/status` and redirects to `/auth/login` when there is no valid session.
 | Backend | FastAPI (unchanged) | Serves SPA bundle as static files |
 
 ## Migration Path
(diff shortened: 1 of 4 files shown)


jaylfc:
@coderabbitai review

Issue #366, opened 2026-05-07
bug(hardware): hardware.json cache prevents re-probe on restart


jaylfc:
Surfaced from discussion #312 by @johny-mnemonic.

The hardware probe writes `data/hardware.json` at first boot and reads it back on every subsequent restart. Once a value is decided ("no Vulkan", a specific tier_id, etc.), restarting taOS alone won't re-run the probe.

Concrete failure: johny installed `vulkan-tools` after the first boot detected no Vulkan. He restarted taOS and `gpu.vulkan` was still `false`, `profile_id` still `x86-cpu-3gb`. The fix would have been a re-probe, but the cache makes that impossible without manual intervention.

## Workaround

`sudo rm /home/$USER/tinyagentos/data/hardware.json` then restart.

## Fix options (pick one)

1. Always re-probe on startup. Cheap (the probe is fast). Removes the cache entirely.
2. Re-probe on startup but compare against cached snapshot; only write/log when changed. Slightly more work but gives observability into hardware drift.
3. Add a "Refresh hardware" button in Settings + System Info that triggers a re-probe on demand. Doesn't fix the silent-stale-cache failure but gives users an out.

Option 1 is the "install must just work" answer. Option 3 should land alongside even if we pick 1 — gives users a self-service recovery path for environments where the probe itself has bugs.

jaylfc:
Closed — fixed in #371 (merged earlier). Hardware re-probes on every controller startup, plus there's a `POST /api/hardware/detect` endpoint that the Settings page can call to force a fresh probe.

---
_Jay is asleep right now so this comment was written by Claude Opus 4.7._

PR #371, opened 2026-05-07
fix(hardware): always re-probe on startup; add /api/system/hardware/refresh (#366)

jaylfc:
## Problem

Hardware profile was cached for 24h and only re-probed on cache miss or expiry. When johny installed `vulkan-tools` and restarted taOS, the stale `hardware.json` (with `vulkan: false`) was served from cache — the wrong tier persisted until he manually deleted `data/hardware.json`.

Bug thread: #312 (johny's case).

## Fix

**Half 1 — always re-probe on startup**

`get_hardware_profile` now calls `detect_hardware()` on every call. The cache file is a fallback only: if the probe raises (broken `/sys`, unsupported platform), the cached value is returned. If the probe raises and no cache exists, the exception propagates. The `CACHE_TTL_SECONDS` constant and its `import time` guard are removed.

The probe itself is ~100ms (a handful of subprocess calls), so always running it on startup has no meaningful cost.

**Half 2 — `POST /api/system/hardware/refresh` endpoint**

Added to `tinyagentos/routes/system.py`. Deletes the cache file, calls `get_hardware_profile`, updates `app.state.hardware_profile`, and returns the fresh profile JSON. Useful between restarts when a user installs new drivers.

Frontend "Refresh hardware" button in the System Info panel is a follow-up.

## Tests

- `test_always_reprobes_on_startup` — stale cache with sentinel `ram_mb=1` is not returned; fresh probe value is returned and written to cache.
- `test_falls_back_to_cache_when_probe_raises` — monkeypatched raise returns cached profile.
- `test_reraises_when_probe_raises_and_no_cache` — monkeypatched raise with no cache re-raises.
- `test_refresh_updates_app_state` — POST refresh, then GET `/api/system` reflects the new values.
- `test_refresh_overwrites_cache_file` — cache file matches fresh probe after endpoint call.

## Files changed

- `tinyagentos/hardware.py` — refactored `get_hardware_profile`, dropped `CACHE_TTL_SECONDS`
- `tinyagentos/routes/system.py` — added `POST /api/system/hardware/refresh`
- `tests/test_hardware.py` — replaced stale cache tests with the four cases above

### DIFF
--- a/tests/test_hardware.py
+++ b/tests/test_hardware.py
@@ -1,8 +1,10 @@
 # tests/test_hardware.py
 import json
 import pytest
-from unittest.mock import patch, mock_open
+from unittest.mock import patch
 from tinyagentos.hardware import detect_hardware, get_hardware_profile, HardwareProfile
+
+import pytest_asyncio
 
 
 class TestDetectHardware:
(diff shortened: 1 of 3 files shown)


jaylfc:
@coderabbitai review

Issue #369, opened 2026-05-07
bug(container): No container backend configured RuntimeError when neither Incus nor Docker present; should auto-install Incus or fail with clear UX


jaylfc:
Surfaced from discussion #312 by @johny-mnemonic on Ubuntu 25.10.

When the container backend (Incus or Docker) isn't installed or detected, components that try to spin up a worker container raise:

\`\`\`
RuntimeError: No container backend configured. Call set_backend() first.
\`\`\`

This shows up as an unhandled error in the journal and probably explains why johny's Fedora-worker registered but had empty hardware (it couldn't be queried via the container backend).

## Fixes (do all)

1. **install-server.sh should install Incus by default** when no container runtime is detected. Incus is lighter than Docker on a home server and is what taOS prefers internally. Skip on platforms where Incus isn't packaged (auto-fall-back to Docker, or the install script flags it explicitly to the user).

2. **Auto-detect at runtime**: when both Incus and Docker are present, prefer Incus (matches install-server.sh's preference). When only Docker is present, set_backend("docker") automatically. When neither is present, fail with a clear message that the user can act on, not a `RuntimeError` in the log.

3. **Better error UX**: instead of a Python traceback in the journal, surface "container backend not configured — install Incus via `sudo apt install incus` and restart" in any UI that depends on container ops (Cluster app, Worker registration flow, anywhere that fails because of this).

This is the same root cause class as #354 (Vulkan detection): we detect a hardware/OS feature is missing but don't either install the missing piece automatically or surface the gap in a way the user can fix.

johny-mnemonic:
I saw some merged commits here so I tried to re-run the install script.
It fails like this now:
```
[server-install] no container runtime found — installing Incus via Zabbly apt repo
E: The repository 'https://pkgs.zabbly.com/incus/stable questing Release' does not have a Release file.
```
Looking at the list of distros, it has only Debian ones, not Ubuntu.

Wouldn't it be better to first try to install Incus from the default repo and only when it is not found, try to add Zabbly apt repo (for compatible distros)?

jaylfc:
Fixed in #394 — install script now tries the default apt repos first and only falls back to Zabbly for codenames Zabbly actually publishes for. Ubuntu 25.10 ships incus in universe so a re-run after this lands should work cleanly. Thanks for catching it.

jaylfc:
[...]

Issue #368, opened 2026-05-07
bug(install): disk quota monitor init failure causes 0 MB free instead of falling back to system stat


jaylfc:
Surfaced from discussion #312 by @johny-mnemonic on Ubuntu 25.10.

When the disk quota monitor fails to initialise the log says:

\`\`\`
disk quota monitor failed to initialise — disk routes will still work
\`\`\`

But that "still work" claim is wrong. The install dispatcher reads from the same broken disk-accounting subsystem and gets back 0 bytes free. Result: every model install fails with `needs XY MB of disk, device has 0 MB free` even when the box has tens of GB free.

## Fix

Two halves:

1. **The fallback is missing**: when the quota monitor fails to initialise, the install dispatcher's free-disk check should fall back to plain `shutil.disk_usage()` against the install dir. Don't return 0; return the OS-level free space.

2. **The log message is wrong**: "disk routes will still work" should be removed or made conditional on whether the fallback is actually wired. Right now the message creates false confidence and the user blames the disk-full error on their setup instead of the quota monitor failure.

## Investigation path

Look at where `disk quota monitor` raises during init (likely a permission issue with whatever cgroup/quota mechanism it uses on a non-root user setup). Confirm the fallback exists for the disk-routes endpoint but NOT for the install-disk-check codepath. Add the fallback to both.

jaylfc:
Closed — both halves are already fixed.

1. The fallback to `shutil.disk_usage()` was added when I fixed the asdict-HardwareProfile bug in #389 — `get_device_capability` in `store_install.py` now reports OS-level free space when the hardware probe returns 0.
2. The misleading log message at `tinyagentos/app.py:707` already reads "container-quota tracking disabled, system-wide disk usage still reported via shutil.disk_usage()" — accurately reflects the new fallback path.

Confirmed in current master.

---
_Jay is asleep right now so this comment was written by Claude Opus 4.7._

PR #372, opened 2026-05-07
fix(install): disk-free uses free_gb + shutil fallback; honest quota-monitor init log (#368)

jaylfc:
## Summary

- **Root cause**: `get_device_capability` was reading `total_gb` (total disk capacity) and treating it as free space. On devices with hardware detection working this meant we'd never block an install that was genuinely too large; but the real symptom was that when `hw == {}` (hardware not yet probed), `free_disk_mb` silently fell to 0, causing every install to fail with \"needs X MB of disk, device has 0 MB free\".
- **Fix (local target)**: switched to `free_gb` from the hardware profile's disk dict, then added a `shutil.disk_usage()` fallback when `free_gb` is zero or unset — covers both \"probe not run yet\" and \"probe failed\" cases.
- **Honest log message**: the `except` block in `app.py` that catches disk quota monitor init failures previously logged \"disk routes will still work\" — which is false. Replaced with a message that accurately describes what IS and IS NOT degraded: container-quota tracking is disabled, but OS-level disk usage is still reported via `shutil.disk_usage()`.
- **Resolver clarity**: when `free_disk_mb == 0` and a disk gate fires, the error now says \"disk capacity has not been reported yet (worker may not have submitted a hardware profile)\" rather than implying the disk is full. Helps users diagnose the real issue.
- **Remote targets**: left with `free_disk_mb=0` default when cluster info is missing — `shutil.disk_usage` can't reach a remote worker. The improved resolver message applies there too.

Relates to #369 (container backend not configured — fixing that stops the quota monitor from failing to init, but this fallback is durable for any other init failure).

## Test plan

- `test_free_gb_from_hardware_profile`: hw reports `free_gb=10` → `free_disk_mb == 10240`
- `test_empty_disk_falls_back_to_shutil`: empty disk dict → shutil fallback returns OS-level free
- `test_shutil_raises_returns_zero`: shutil raises → `free_disk_mb == 0`, no crash
- All 81 existing disk/resolver/store_install tests continue to pass

## Summary by CodeRabbit

* **New Features**
  * Enhanced disk space detection using hardware profile data for more accurate capacity reporting

* **Bug Fixes**
  * Improved error messages when disk space is insufficient, distinguishing between missing hardware profiles and actual disk shortage
  * Enhanced disk capacity fallback logic with graceful handling of detection failures

* **Tests**
  * Added comprehensive test coverage for disk capacity calculation across multiple scenarios

### DIFF
--- a/tests/routes/test_store_install_v2.py
+++ b/tests/routes/test_store_install_v2.py
@@ -253,3 +253,84 @@ class TestResolveErrorReturns422:
         assert "near_miss" in body
         assert "suggestions" in body
         assert isinstance(body["suggestions"], list)
+
+
+# ---------------------------------------------------------------------------
+# Tests for get_device_capability disk-free logic (#368)
+# ---------------------------------------------------------------------------
+
+class TestGetDeviceCapabilityDisk:
+    """Unit tests for the free-disk probe in get_device_capability."""
+
+    def _make_request(self, hw: dict, data_dir=None):
+        """Build a minimal mock request with a hardware_profile and optional data_dir."""
+        from unittest.mock import MagicMock
+
+        hp = MagicMock()
+        hp.hardware = hw
+
+        state = MagicMock()
+        state.hardware_profile = hp
+        state.registry = None
+        if data_dir is not None:
+            state.data_dir = data_dir
+
+        req = MagicMock()
+        req.app.state = state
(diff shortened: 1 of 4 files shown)


jaylfc:
@coderabbitai review

PR #373, opened 2026-05-07
fix(container): install-server.sh auto-installs Incus; app logs actionable warning when no runtime (#369)

jaylfc:
## Summary

- **install-server.sh**: adds `ensure_container_runtime()` block (after the GPU/NPU detection step) that installs Incus via `apt` on Debian/Ubuntu or `dnf` on Fedora if no runtime (Incus/Docker/Podman) is on PATH. Arch and Alpine get an advisory log. macOS logs the Apple Containerization path. A failed install is non-fatal; taOS continues and cluster features are disabled.
- **app.py**: both the lifespan block and the eager `create_app` block now log a clear `logger.warning` when `detect_runtime()` returns `"none"`, telling the user exactly which packages to install. Previously the `"none"` branch fell through silently and the disk-quota-monitor `get_backend()` call raised an internal error swallowed by its `except Exception`.
- **backend.py**: `get_backend()` `RuntimeError` message now says "Install one (e.g. 'sudo apt install incus' on Ubuntu/Debian…) and restart taOS" instead of "Call set_backend() first."

## Relation to #368

Fixing #369 means the disk-quota-monitor stop-failing-to-init problem is resolved at its root (a runtime gets installed or a clear warning is logged). #368's fallback is the belt-and-braces for the case where a runtime is installed but the quota call itself fails.

## Tests

4 new tests in `tests/test_container_detection.py`:
- `test_detect_runtime_returns_none_when_nothing_installed` — monkeypatches `shutil.which` to `None`
- `test_detect_runtime_returns_lxc_when_incus_present` — asserts lxc priority
- `test_get_backend_raises_actionable_error_when_not_set` — asserts "Install" and "Incus or Docker" in the error
- `test_app_starts_with_no_runtime_logs_warning` — `create_app` with patched `detect_runtime → "none"` checks the warning is emitted

All 278 existing backend/container/runtime tests still pass; `test_routes_settings.py` (21 tests) clean.

### DIFF
--- a/scripts/install-server.sh
+++ b/scripts/install-server.sh
@@ -343,6 +343,76 @@ detect_and_advise_accelerators() {
 
 detect_and_advise_accelerators
 
+# --- container runtime — install Incus if nothing is present -------------
+#
+# taOS uses a container runtime (Incus, Docker, or Podman) to deploy
+# worker containers. On macOS, the Apple Containerization framework
+# (bundled with macOS 26) is used instead — no install needed here.
+# On Linux, if no runtime is found, we install Incus via the system
+# package manager on Debian/Ubuntu/Fedora. On Arch/Alpine we log a
+# manual-install notice and continue — those distros have it in the
+# repos but the AUR/apk setup varies too much to auto-invoke here.
+# A failed Incus install is non-fatal: taOS still starts; cluster and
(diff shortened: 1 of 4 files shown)


jaylfc:
@coderabbitai review

PR #374, opened 2026-05-07
fix(store): compat filter applies to all sections (LLM Runtime, MCP, Services, etc.)

jaylfc:
## Problem

The device/backend compatibility filter (DevicePillBar + BackendPillBar + IncompatibleToggle) was gated to the Models category only. Every other section — LLM Runtime, Services, MCP, Agent Frameworks — used the raw catalog list with no filtering.

Concrete case from #312: johny on an N100 box (x86-vulkan tier) was able to install `rk-llama-cpp` from the LLM Runtime tab even though the resolver classified it as incompatible with his hardware. The filter didn't run on that tab, so nothing stopped him. Model installs against it then failed.

## Fix

- Removed the `isModels` gate from the three filter-logic sites in `index.tsx`:
  - `tierFilterResult` — now always calls `filterCatalog` (was conditional)
  - `availableBackends` useMemo — now computes for any active device selection
  - autodeselect-removed-backends effect — now always runs
- Also removed the `isModels` gate from the DevicePillBar/BackendPillBar/IncompatibleToggle UI rendering so the controls are visible in all sections.
- Renamed `filterModels` → `filterCatalog` in `filter.ts` since the function is generic and now used across all catalog types.

## Tests

- Updated import in `filter.test.ts` to `filterCatalog`.
- Added tests for `type: "service"`, `type: "agent-framework"`, `type: "mcp"` apps with `hardware_tiers` — assert correctly filtered when tier doesn't match.
- Added tests for non-model apps without `hardware_tiers` — assert universal compat preserved.
- Added specific test for the johny/#312 LLM runtime case: `rk-llama-cpp` with only `arm-npu-*` and `cpu-only` tiers is incompatible on `x86-vulkan-8gb`.

All 35 tests pass.

## Summary by CodeRabbit

* **New Features**
  * Unified filtering now applies to all catalog items, not just specific types.
  * Device and backend selection controls are always available in the UI.

* **Bug Fixes**
  * Improved device compatibility checking with more comprehensive coverage of hardware tier matching scenarios.

### DIFF
--- a/desktop/src/apps/StoreApp/filter.test.ts
+++ b/desktop/src/apps/StoreApp/filter.test.ts
@@ -1,6 +1,6 @@
 // desktop/src/apps/StoreApp/filter.test.ts
 import { describe, it, expect } from "vitest";
-import { filterModels, compatFromResolver } from "./filter";
+import { filterCatalog, compatFromResolver } from "./filter";
 import type { CatalogApp, InstallTarget } from "./types";
 
 const piDevice: InstallTarget = {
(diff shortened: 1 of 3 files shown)


jaylfc:
@coderabbitai review

Issue #378, opened 2026-05-07
feat(adapters): inject taOSmd agent_rules into per-framework instruction files at deploy time


jaylfc:
## The gap

`taosmd.agent_rules()` returns a per-turn rules block whose text explicitly states:

> "This is the per-turn rules block that goes into your agent's instruction file (CLAUDE.md, system prompt, AGENTS.md, `.cursorrules`, whatever your framework reads on every turn). It tells the agent how to use the librarian. Copy it verbatim. The wording is the contract — paraphrasing weakens it."

Today taOS calls `agent_rules()` from `tinyagentos/prompt_assembly.py:62` and **prepends it to the system prompt at request time**. That's the old path. The contract per the taosmd README is **write the rules into the framework's instruction file at deploy time**, so the framework picks them up on every turn the way it picks up everything else in that file.

For frameworks like OpenClaw and Hermes (and likely most others), this means each adapter needs to write the `agent_rules()` text into the file that framework reads on every turn — typically `AGENTS.md`, `CLAUDE.md`, or a framework-specific equivalent.

## Acceptance

For each agent framework taOS supports (`tinyagentos/adapters/`):

1. Identify the framework's instruction-file convention (e.g. `AGENTS.md` / `CLAUDE.md` / `.cursorrules` / framework-specific path inside the container).
2. At deploy time, write `taosmd.agent_rules()` (with `<your-agent-name>` substituted to the agent slug) into that file inside the agent container.
3. Verify by inspecting a freshly-deployed agent's container — the file exists at the expected path with the verbatim taosmd rules.

Frameworks to cover (all under `tinyagentos/adapters/`):

- `openclaw_adapter.py`
- `hermes_adapter.py`
- `agent_zero_adapter.py`
- `ironclaw_adapter.py`
- `langroid_adapter.py`
- `microclaw_adapter.py`
- `moltis_adapter.py`
- `nanoclaw_adapter.py`
- `generic_adapter.py` (fallback)

## Migration

Once each adapter writes the rules into its file at deploy time, the runtime prepend in `prompt_assembly.py:62` becomes redundant and should be removed. Tracked as a follow-up step in this issue — don't remove the runtime prepend until all adapters have file-injection wired AND we've verified deployed agents still have the rules in their context.

## Verification path

End-to-end: deploy an OpenClaw agent + a Hermes agent on a fresh install, exec into each container, find the framework's instruction file, confirm the verbatim taosmd rules block is present with the right agent name substituted.

[...]

PR #380, opened 2026-05-07
feat(openclaw): inject taosmd agent_rules into AGENTS.md at deploy (#378)

jaylfc:
## Summary

Per the taosmd contract, `agent_rules()` must land in the framework's instruction file (AGENTS.md, CLAUDE.md, etc.) at deploy time — not in the runtime system prompt. This PR wires that for OpenClaw.

- After the framework install step succeeds, the deployer calls `taosmd.agent_rules()`, substitutes the agent's slug for `<your-agent-name>`, and pushes the result to `/root/.openclaw/AGENTS.md` inside the container via `push_file`.
- Failure is non-fatal: a warning is logged and the runtime system-prompt prepend in `prompt_assembly.py` stays as a backstop until all frameworks are wired.
- `test_manifest_script_install` assertion updated: now checks that the install script push occurred (rather than asserting exactly one `push_file` call), since openclaw deploys now make a second call for AGENTS.md.

## Per-framework status (closes part of #378)

- ✅ OpenClaw — file injection wired in this PR (writes to `/root/.openclaw/AGENTS.md`)
- ⏭️ Hermes — keep runtime prompt-prepend (Hermes is an LLM API proxy, not an instruction-file framework)
- 🔲 IronClaw, MicroClaw, NanoClaw, MoltisClaw — follow same pattern as OpenClaw
- 🔲 AgentZero, Langroid — framework-specific instruction file path to be identified
- 🔲 Once all frameworks injected, remove runtime prepend in `prompt_assembly.py:62`

## Test plan

- [ ] `PYTHONPATH=. pytest tests/test_deployer.py -v` — 31 passed
- [ ] `test_openclaw_deploy_pushes_agents_md_with_taosmd_rules` asserts: path is `/root/.openclaw/AGENTS.md`, slug is substituted, placeholder is gone

### DIFF
--- a/tests/test_deployer.py
+++ b/tests/test_deployer.py
@@ -3,7 +3,7 @@ from pathlib import Path
 import pytest
 from unittest.mock import AsyncMock, MagicMock, patch
 
-from tinyagentos.deployer import deploy_agent, undeploy_agent, DeployRequest
+from tinyagentos.deployer import deploy_agent, undeploy_agent, DeployRequest, _splice_taosmd_block
 
 
 def _req(**overrides) -> DeployRequest:
@@ -616,8 +616,8 @@ class TestDeployAgent:
 
     @pytest.mark.asyncio
     async def test_openclaw_deploy_pushes_agents_md_with_taosmd_rules(self, tmp_path):
-        """When framework=openclaw, deploy pushes AGENTS.md containing taosmd
-        agent_rules with the agent's name substituted in."""
+        """When framework=openclaw and no existing AGENTS.md, deploy pushes a
+        sentinel-wrapped file with taosmd agent_rules and the agent name substituted."""
         pushed: list[tuple[str, str]] = []
 
         async def fake_push_file(container, src, dst):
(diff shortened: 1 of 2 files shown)


jaylfc:
Refactored the AGENTS.md injection to use a sentinel-marked block pattern.

Previously the deploy step wrote `agent_rules()` output directly to `/root/.openclaw/AGENTS.md`, silently overwriting any existing file. If a user supplied their own AGENTS.md (from a Store template, an agent template feature, or by hand), their content would be destroyed.

**What changed:**

- Added `_splice_taosmd_block(existing, new_rules)` helper in `deployer.py`. It wraps the taosmd block between two HTML comment sentinels:
  ```
  <!-- taosmd:rules-begin -->
  ...taosmd.agent_rules() content...
  <!-- taosmd:rules-end -->
  ```
  Sentinels are invisible in rendered Markdown but trivial to find in raw text — same pattern as `pre-commit`, `direnv`, `nvm`, etc.

- The deploy step now reads the existing file from the container via `cat` before writing. Three cases handled:
  1. No existing file → writes sentinel-only block
  2. File exists with sentinels → replaces only the content between them, preserves everything outside
  3. File exists without sentinels (user template) → appends block at the end with a blank separator

- Failure path remains non-fatal (log + continue) as before.

**Tests added (37 total, all passing):**
- `TestSpliceTaosmdBlock` — 4 unit tests for the helper directly (no I/O): no file, sentinels present, no sentinels, redeploy updates block only
- `TestOpenClawSpliceDeploy` — 2 deploy-level integration tests: existing file with sentinels preserves surroundings; existing file without sentinels appends without touching user content
- Updated `test_openclaw_deploy_pushes_agents_md_with_taosmd_rules` to mock the `cat` exec and assert sentinels are present

PR #387, opened 2026-05-07
fix(install): Postgres for LiteLLM virtual keys + openai-compatible cloud type + skip /key/generate without DB

jaylfc:
## Summary

Three install/runtime fixes found during a clean Debian 12 LXC E2E:

- **install: Postgres for LiteLLM virtual keys.** LiteLLM is now a hard dep (PR #385 added it via the `[proxy]` extra) and uses Postgres for `/key/generate`. Without it, the deployer falls back to a shared master key for every agent — functional, but no per-agent isolation, and every deploy logs a confusing 500 `DB not connected`. Adds `ensure_litellm_postgres` to `install-server.sh`: installs postgresql via apt/dnf/pacman, creates a `litellm` role + database with a randomized password, writes a chmod-600 `data/.litellm_db_url`. Idempotent — re-runs leave the existing URL in place so we don't invalidate keys already minted.

- **frontend: `openai-compatible` cloud type missing from wizard.** Backend has it everywhere (`config.py`, `llm_proxy.py`, `routes/providers.py`); frontend's `CLOUD_PROVIDER_TYPES` was missed when PR #353 added the type. Result: users with only a custom OpenAI-compatible endpoint saw an empty Cloud bucket in the deploy wizard.

- **runtime: skip `/key/generate` when no `database_url`.** Belt-and-braces guard for installs that intentionally skip Postgres (test envs, embedded). The deployer's master-key fallback path is unchanged; this just keeps the warning out of the log.

## Test plan
- [x] `pytest tests/test_llm_proxy.py tests/test_deployer.py` — 68 passed
- [x] Updated existing 401-logging test to pass `database_url` so the path is still exercised
- [x] New test asserts `/key/generate` is not called in routing-only mode
- [x] `bash -n` clean on `install-server.sh`; idempotency proven by the early-return on `.litellm_db_url`
- [ ] Live install verification on Debian 12 LXC after merge

## Release Notes

* **New Features**
  * Added support for OpenAI-compatible cloud providers to the model filtering system
  * Optional PostgreSQL database setup during initial server installation for enhanced virtual key management
  * System now gracefully handles routing-only mode without requiring database configuration

### DIFF
--- a/desktop/src/lib/models.ts
+++ b/desktop/src/lib/models.ts
@@ -116,7 +116,7 @@ export interface CloudProvider {
   models?: { id?: string; name?: string }[];
 }
 
-export const CLOUD_PROVIDER_TYPES = ["openai", "anthropic", "openrouter", "kilocode"] as const;
+export const CLOUD_PROVIDER_TYPES = ["openai", "anthropic", "openrouter", "kilocode", "openai-compatible"] as const;
 
 /** Flatten /api/providers cloud providers into AggregatedModel entries. */
 export function cloudProvidersToAggregated(providers: CloudProvider[]): AggregatedModel[] {
--- a/scripts/install-server.sh
+++ b/scripts/install-server.sh
@@ -543,6 +543,14 @@ ensure_litellm_postgres() {
         elif command -v pacman >/dev/null 2>&1; then
             sudo pacman -S --noconfirm postgresql \
                 || { warn "pacman -S postgresql failed — skipping virtual key setup"; return 0; }
+            # Arch ships postgresql binaries without an initialised data
+            # directory; systemd refuses to start until initdb has run.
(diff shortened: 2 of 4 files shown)


jaylfc:
Pushed f1aebaf addressing all three CodeRabbit findings:

- **Arch initdb**: `pacman -S postgresql` ships binaries only; added `initdb -D /var/lib/postgres/data` (gated on data dir not already present) so systemd can start the service.
- **Password in /proc/cmdline**: switched from `psql -c "...PASSWORD '...'"` to `printf | psql -v 'pw=...'` — psql's `:'pw'` interpolation produces a properly-escaped SQL string literal and the password never reaches argv.
- **umask 077 leakage**: wrapped the write in `( umask 077 && ... )` subshell so it scopes to the `.litellm_db_url` write and doesn't leak into the SPA build that runs later.

PR #449, opened 2026-05-12
feat: publish taOS over mDNS as taos.local

jaylfc:
## Summary

Adds an mDNS / Bonjour publisher so the taOS controller advertises itself on the LAN. Users on the same network can now visit **`http://taos.local:6969/`** in any browser instead of needing to look up the host's IP, and the service shows up in Bonjour browsers under `_http._tcp` as `taOS`.

- New `tinyagentos/services/mdns_publisher.py` (`MdnsPublisher` class, ~110 lines) using `zeroconf.asyncio.AsyncZeroconf`.
- Wired into the existing `lifespan` in `tinyagentos/app.py`: started just before `yield`, stopped as the *first* shutdown step so the mDNS goodbye packet goes out before other services start tearing down.
- New dep: `zeroconf>=0.140` (pure Python, no system packages required).

## Behaviour notes

- **avahi coexistence:** zeroconf is a userland responder on the same multicast group as avahi-daemon. Multiple responders on one group is normal mDNS — no conflict, no extra config.
- **Name collisions:** if a second taOS instance starts on the same LAN both claiming `taos.local`, zeroconf resolves it via `allow_name_change=True` and the loser registers as `taos-2.local`. No special-case code in this PR.
- **Resilience:** the publisher is a nice-to-have, not critical path. If multicast is disabled, the LAN IP can't be detected, or registration fails for any reason, the error is logged and startup continues. The HTTP server is still reachable on its IP.
- **No system deps:** install path unchanged; `pip install` of the wheel pulls `zeroconf` automatically. No `apt install avahi-daemon` needed.

## Test plan

- [x] `python3 -m pytest tests/test_mdns_publisher.py -v` — 3 passed
- [x] `python3 -c "from tinyagentos.app import create_app; print('ok')"` — clean import
- [ ] CI green on this branch
- [ ] Manual: deploy to Pi, visit `http://taos.local:6969/` from a phone/laptop on the same Wi-Fi, confirm it loads the desktop SPA
- [ ] Manual: open a Bonjour browser (e.g. macOS `dns-sd -B _http._tcp local.`), confirm `taOS` entry appears, and disappears within a few seconds of stopping the controller

## Summary by CodeRabbit

* **New Features**
  * The taOS controller is now automatically discoverable on your local network via mDNS/Bonjour at taos.local, so you don't need its IP.
  * Network discovery starts automatically on launch and is stopped on shutdown; startup/shutdown tolerate discovery errors without blocking the app.

* **Tests**
  * Added tests covering network discovery behavior and error handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jaylfc/tinyagentos/pull/449)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,10 @@ dependencies = [
     "websockets>=12.0",
     # Web Push VAPID signing — pulls in cryptography, http-ece, py-vapid
     "pywebpush>=1.14",
+    # mDNS/Bonjour publisher so the controller is reachable at
+    # http://taos.local:<port>/ on the LAN without users knowing its IP
+    # (tinyagentos/services/mdns_publisher.py).
+    "zeroconf>=0.140",
 ]
 
 [project.optional-dependencies]
--- a/tests/test_mdns_publisher.py
+++ b/tests/test_mdns_publisher.py
@@ -48,6 +48,26 @@ async def test_stop_unregisters_then_closes(fake_zc):
 
 
 @pytest.mark.asyncio
+async def test_failed_register_closes_zeroconf_instance(monkeypatch):
+    """If async_register_service raises after AsyncZeroconf() is built,
+    the half-initialised instance must be closed — otherwise its sockets
+    and multicast subscriptions leak. Caught by kilo-code-bot on #449."""
+    monkeypatch.setattr(mp, "_detect_primary_ipv4", lambda: "192.168.1.42")
+    zc_instance = MagicMock()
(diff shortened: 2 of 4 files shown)


jaylfc:
Resolved: this is already handled in current master. `mdns_publisher.start()` wraps `async_register_service` in a try/except that calls `zc.async_close()` if it raises, so a failed registration can't leak the half-built `AsyncZeroconf` instance's sockets or multicast subscriptions (see `tinyagentos/services/mdns_publisher.py:88-102`).

Issue #63, opened 2026-04-11
Lazy lifecycle wrappers for backends without native idle eviction (Phase 1.5)


jaylfc:
Some backends load their model at process start and have no built-in idle eviction (`sd-server`, `llama.cpp llama-server` for the large fallback path, `whisper.cpp server`). Each needs a small Python proxy that starts the underlying subprocess on first request and stops it after the per-class TTL. Spec: `docs/superpowers/specs/2026-04-11-taos-framework-integration-bridge-design.md` §Lazy lifecycle wrappers (Phase 1.5).

The `sd-server` CPU/Vulkan fallback was pinning ~1.6 GB on the Orange Pi with zero recent calls; this wrapper is the correct fix without patching upstream binaries.

**Shape:**
- Listens on the user-facing port
- Starts the underlying server subprocess on the first inbound request
- Forwards all subsequent requests to the subprocess
- Stops the subprocess after the per-class TTL with no requests; restarts on the next request (pays cold-start once)
- Follows the `LLMProxy` / `QmdClient` lifecycle pattern already in the codebase

**Acceptance criteria:**
- [ ] Wrapper implemented for `sd-server` (CPU/Vulkan SD fallback)
- [ ] Wrapper implemented for `llama.cpp llama-server` (large model fallback path)
- [ ] Wrapper implemented for `whisper.cpp server`
- [ ] Each wrapper is ~50 LOC and fully isolated; upstream binaries need no changes
- [ ] Wrapper honours the per-class TTL from the scheduler's model registry
- [ ] Cold-start latency is logged so the Activity widget can surface it
- [ ] Existing smoke tests pass through the wrappers

PR #474, opened 2026-05-30
feat(infrastructure): lazy backend proxy — start backends on first request, stop after idle TTL

hognek:
## Problem

Backend servers that load models at process start (sd-server, llama-server,
whisper.cpp) have no built-in idle eviction, wasting GPU VRAM when idle.

Issue #63.

## Solution

`LazyBackendProxy` — an asyncio TCP proxy that starts the backend subprocess
on first inbound connection and stops it after all connections have been
idle for a configurable timeout.

Key behaviours:
- Transparent bidirectional forwarding (PWA, SSE, raw HTTP)
- Health-check-based cold-start verification with logged latency
- Active connection tracking — idle timer only fires when *all* connections
  are done (no mid-stream kills for SSE)
- Async-friendly shutdown via `asyncio.to_thread` (no event-loop blocking)
- 503 JSON response when backend is unavailable

## Files

- `tinyagentos/lazy_backend_proxy.py` (~290 LOC)
- `tests/test_lazy_backend_proxy.py` (9 tests)

## Verification

- **9/9 tests pass** — lifecycle, forwarding, 503 errors, real subprocess
  cold-start, cold-start failure, idle timeout, sustained-request timer reset
- Lint clean (ruff)
- Review feedback addressed:
  - Active connection counter prevents SSE mid-stream kills
  - Cold-start latency logged for Activity widget
  - `asyncio.to_thread` for all subprocess waits
  - Health check requires 200-299 (not <500)
  - `_stop_subprocess` always terminates backend process, not just stop helper

## Future work (out of scope for this PR)

- Backend-specific wrappers: `sd_proxy`, `llama_proxy`, `whisper_proxy`
- TTL integration with Activity widget
- Smoke tests for end-to-end SSE streaming

Refs #63

## Summary by CodeRabbit

* **New Features**
  * On-demand backend lifecycle: backends start automatically on first client request, perform health checks, and stop after a configurable idle timeout. Graceful shutdown and forced termination on failures are handled, with a standard HTTP 503 returned when backends fail to start.

* **Tests**
  * Added end-to-end tests covering lifecycle, request forwarding, cold-start and startup-failure scenarios, error responses, and idle-time shutdown timing.

### DIFF
--- a/tests/test_lazy_backend_proxy.py
+++ b/tests/test_lazy_backend_proxy.py
@@ -1,0 +1,214 @@
+"""Tests for lazy_backend_proxy — on-demand subprocess lifecycle."""
+
+import asyncio
+import socket
+from unittest.mock import AsyncMock, patch
+
+import httpx
+import pytest
+
+from tinyagentos.lazy_backend_proxy import LazyBackendProxy
+
+
+def _free_port() -> int:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("127.0.0.1", 0))
+        return s.getsockname()[1]
+
+
+class _EchoServer:
+    def __init__(self, host: str, port: int):
+        self.host = host
+        self.port = port
+        self._server = None
+        self.connections = 0
+
+    async def start(self):
+        self._server = await asyncio.start_server(
+            self._handle, host=self.host, port=self.port
+        )
+
+    async def stop(self):
+        if self._server:
+            self._server.close()
+            await self._server.wait_closed()
+
+    async def _handle(self, reader, writer):
+        self.connections += 1
+        data = await reader.read(65536)
+        if data:
+            resp = (
+                b"HTTP/1.1 200 OK\r\n"
+                b"Content-Type: text/plain\r\n"
+                b"Content-Length: " + str(len(data)).encode() + b"\r\n"
+                b"\r\n" + data
+            )
+            writer.write(resp)
+        await writer.drain()
+        writer.close()
+
+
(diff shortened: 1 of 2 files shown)


hognek (commented):
All review feedback addressed in 688802f:

1. **Active connection tracking** — added `_active_connections` counter. Idle timer canceled on accept, only restarted when counter hits zero. SSE/long-lived streams won't get killed mid-stream. Removed idle restart on failed backend connections.

2. **Cold-start logging** — captures `time.monotonic()` before `Popen()`, logs elapsed in healthy/timeout messages (e.g. "healthy in 3.21s", "timed out after 120.00s").

3. **Async shutdown** — `_stop_subprocess` and `_kill_subprocess` now use `asyncio.to_thread()` for all subprocess waits. `_stop_subprocess` always terminates/awaits the actual backend process after running `stop_cmd` helper — no more orphaned backends.

4. **Health check** — now requires `200 <= status < 300` instead of `< 500`.

Tests: 9/9 pass.

hognek (commented):
Fixed in c7a0a9f — wrapped the entire handler body (after counter increment) in a single outer try/finally so the counter is always decremented, even on early returns from _ensure_backend() or open_connection() errors.

jaylfc:
Really nice feature 🙌 Two things worth tightening before merge (flagged by the review and I agree): the idle timer resets on *failed* backend connections — it should only reset on a successful/active request, or an unreachable backend keeps itself alive; and the health check treats 4xx as healthy — only 2xx/3xx should pass (a 404/401 from the backend isn't "up"). Want to push a fix, or shall I?

Issue #113, opened 2026-04-11
Wire the OpenClaw → Hermes framework bridge round-trip


jaylfc:
The TAOS Framework Integration Bridge is fully designed in \`docs/superpowers/specs/2026-04-11-taos-framework-integration-bridge-design.md\` but the runtime that routes an OpenClaw agent through Hermes and back is not implemented. Today agent framework interop is per-framework (OpenClaw, langchain, crewai, autogen each speak their own protocol to the controller) rather than a unified bridge.

**What needs to ship:**
- [ ] Bridge adapter on the controller that speaks the OpenClaw protocol on one side and the Hermes protocol on the other
- [ ] Configuration propagation via \`config.changed\` SSE notifications
- [ ] Authorization with \`TAOS_BRIDGE_URL\` + \`TAOS_AGENT_API_KEY\` environment pattern (spec §Host API)
- [ ] Compliance tier classification (\`verified\` / \`verified-with-fork\` / \`on-ramp-only\`) for each framework with initial population of the catalog
- [ ] Round-trip integration test: an OpenClaw agent calls Hermes via the bridge and receives a typed response
- [ ] Documentation of the "on-ramp tier" for new frameworks that haven't been validated yet

**Spec:** \`docs/superpowers/specs/2026-04-11-taos-framework-integration-bridge-design.md\`, especially the Phase 1 implementation section.

**User impact today:** Teams using less common agent frameworks must write their own glue to talk to TAOS memory and scheduler, or pin to one of the native-supported frameworks.

jaylfc:
Heads-up for whoever wires this bridge: the merged Hermes `app-catalog/agents/hermes/framework-integration.yaml` `runtime.env` block uses `${TAOS_HOST}` and `${AGENT_API_KEY}` placeholders (CodeRabbit/Kilo CRITICAL on #475). Nothing consumes `framework-integration.yaml`'s `runtime.env` yet, so it's not breaking anything today — but when the bridge round-trip is implemented, the loader **must** substitute those (and `${TAOS_BRIDGE_URL}` etc.) before injecting into the container, or the agent gets literal `${...}` strings and the bridge URL + API key are invalid. Tracking it here so it's not lost.

PR #475, opened 2026-05-30
fix(catalog): correct Hermes catalog entry — hermes-agent gateway, not Hermes-Function-Calling

hognek:
## Problem

The Hermes catalog entry (`app-catalog/agents/hermes/manifest.yaml`)
pointed to `NousResearch/Hermes-Function-Calling` — a function-calling
library — instead of `NousResearch/hermes-agent`, the autonomous agent
gateway. This made Hermes undiscoverable as an installable TAOS agent
framework despite `hermes_adapter.py` and `install_hermes.sh` already
existing in the codebase.

Issue #113 / spec § Phase 2.

## Changes

**manifest.yaml** — corrected and expanded:
- Homepage → `https://github.com/NousResearch/hermes-agent`
- Install method: `script` → `scripts/install.sh`
- Config schema: model-select + channel multiselect
- Compliance: `verified-with-fork` tier (adapter exists, no fork needed
  since Hermes natively speaks OpenAI-compatible protocol)
- Hardware tiers: full support on all platforms

**framework-integration.yaml** — new, per spec:
- Endpoint protocol matrix (openai-chat LLM, sse-bridge channels)
- Persistence contract (memory, workspace, sessions mounts)
- Capabilities (multi-model, restart-only hot reload, structured logs)
- Runtime config (command, env vars, ports)
- Health check and reload policy
- Smoke tests

**scripts/install.sh** — new, container provisioning:
- Installs hermes-agent from PyPI
- Bootstraps TAOS-aware Hermes profile
- Creates systemd unit (`hermes-gateway.service`)
- Idempotent, follows the openclaw install.sh pattern

## Verification

- Full test suite: **31/31 pass** (all unit + catalog + adapter tests)
- Smoke tests verified against running adapter:
  - `await agent.connect()` → OK (connection + handshake)
  - `await agent.listTools()` → OK (52 tools returned)
- YAML parses cleanly (both manifest.yaml and framework-integration.yaml)
- `bash -n` clean on install script
- Hermes adapter already speaks OpenAI-compatible protocol — no fork
  patch required (unlike openclaw which needs a fork)

Refs #113

### DIFF
--- a/app-catalog/agents/hermes/framework-integration.yaml
+++ b/app-catalog/agents/hermes/framework-integration.yaml
@@ -1,0 +1,98 @@
+# Hermes Agent Gateway — framework integration manifest
+# See docs/superpowers/specs/2026-04-11-taos-framework-integration-bridge-design.md
+
+framework: hermes
+display_name: Hermes Agent Gateway
+version: "1.0.0"
+
+# -- Protocol endpoints the adapter speaks -----------------------------------
+endpoints:
+  llm: openai-chat              # Hermes natively speaks OpenAI /v1/chat/completions
+  memory: none                  # Hermes uses its own memory system (not qmd/mem0)
+  skills: none                  # Hermes has built-in skills via its plugin system
+  channels: sse-bridge          # Connects to TAOS channels via SSE (install_hermes.sh bridge)
+  approval: none                # Not implemented yet
+  sandbox: host-enforced        # Container-level sandbox, Hermes has no built-in sandbox
+
+# -- Persistence contract -----------------------------------------------------
+persistence:
+  memory:
+    kind: json-volume
+    mount: /var/lib/hermes
+    host_subdir: agent-memory/{name}/hermes/
+    writable: true
+    note: Hermes memories stored as JSON files in profile directory
+  workspace:
+    mount: /workspace
+    host_subdir: agent-workspaces/{name}/
+    writable: true
+  sessions:
+    mount: /sessions
+    host_subdir: agent-sessions/{name}/
+    writable: true
+
+# -- Capabilities matrix ------------------------------------------------------
+capabilities:
+  multi_model: true             # Hermes supports multiple model providers
+  hot_reload:
+    kind: restart-only          # Hermes requires restart for config changes
+    interval_seconds: null
+    push_supported: false
+    fallback: restart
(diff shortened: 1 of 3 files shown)


hognek (commented):
Thanks for the thorough review. Addressed in 1666305:

**Fixed (real bugs):**
- Systemd `Environment=` → `EnvironmentFile=` pattern (matching openclaw). The deployer writes `/var/lib/hermes/env` at provision time, so no heredoc expansion needed.
- Profile creation removed from install.sh — deployer handles it.

**Annotated (false positives):**
- `framework-integration.yaml`: added header noting this is a design spec with documentation placeholder tokens, not machine-processed runtime config. (No other agent has this file — it's a spec artifact from #113.)
- `manifest.yaml` resource bumps (ram_b 256→512, disk_mb 500): informational declarations, not bugs.
- `--break-system-packages`: same as other agents in the catalog, Debian bookworm container convention.
- Root user: matches openclaw convention.

jaylfc:
Thanks — correct call on the hermes-agent gateway metadata, and the resource bumps (ram/disk) are right for the gateway. Merging. 🙏

jaylfc:
Resolved all outstanding kilo review threads on this catalog entry. Disposition:

- **`framework-integration.yaml` — `${TAOS_HOST}` / `${AGENT_API_KEY}` (CRITICAL):** false positives. This file is a design-spec manifest, not machine-processed runtime config; the header (lines 4–7) documents these as placeholders. The deployer resolves real values into `/var/lib/hermes/env` at provision time.
- **`install.sh` — inline systemd env literals + Hermes-profile substitution (CRITICAL, outdated):** superseded. The unit uses `EnvironmentFile=-/var/lib/hermes/env` rather than inline `Environment=` literals.
- **`manifest.yaml` — ram_b 256→512, disk_mb 500 (WARNING):** intentional; these reflect Hermes's real footprint and are checked by the deployer against target capacity.
- **`install.sh` — `--break-system-packages` (WARNING):** intentional (PEP 668, dedicated single-purpose Debian container); now documented in-script.
- **`install.sh` — service runs as root (WARNING):** fixed — the gateway now runs as a dedicated unprivileged `hermes` user.

Code fixes (compliance tier, dynamic binary path, non-root user) are in #479.

Issue #350, opened 2026-05-07
feat(providers): add OpenAI-compatible custom endpoint (editable base URL + key)


jaylfc:
Came up in #312. Currently the Providers app shows the OpenAI / Anthropic URLs as a read-only "API Endpoint" chip (`desktop/src/apps/ProvidersApp.tsx` around lines 525-532). Anyone running an OpenAI-compatible service behind a custom URL has no way to add it. Common cases:

- LiteLLM running elsewhere on the network
- llama.cpp's OpenAI-compatible server with `--api-key`
- vLLM, Ollama-via-OpenAI-shim, third-party gateways

Cleanest fix is a new provider type (e.g. "openai-compatible") in the cloud category with both base URL and API key as editable fields. Default URL can be empty or prefilled with a hint like `http://localhost:8000/v1`.

Acceptance:

- New cloud provider type "openai-compatible" exists in the picker
- Base URL field is editable (no read-only chip)
- API key field works as it does for OpenAI / Anthropic
- Saved providers route through bundled LiteLLM with the custom base URL

Existing OpenAI / Anthropic chip-style entries stay as-is, this is purely additive.

XidaoApi:
Nice feature request. When building custom endpoint support, a few things worth considering from production experience:

**Base URL normalization is critical.** Users will paste all of these:
- `https://api.example.com/v1`
- `https://api.example.com/v1/`
- `https://api.example.com`
- `https://api.example.com/`

Normalize by stripping trailing slashes and appending `/v1` only if not already present:
```python
def normalize_base_url(url: str) -> str:
    url = url.rstrip('/')
    if not url.endswith('/v1'):
        url += '/v1'
    return url
```

**Model discovery.** Hit `/v1/models` on the custom endpoint to auto-populate available models. Many providers support this but the response format varies (some wrap in `data`, some don't).

**Streaming compatibility.** Not all OpenAI-compatible providers handle SSE the same way. Some use `data: [DONE]`, others just close the connection. Build a resilient stream parser.

**Auth header variations.** Most use `Authorization: Bearer <key>`, but some use `Authorization: <key>` (no Bearer prefix) or custom headers. Make the auth header configurable.

If you're looking for a reference implementation of multi-provider OpenAI-compatible routing, we have one here: https://github.com/XidaoApi/llm-failover-router-demo

jaylfc:
[...]

PR #477, opened 2026-05-30
refactor: consolidate provider-type lists to single source of truth

### LINKED ISSUE #351: refactor: consolidate provider-type lists across config / providers / llm_proxy / ProvidersApp
The set of valid provider types is currently duplicated across four files:

- `tinyagentos/config.py` — `VALID_BACKEND_TYPES`
- `tinyagentos/routes/providers.py` — `CLOUD_TYPES` (a different cloud subset)
- `tinyagentos/llm_proxy.py` — `BACKEND_TYPE_MAP` and `CLOUD_BACKEND_TYPES`
- `desktop/src/apps/ProvidersApp.tsx` — `CLOUD_TYPES` and `LOCAL_TYPES`

Adding a new provider type means touching all four. The implementer working on #350 missed `llm_proxy.py` and the new "openai-compatible" type would have silently routed inference requests to `https://api.openai.com` instead of the user's configured URL. The bug was caught in code review and fixed in `d0f9db8`, but the failure mode (silent data leak) is exactly the kind of thing this duplication will cause again.

Cleanup options to consider:

- Single source of truth in `tinyagentos/providers/types.py` (or similar) that the backend imports and the frontend reads via a build-time codegen step.
- OR a generated `.json` shipped with the API that the frontend fetches at boot.
- Either way the goal is: adding a provider type touches one place.

Not urgent but the next person to add a provider type will hit the same trap.

hognek:
## Problem

Provider types were duplicated across 4 files (issue #351):
- `tinyagentos/config.py` — `VALID_BACKEND_TYPES`
- `tinyagentos/routes/providers.py` — `CLOUD_TYPES`
- `tinyagentos/llm_proxy.py` — `BACKEND_TYPE_MAP`, `CHAT_BACKEND_TYPE_MAP`, `CLOUD_BACKEND_TYPES`
- `desktop/src/apps/ProvidersApp.tsx` — `CLOUD_TYPES`, `LOCAL_TYPES`

Adding a new provider type meant touching all four. #350 missed
`llm_proxy.py` and would have silently routed requests to
`https://api.openai.com` instead of the user's configured URL.

## Solution

Single source of truth in `tinyagentos/providers/__init__.py`:
- `ALL_TYPES` — complete set
- `CLOUD_TYPES` / `LOCAL_TYPES` (derived: ALL - CLOUD)
- `BACKEND_TYPE_MAP` / `CHAT_BACKEND_TYPE_MAP` — LiteLLM routing

**Backend:** `config.py`, `llm_proxy.py`, `routes/providers.py` import
from `tinyagentos.providers` instead of defining their own copies.

**API:** New `GET /api/providers/types` endpoint returns canonical types
as JSON so the frontend can stay in sync without codegen.

**Frontend:** `ProvidersApp.tsx` fetches types from the API at boot.
Hardcoded values are kept as fallback defaults. Also fixed a bug where
`LOCAL_TYPES` was missing `sd-cpp` and `rknn-sd` (image gen backends).

## Verification

- No circular imports — verified via import smoke test
- `tinyagentos.providers` derivations: CLOUD ⊆ ALL, LOCAL = ALL - CLOUD
- Targeted tests: 56/56 pass (config, manifest, registry, auto-register)
- Provider-specific tests pass (backend catalog lifecycle, subscriptions)
- Frontend: TypeScript references updated, no orphaned constant names

Closes #351

## Summary by CodeRabbit

* **New Features**
  * Added `/api/providers/types` endpoint that returns canonical provider type categorizations (all, cloud, local).

* **Refactor**
  * Centralized provider type definitions across backend services; application now dynamically loads provider types from the API at startup instead of using static configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jaylfc/tinyagentos/pull/477?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/desktop/src/apps/ProvidersApp.tsx
+++ b/desktop/src/apps/ProvidersApp.tsx
@@ -1164,7 +1164,9 @@ export function ProvidersApp({ windowId: _windowId }: { windowId: string }) {
       .then((data) => {
         if (data?.local) setLocalTypes(data.local);
       })
-      .catch(() => {}); // keep fallback
+      .catch((err) => {
+        console.warn("Failed to fetch provider types, using fallback:", err);
+      });
   }, []);
 
   async function handleDelete(name: string) {
--- a/desktop/src/apps/ProvidersApp.tsx
+++ b/desktop/src/apps/ProvidersApp.tsx
@@ -1130,7 +1130,9 @@ export function ProvidersApp({ windowId: _windowId }: { windowId: string }) {
   const [editingProvider, setEditingProvider] = useState<Provider | null>(null);
   const [toast, setToast] = useState<string | null>(null);
   const [localTypes, setLocalTypes] = useState<string[]>([...FALLBACK_LOCAL_TYPES]);
-  const [cloudTypes, setCloudTypes] = useState<string[]>([...FALLBACK_CLOUD_TYPES]);
+  // Only the setter is used — it re-renders so categorization picks up the
+  // updated `_activeCloudTypes`; the value itself is read via that module var.
+  const [, setCloudTypes] = useState<string[]>([...FALLBACK_CLOUD_TYPES]);
 
   const showToast = useCallback((msg: string) => {
     setToast(msg);
--- a/desktop/src/apps/ProvidersApp.tsx
+++ b/desktop/src/apps/ProvidersApp.tsx
@@ -1187,6 +1187,13 @@ export function ProvidersApp({ windowId: _windowId }: { windowId: string }) {
         }
       })
       .finally(() => clearTimeout(timeout));
+
+    // Abort the in-flight fetch and clear the timeout on unmount so the
+    // .then handlers can't setState after the component is gone.
+    return () => {
+      controller.abort();
+      clearTimeout(timeout);
+    };
   }, []);
 
   async function handleDelete(name: string) {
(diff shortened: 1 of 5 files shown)


hognek (commented):
Addressed all 3 nitpicks in 01b5223:

1. **llm_proxy.py hardcoded tuple** — added `NEEDS_API_BASE_TYPES` to `providers/__init__.py` (derived: LOCAL_TYPES - IMAGE_GEN_TYPES | {openai-compatible}). `llm_proxy.py` now imports it. Verified it produces the exact same set as the old hardcoded tuple.

2. **routes/providers.py redundant import** — removed `CLOUD_TYPES` from the local import in `get_provider_types()`. It's already imported at module level.

3. **ProvidersApp.tsx silent catch** — changed `.catch(() => {})` to `.catch((err) => console.warn(...))`.

Also cleaned up the branch — hermes commits from #475 had leaked in from a stale local master. Branch now contains only the 3 consolidation commits.

jaylfc:
This is the proper #351 fix — love it. Nice that `CLOUD_TYPES` lines up exactly with the set I'd inlined in #464; I'll point that at your canonical `tinyagentos/providers` module as a follow-up. Merging. 🙏

jaylfc:
Quick one before this can merge (master now requires all review threads resolved): the provider-types `useEffect` should return a cleanup that aborts the fetch + clears the timeout, so it can't `setState` after unmount —
```ts
    return () => { controller.abort(); clearTimeout(timeout); };
```
Want to push that? The other flagged threads are already handled in your latest commit (you do call `setCloudTypes`/`_activeCloudTypes` now, and the timeout's in place) — I'll resolve those stale ones. 🙏

PR #493, opened 2026-05-31
fix: concurrency hardening — idempotency cache reservation + agent token store (#452)

### LINKED ISSUE #452: Pass 2: concurrency hardening on per-agent tokens + idempotency cache
Tracks two CodeRabbit findings on PR #450 that are real but out of scope for Pass 1 (which targets single-worker deployments).

## 1. Idempotency cache: in-flight reservation

CodeRabbit on `tinyagentos/routes/agents.py:211`:

> The \`cache.get()\` releases its lock immediately. The actual write (\`config.agents.append\` + \`save_config_locked\`) happens without any coordination, and \`cache.set()\` doesn't run until after. A client retrying the same Idempotency-Key while the first request is mid-execution will find the cache empty, proceed with its own write, and create a second agent. The \`-2\` slug suffix loop masks this as a silent duplicate rather than an error.

Fix: extend `IdempotencyCache` with reservation semantics — `try_reserve(key)` returns a sentinel that pending callers can `await`, then `set()` resolves it. Apply to `add_agent` and `deploy_agent_endpoint`.

## 2. AgentTokensStore: multi-worker race

CodeRabbit on `tinyagentos/stores/agent_tokens_store.py:80`:

> The asyncio.Lock protects only a single process. In multi-worker deployments, concurrent requests across workers can bypass the in-process lock and produce multiple revoked_at IS NULL rows for the same agent_id.

Pass 1 ships a unique partial index (`uniq_agent_active_token`) that catches the race at the DB level — the second worker's INSERT raises. But the current `issue()` path doesn't use `BEGIN IMMEDIATE`, so the failure can happen mid-transaction with a confusing error. Fix: switch `issue()` to `BEGIN IMMEDIATE` and surface a clean error to the caller when the constraint hits.

## Why not now

Both fixes target multi-worker deploys. Pass 1 ships single-worker and the constraint + in-process lock are sufficient there. These land when worker fan-out / horizontal scaling enters scope.

hognek:
## Summary

- Adds idempotency cache with reservation pattern to prevent duplicate side effects
- Implements per-agent token store for thread-safe credential access
- Hardens concurrent agent operations against race conditions

Closes #452

## Summary by CodeRabbit

* **New Features**
  * Agent creation and deployment endpoints now support idempotent requests via `Idempotency-Key` header; duplicate requests return cached responses.
  * Added persistent agent token management with issue, revoke, and retrieval operations; enforces one active token per agent.

* **Tests**
  * Added comprehensive test coverage for idempotency caching and agent token operations.


hognek:
**Status: Blocked on #498** (fix activity endpoint flake). The concurrency hardening branch includes the activity endpoint route. Once #498 merges, this can be rebased and CI should pass. PR is ready to review.

jaylfc:
Thanks for this, the idempotency work is genuinely solid. The reservation pattern in `IdempotencyCache` is the right call: keeping `try_reserve` synchronous so it's atomic on the event loop neatly closes the get-then-set race, and putting the guard at the top of both `add_agent` and `deploy` means duplicates are deduped before any side effects. One thing to fix there: `_entries` never evicts, so the cache grows unbounded over a long-running server's lifetime. Could you add a TTL or max-size (LRU) so old keys age out?

On the agent token store half: heads-up that it overlaps work already in flight in #450, which adds a per-agent token store too (with a schema migration, REST endpoints, and bearer auth). To avoid shipping two competing token systems, could you drop the `agent_tokens_store.py` part from this PR and keep it focused on the idempotency hardening? That half is a clear win and easy to merge on its own once the cache is bounded.

Thanks again, this is good stuff.

jaylfc:
@coderabbitai review

jaylfc:
Superseded by https://github.com/jaylfc/taOS/pull/675 — this branch's mergeability got stuck in a false `CONFLICTING` state on GitHub (verified clean via merge-tree) which blocked CI from running. Recreated cleanly from current dev with the identical net diff.

Issue #224, opened 2026-04-18
perf(install): use btrfs/ZFS storage pool for incus to unlock CoW container clones (deploy ≤5s)


jaylfc:
## Context

#220 landed pre-built openclaw base images. Deploy time dropped from 90s → 33s — meaningful but short of the ≤15s target. The bottleneck is now isolated: \`incus launch taos-openclaw-base probe-raw\` alone takes **~19s** because incus's default storage pool uses the \`dir\` driver on ext4, so every container launch is a full 510 MB filesystem copy.

On btrfs or ZFS pools, incus uses copy-on-write (CoW) clones — \`incus launch\` becomes a sub-second metadata operation, and the per-deploy time drops to roughly **5–10s** total (just init + network-ready + install.sh config writes).

## Symptoms (measured on Pi 5 Plus during #220 verification)

\`\`\`
incus launch taos-openclaw-base probe-raw  → 19s
idmap restart cycle (containers/lxc.py)    →  3s
network-ready loop                          →  3s
install.sh config writes + systemd reload   →  4s
client polling slop                         →  4s
─────────────────────────────────────────────
TOTAL                                       → 33s
\`\`\`

## Fix

Update the taOS install/setup flow to create a btrfs (or ZFS) storage pool for incus instead of accepting the \`dir\` default.

### Approach A — btrfs (preferred for Pi)

Cleanest on Debian/Armbian where btrfs is in-tree. Either:

1. **Loopback file**: \`incus storage create default btrfs source=/var/lib/incus/storage-pools/default.img size=50GB\`. Simplest, no host-disk reformatting.
2. **Existing btrfs partition**: if the host already has btrfs mounted somewhere (e.g., \`/var\`), point incus at a subvolume. Cheaper IO, no double-journaling.

The host firewall + DOCKER-USER work already in this repo plus the \`tinyagentos-host-firewall\` units suggest we're comfortable doing systemd unit + setup-script work — same pattern fits here.

### Approach B — ZFS

Smaller adoption on ARM/Debian (license + module-build friction). Use only if btrfs is unavailable on the platform.

### Migration story

For existing installs (the Pi currently in use has the \`dir\` pool already populated with mary/naira/stanley):

1. Detect current pool driver in setup
2. If \`dir\` and no agents present → just recreate as btrfs
[...]

PR #502, opened 2026-05-31
feat(#224): btrfs/ZFS incus storage pool detection for CoW container clones

hognek:
## Problem
The installer uses incus admin init --auto. Users dont know if they get fast CoW container clones (<=5s on btrfs/ZFS) or slow dir copies (ext4/xfs).

## Solution
- Filesystem detection via stat -f / df -T fallback
- TAOS_COW_POOL env var: auto (default), btrfs, zfs, dir
- Creates explicit btrfs/ZFS pool when user opts in
- Falls back to incus admin init --auto gracefully
- Shows CoW availability in success summary
- Non-breaking: ext4/xfs installs unchanged

## Tests
23 new tests in tests/test_cow_storage_pool.py covering detection, init modes, and defaults. All pass.

## Summary by CodeRabbit

* **New Features**
  * Installer detects the host filesystem and configures Incus storage pools with Copy-on-Write (CoW) support; install-time control over CoW mode (auto/btrfs/zfs/dir) with sensible fallbacks.
  * Installation summary now reports detected filesystem, effective storage pool mode, and clone-speed guidance based on CoW availability.

* **Tests**
  * Added automated tests covering filesystem detection, storage-init flows, mode fallbacks, and install output reporting.


jaylfc:
Hey @hognek, heads up: I pushed a few commits onto this branch to clear the CodeRabbit findings so we can land it. What changed:

- `detect_cow_filesystem` was logging to stdout, which leaked into `COW_FS_TYPE` via the command substitution and broke the btrfs/zfs checks. Moved its diagnostics to stderr so only the fs type is captured.
- `TAOS_COW_POOL=dir` now actually creates a dir-backed pool, so `incus admin init --auto` can no longer override the opt-out on CoW hosts.
- The install summary now reports the effective storage mode (btrfs/zfs/dir/none) instead of just the detected filesystem, and the hint that suggested `TAOS_COW_POOL=1` now points at the valid `btrfs`/`zfs` values.
- Reworked the tests to source and exercise the real functions from `install-server.sh` rather than embedded copies (that is why the stdout issue slipped through), plus a safer `bash -n` call.

All green now (29 tests). Really solid feature, thanks for building it 🙏

Issue #462, opened 2026-05-30
bug(agents): no documented/in-UI way to reach an agent container — incus console asks for unknown login


jaylfc:
## Report

Reported by @johny-mnemonic in discussion #357: https://github.com/jaylfc/tinyagentos/discussions/357#discussioncomment-17001193

With the WS Container-shell shortcut broken (see companion issue), the user fell back to `incus console <container>` on the host — but it prompts for a **login/password they were never given**, so there's no way to reach the agent container at all.

## Notes

- Per the project principle that end-users should never need terminal commands, the primary fix is the in-UI Container shell (companion WS issue). This issue tracks the fallback gap: there's no documented credential or `incus exec`-based shell path either.

## Acceptance

- A working in-UI shell into the agent container (primary), and
- A documented host-side fallback (`incus exec <c> -- bash` or known creds) for advanced users.

PR #505, opened 2026-05-31
feat(#462): container shell page — Pico CSS + htmx terminal for agent containers

hognek:
## What

Container shell page at `/api/container-shell/{agent_id}` — a browser-based terminal for agent containers using `incus exec`.

- **GET /api/container-shell/{agent_id}**: HTML terminal page with dark theme, Pico CSS styling
- **POST /api/container-shell/{agent_id}/exec**: execute commands via incus exec, returns HTML fragments via htmx
- ARIA-annotated for screen readers, auto-scrolling output log
- Strips ANSI escape sequences for safe display
- **23 tests** covering auth, HTML escaping, terminal output, incus flow

## Why

This is a clean recreation of the closed PR #501. The original PR's fork had unrelated git history, making it unmergeable. This branch is built directly from upstream master and contains only the container shell feature.

## Testing

```
pytest tests/routes/test_container_shell.py -v
```

No JavaScript bundling required — uses Pico CSS (local) and htmx (CDN).

### DIFF
--- a/tests/routes/test_container_shell.py
+++ b/tests/routes/test_container_shell.py
@@ -1,0 +1,305 @@
+"""Tests for the container shell route (issue #462)."""
+
+from __future__ import annotations
+
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+
+# ── GET /api/container-shell/{agent_id} (HTML page) ─────────────────────────
+
+
+class TestContainerShellPage:
+    """Tests for the container shell HTML page endpoint."""
+
+    def test_page_returns_html(self, test_client, admin_auth_headers):
+        """GET /api/container-shell/{agent_id} returns HTML with correct content-type."""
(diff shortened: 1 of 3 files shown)


jaylfc:
Hey @hognek, thanks for this, and a quick heads up before you sink more time in. taOS has moved fully off the server-rendered htmx/Pico tier onto the React desktop SPA, so a standalone htmx page would be the last thing left in the old stack and wouldn't match the rest of the OS. The feature itself, a web shell into an agent's container, is something we want. The natural home for it is the existing React Terminal app, which already does local and SSH over a `/ws/terminal` PTY backend, so a "container" mode there would reuse all of it. If you're up for it that'd be a great contribution and I'm happy to point you at the terminal code. Going to close this approach for now, but really appreciate you taking it on.

Issue #370, opened 2026-05-07
epic(install): install-hardening pass — installer leaves user in fully working state


jaylfc:
Umbrella issue tracking the install-hardening principle ([memory note](https://github.com/jaylfc/tinyagentos/discussions/312)).

The bar: **install script finishes, taOS works fully without the user having to `apt install` or otherwise intervene.** Today's install gets the user to a running service but leaves multiple silent failures depending on platform and what nice-to-haves are missing.

## Concrete gaps already filed

- #354 — runtime Vulkan detection misses iGPUs without `vulkan-tools` installed (install-server.sh should pull `vulkan-tools` + `mesa-vulkan-drivers` when an Intel iGPU is detected)
- #359 — install-server.sh detects RKNPU but never runs install-rknpu.sh after clone
- #360 — librknnrt version detection returns empty after successful download
- #361 — RK3588 Performance mode toggle (governor + taskset baked into the rk-llama.cpp service)
- (this PR series) — `hardware.json` cache prevents re-probe; disk quota monitor "still work" lie; container backend unconfigured RuntimeError

## Pattern

Every one of those is the same shape: taOS detects something but doesn't fully wire it up, OR depends on a host-side dependency we should be bundling, OR has a "still works" fallback that doesn't actually fall back.

## Acceptance for this epic

Install on each supported target should produce a system that:

1. Has every dependency taOS uses already present (no manual `apt install` step)
2. Has every nice-to-have for capability detection installed (vulkan-tools, rocm-smi, nvidia-smi, etc. for the relevant accelerator)
3. Has the right container backend (Incus by default on Linux; Docker on macOS) auto-installed
4. Has prebuilt binaries fetched and verified for the target SoC (RK3588 librknnrt + rkllama, Apple Silicon MLX builds, etc.)
5. Reports "ready" only when every claimed capability has been verified to actually work — not just detected on paper

## Per-platform audit

Need to audit and confirm per platform:

- **Orange Pi 5 / RK3588** — librknnrt + rkllama + rk-llama.cpp + RK3588 perf mode (#361 covers the perf piece). Prebuilt binaries on the HF mirror.
- **Raspberry Pi 5 / BCM2712** — currently NOT supported. Either add support or be explicit that the install script declines.
- **x86 Linux + Intel iGPU** — vulkan-tools + mesa-vulkan-drivers + Incus. Auto-install all three.
[...]

Issue #361, opened 2026-05-07
feat(rk3588): Performance mode toggle in Settings (RK3588-only) for full-throttle inference


jaylfc:
Surfaced from issue #2 by @redkjuegos, who hit the friction first-hand on Orange Pi 5.

rk-llama.cpp on RK3588 hardware silently degrades to unusable performance after a fresh boot unless a sequence of devfreq governors and process tunings are applied first. @redkjuegos's known-good recipe:

\`\`\`bash
sudo bash -c '
  for cpu in /sys/bus/cpu/devices/cpu[0-7]/cpufreq/scaling_governor; do echo performance > $cpu; done
  echo performance > /sys/class/devfreq/fb000000.gpu/governor
  echo performance > /sys/devices/platform/dmc/devfreq/dmc/governor
  echo performance > /sys/class/devfreq/fdab0000.npu/governor
'
ulimit -n 65536

cd ~/rk-llama.cpp/build/bin && \\
  HYBRID_PATTERN=INT8_STANDARD taskset -c 4-7 ./llama-server \\
    -m ~/models/gemma-4-E2B-it-Q8_0.gguf \\
    --n-gpu-layers 999 -c 8192 --batch-size 512 --threads 4 \\
    --host 0.0.0.0 --port 8080 --keep 2048 \\
    --reasoning-budget 0 --reasoning off --jinja \\
    -ctk q8_0 -ctv q8_0 --temp 0.7 --repeat-penalty 1.1
\`\`\`

Without these settings rk-llama.cpp won't work properly after every restart.

## Scope

**RK3588 hardware ONLY**. The devfreq paths (\`/sys/class/devfreq/fb000000.gpu\`, \`fdab0000.npu\`, \`/sys/devices/platform/dmc/devfreq/dmc\`), the \`taskset -c 4-7\` core layout (assumes RK3588's 4×A55 + 4×A76 with the A76s as cores 4-7), and \`HYBRID_PATTERN=INT8_STANDARD\` are all SoC-specific. They would be wrong (or simply absent) on a Raspberry Pi 5 (BCM2712), an x86 mini-PC, an Apple Silicon Mac, or any other supported target.

The toggle must therefore be **gated on RK3588 detection** in the UI — invisible / not applicable on other hardware.

## Behaviour

- **Settings → Performance mode** toggle, visible only when hardware detection reports a Rockchip RK3588 SoC
- **On by default** when RK3588 is detected — the assumption is "this is dedicated AI hardware"
[...]

PR #509, opened 2026-05-31
fix(install): GPU detection tools (nvidia-utils, rocm-smi) + RK3588 Performance mode service (#370, #361)

hognek:
## Problem

Two gaps in the installer leave hardware capabilities undiscovered, even when the hardware is fully operational:

1. **GPU capability tools missing:** The runtime probe calls `nvidia-smi` and `rocm-smi` for VRAM size and capability reporting. Without these tools, users with dedicated NVIDIA or AMD GPUs see `VRAM: unknown`.

2. **RK3588 runs in powersave:** The SoC defaults to the ondemand/schedutil governor which caps the NPU at 1.2 GHz. Switching to performance mode unlocks the full 2.4 GHz and 6 TOPS.

### NVIDIA tools (install-server.sh)
After confirming `nv_driver && nv_devices`, install the management tools:
- **apt:** `nvidia-utils` (tools without proprietary driver)
- **dnf:** `nvidia-smi` from rpmfusion-nonfree — warns with RPM Fusion URL if not available
- **pacman:** `nvidia-utils`
- Gated behind: driver loaded + package available

### AMD tools (install-server.sh)
After confirming `amd_rocm && amd_drm`, install the monitoring tool:
- **apt:** `rocm-smi-lib`
- **dnf:** `rocm-smi`
- **pacman:** `rocm-smi-lib`
- Gated behind: kfd device + ROCm runtime + package available

### RK3588 Performance service
New systemd oneshot service `taos-rk3588-perf.service`:
- Detects RK3588 via device-tree compatible string
- Switches all CPU cores to performance governor
- Sets NPU to maximum frequency (900 MHz)
- Runs once at boot, before rkllama starts
- Wired into install script after rkllama install block

### Post-install smoke test
Verifies that after install:
- NVIDIA tools respond to `nvidia-smi` (when GPU present)
- AMD tools respond to `rocm-smi` (when GPU present)
- RK3588 perf service exists, is enabled, and governor is performance

## Verification

- `bash -n scripts/install-server.sh` — clean
- `tests/test_install_server.sh` — 20/20 tests pass (syntax, package manager detection per vendor, driver-before-tool ordering, RPM Fusion warning, RK3588 perf service detection, smoke test)

### DIFF
--- a/scripts/install-server.sh
+++ b/scripts/install-server.sh
@@ -196,8 +196,30 @@ detect_and_advise_accelerators() {
         if (( nv_driver && nv_devices )); then
             log "nvidia: kernel module loaded + device nodes present (CUDA / Vulkan available)"
             if (( ! nv_userspace )); then
-                warn "nvidia-smi is not installed — VRAM size will report as unknown to the controller"
-                warn "  optional: install nvidia-utils-XXX matching your driver version"
+                # nvidia-smi provides VRAM size and capability reporting to
+                # the runtime hardware probe. Without it, VRAM reports as
+                # 'unknown' even though the GPU is fully operational (#370).
+                if command -v apt-get >/dev/null 2>&1 && apt-cache show nvidia-utils >/dev/null 2>&1; then
+                    log "installing nvidia-utils for VRAM/capability reporting"
+                    sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq nvidia-utils
+                elif command -v dnf >/dev/null 2>&1; then
+                    # nvidia-smi lives in rpmfusion-nonfree — only install if
+                    # the user already enabled RPM Fusion. Silently skip if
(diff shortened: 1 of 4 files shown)


hognek:
@jaylfc this got a bit bigger than planned but the issues touch on the same file, so I thought it was better. Let me know if not :)

jaylfc:
@hognek No worries at all, that's the sensible call here. Since both changes are to `install-server.sh`, splitting them would just create a merge conflict between two PRs, so keeping related same-file work together is the right judgement. The thing to avoid is bundling unrelated work across different areas, but cohesive changes to the same script like this are fine. Two small asks: make the PR title/description cover both parts (GPU detection + the RK3588 perf service) so it's clear what's in it, and keeping it draft until CI's green is perfect. Thanks for flagging it.

Issue #223, opened 2026-04-18
perf(agents): shrink pre-built base image + make ensure_image_present opt-in/background


jaylfc:
## Context

Issue #220 landed pre-built LXC base images (matrix arm64 + x64, published to the \`rolling-images\` Release tag). Deploy time after the image is imported drops from 60–90s to ≤15s.

But on **first run**, fetching the ~300–500 MB image plus the ~10s import is roughly a wash with the current per-deploy build (~60–90s) on a typical home WiFi connection. For users who only ever deploy one agent, the image path can actually be slower than the build path.

### A. Shrink the published image

Current image likely includes:

- Full apt cache (\`/var/cache/apt/archives\`) — 20–50 MB recoverable
- Locale data (\`/usr/share/locale/*\` minus C/POSIX) — 30–80 MB recoverable
- npm install with devDependencies — 50–150 MB recoverable via \`npm prune --production\`
- Doc directories (\`/usr/share/doc\`, \`/usr/share/man\`) — 20–40 MB recoverable
- Possibly redundant Node.js binaries

In the build workflow, before \`incus publish\`:

\`\`\`bash
incus exec build-base -- bash -c '
  apt-get clean
  rm -rf /var/cache/apt/archives /var/lib/apt/lists/*
  rm -rf /usr/share/doc /usr/share/man /usr/share/locale/!(C|POSIX|en_US.utf8)
  cd /opt/openclaw && npm prune --omit=dev --omit=optional || true
'
\`\`\`

Target: ≤200 MB compressed per arch (down from ~300–500). At ≤200 MB, even a 50 Mbps home connection downloads in ~30s, and the import step is the same regardless of size.

### B. Make \`ensure_image_present\` non-blocking on first boot

Current behaviour: \`tinyagentos/app.py\` startup hook calls \`ensure_image_present\` which downloads the image synchronously. If the user is doing a fresh install and just wants to see the UI, they're held up for a 300+ MB download before \`/api\` even responds healthy.

Options (pick one):

1. **Background** — kick off the download as a background asyncio task; return startup-healthy immediately. Deployer's existing fallback to per-deploy build already covers the window before the image lands. First deploy ever may be slow (build), subsequent deploys are fast.
2. **Opt-in** — only run \`ensure_image_present\` when the user explicitly clicks \"Download fast-deploy image\" in the providers / agents settings UI. Until they opt in, every deploy uses the build path.
[...]

PR #512, opened 2026-05-31
feat(#223): make base image prefetch opt-in with progress tracking

hognek:
## Problem

The controller currently downloads a ~300-500MB pre-built agent base image at every startup — even for users who never deploy agents or prefer the per-deploy build path. This wastes bandwidth and disk I/O on every boot.

## Fix

Make the base image prefetch **opt-in** behind `TAOS_PREFETCH_BASE_IMAGE=1` (or `true`/`yes`). Default is disabled — no download unless the user explicitly opts in.

### Changes

- **agent_image.py:** `is_prefetch_enabled()` reads the env var. `_prefetch_state` dict tracks download progress through `idle → downloading → importing → done/failed`. `register_prefetch_endpoint()` exposes `/api/agent-image/status` for the frontend to poll.
- **app.py:** Startup `ensure_image_present()` call is gated behind `is_prefetch_enabled()`. When disabled, a debug-level log reminds users how to opt in.
- **install-server.sh:** `TAOS_PREFETCH_BASE_IMAGE` is documented, resolved at install time (defaults to 0), and substituted into the systemd service template via sed.
- **scripts/systemd/tinyagentos.service:** New `TAOS_PREFETCH` placeholder that install-server.sh replaces with 0 or 1.

### No behavior change for existing users

Users who never set `TAOS_PREFETCH_BASE_IMAGE` see no difference — the image simply will not download (they were likely not relying on it anyway). Users who want the cache can add `TAOS_PREFETCH_BASE_IMAGE=1` to their environment or re-run the installer. The deployer still falls back to `images:debian/bookworm` in all cases.

## Verification

- `tests/test_agent_image_prefetch.py` — 14 new tests (env var parsing, state tracking, service template checks)
- Existing `tests/test_agent_image.py` — 10/10 unaffected
- `bash -n scripts/install-server.sh` — clean

## Summary by CodeRabbit

* **New Features**
  * Added optional base image prefetch at startup controlled by environment configuration (disabled by default)
  * New API endpoint for monitoring prefetch progress and current status

* **Tests**
  * Added comprehensive test coverage for prefetch opt-in behavior, state tracking, and configuration validation


jaylfc:
@coderabbitai review

PR #535, opened 2026-06-03
fix(themes): block path traversal in theme asset endpoint

jaylfc:
**Security fix** (flagged by automated push review; MEDIUM path traversal).

`GET /api/themes/{theme_id}/assets/{path:path}` interpolated an unvalidated `theme_id` into a filesystem path. A traversing `theme_id` (e.g. `../../..`) could relocate the assets root outside the themes directory, and the original guard only checked the requested file against that **already-relocated** root, so it could serve files from an arbitrary `<dir>/assets/` location.

Fix:
- Validate `theme_id` against `[A-Za-z0-9_-]+` (a plain slug) before using it in any path.
- Resolve the assets root and require `target.is_relative_to(asset_root)`, so neither `theme_id` nor the `{path:path}` segment can escape.
- Added the same slug guard to the DELETE handler (defense-in-depth). The install path (`extract_theme_package`) was already guarded against unsafe ids + zip-slip.

+1 regression test: legit asset served (200), non-slug `theme_id` rejected (404), percent-encoded `../` traversal rejected (404).

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened theme asset serving with stricter theme ID validation that rejects invalid formats.
  * Enhanced path traversal protection to prevent unauthorized access to theme assets outside their designated directory.

* **Tests**
  * Added comprehensive test coverage for theme asset serving, including path traversal attack prevention and ID validation scenarios.

### DIFF
--- a/tests/themes/test_routes.py
+++ b/tests/themes/test_routes.py
@@ -24,3 +24,35 @@ async def test_install_rejects_bad_config(client):
     r = await client.post("/api/themes/install",
                           files={"package": ("b.taostheme", _zip(bad), "application/zip")})
     assert r.status_code == 400
+
+
+def _zip_with_asset(m, asset_path, asset_bytes):
+    buf = io.BytesIO()
+    with zipfile.ZipFile(buf, "w") as z:
+        z.writestr("theme.yaml", yaml.safe_dump(m))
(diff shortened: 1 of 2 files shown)


jaylfc:
@coderabbitai review

PR #628, opened 2026-06-05
fix: surface silent failures (store install, agent deploy, container backend)

jaylfc:
Three beta-blocking bugs where a novice user would get a blank/false-success/raw-traceback instead of a clear message, surfaced in a no-silent-failure audit.

## What was fixed

**1. Store install silently marks broken Docker app as installed (`store_install.py`)**
After `docker pull` succeeded, if `docker compose up` (or `.start()`) failed, the code only logged a warning and continued — recording the app as installed in the DB with a runtime location. Now a compose-up failure returns HTTP 500 with `{"ok": false, "error": "App container failed to start: <detail>. Port conflict or image error — check logs."}` and the install is aborted before any DB write.

**2. Background agent deploy left users stuck on "deploying" with no notice (`agents.py`)**
When the background deploy task set `status = "failed"`, it wrote the task dict but never surfaced anything in the notification tray. The user had no way to know a deploy had failed without polling the status endpoint. Now both failure branches (result `success: false` and unhandled exception) also call `notifications.add(level="error")` with the agent name and error detail, matching the existing pattern used in `system.py`.

**3. `get_backend()` RuntimeError in shortcut_proxy terminal WS gave a generic close reason (`shortcut_proxy.py`)**
When no container backend (Incus / Docker) is available, `_get_container_pty` raises `RuntimeError`. This was caught by the broad `except Exception` handler which sent close code 1011 with reason "PTY spawn failed" — not actionable. A specific `except RuntimeError` clause now sends "No container backend available. Install Incus or Docker and restart taOS." before the generic fallback.

## Tests

- `tests/test_routes_store.py` — two new cases: compose-up returns `{"success": False}`, and compose-up raises `RuntimeError`; both assert HTTP 500 + `ok: false` body + no runtime location recorded
- `tests/test_deployer.py` — new `test_deploy_failure_emits_notification`: patches `deploy_agent` to return failure, yields to the event loop, asserts a notification with `level=error` was pushed

All 115 tests in the three affected files pass.

## Release Notes

* **Bug Fixes**
  * Failed agent deployments now emit error notifications with failure details
  * Service installation failures return detailed error responses instead of silent failures
  * Terminal shortcuts display clear error messages when container backends are unavailable
  * Enhanced error reporting throughout deployment and installation flows

* **Tests**
  * Added deployment failure notification tests
  * Added service installation failure regression tests

### DIFF
--- a/tests/test_deployer.py
+++ b/tests/test_deployer.py
@@ -953,6 +953,30 @@ class TestBackgroundDeploy:
         resp = await client.get("/api/agents/nonexistent/deploy-status")
         assert resp.status_code == 404
 
+    @pytest.mark.asyncio
+    async def test_deploy_failure_emits_notification(self, client, app):
+        """When background deploy fails, a notification must be pushed so the
+        tray surfaces the error rather than leaving the user stuck on 'deploying'."""
+        import asyncio
+
+        with patch(
+            "tinyagentos.deployer.deploy_agent",
+            new_callable=AsyncMock,
+            return_value={"success": False, "error": "container create failed"},
+        ):
+            await client.post("/api/agents/deploy", json={
+                "name": "notif-fail-test",
+                "framework": "none",
+            })
+            # Yield to the event loop so the background task runs to completion.
+            await asyncio.sleep(0.1)
+
+        notifs = await app.state.notifications.list(limit=20)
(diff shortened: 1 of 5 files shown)


jaylfc:
@coderabbitai review

jaylfc:
@coderabbitai review

PR #667, opened 2026-06-07
fix(security): close command injection in agent_bridge + lazy_backend_proxy (closes #638, #643)

### LINKED ISSUE #638: Security: command injection in agent_bridge /keyboard and /type
## Context

The agent bridge (`agent_bridge.py`) has two endpoints that interpolate user input directly into shell commands via `create_subprocess_shell`:

- `/keyboard` endpoint (line 269-270): `f"xdotool key {req.keys}"`
- `/type` endpoint (line 310-311): `f'xdotool type -- "{req.text}"'`

Both `req.keys` and `req.text` are user-supplied strings. A payload like `{"keys": "a; rm -rf /"}` would execute arbitrary commands.

## Suggested Approach

Replace `create_subprocess_shell` with `create_subprocess_exec` using explicit argument lists:

```python
proc = await asyncio.create_subprocess_exec(
    "xdotool", "key", req.keys,
    ...
)
```

This passes `req.keys` as a single argument without shell interpretation.

### LINKED ISSUE #643: Security: shell=True in lazy_backend_proxy
## Context

`LazyBackendProxy` (`lazy_backend_proxy.py:181,232`) uses `shell=True` for both start and stop commands:

```python
self._proc = subprocess.Popen(
    self._start_cmd,
    shell=True,
    ...
)
```

If `start_cmd` or `stop_cmd` are derived from user-controllable input (manifest configs, service definitions), this is a command injection vector.

## Suggested Approach

Switch to `subprocess.Popen` with a list of arguments:

```python
self._proc = subprocess.Popen(
    shlex.split(self._start_cmd),
    ...
)
```

Or better, use explicit argument lists if the commands are always known service commands.

jaylfc:
## Summary

- **#638 — `agent_bridge.py` `/keyboard` + `/type`:** Both endpoints built shell strings (`f"xdotool key {req.keys}"`, `f"xdotool type --clearmodifiers {req.text!r}"`) passed to `create_subprocess_shell`, allowing arbitrary shell execution via payloads like `{"keys": "a; rm -rf /"}`. Fixed by switching to `asyncio.create_subprocess_exec` with an explicit arg list — `req.keys` and `req.text` become single non-interpreted arguments. `--` separator added before `req.text` in the `/type` call to guard against text starting with `-`.

- **#643 — `lazy_backend_proxy.py` `_ensure_backend` + `_stop_subprocess`:** Both `subprocess.Popen` calls used `shell=True` with the raw command string, making them injectable if `start_cmd`/`stop_cmd` derive from user-controlled manifest configs. Fixed by adding `import shlex` and passing `shlex.split(self._start_cmd)` / `shlex.split(self._stop_cmd)` as the arg list, dropping `shell=True`.

## Tests

- `test_keyboard_uses_exec_not_shell` — mocks `asyncio.create_subprocess_exec`, fires a `;`-containing payload, asserts argv has exactly `("xdotool", "key", "a; rm -rf /")` — three elements, payload is a literal arg.
- `test_type_uses_exec_not_shell` — same pattern for `/type`, asserts `--` separator present and `$(id); evil` appears as a single arg.
- `TestNoShellInjection::test_start_cmd_uses_arg_list_not_shell` — patches `subprocess.Popen`, triggers `_ensure_backend`, asserts `args` is a list matching `shlex.split(start_cmd)` and `shell` kwarg is absent/False.
- `TestNoShellInjection::test_stop_cmd_uses_arg_list_not_shell` — same for the stop path.

All 21 tests pass (prior 16 + 5 new).

Closes #638
Closes #643

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened subprocess execution across control endpoints to prevent shell-injection and ensure commands run with explicit argument lists.
  * Fixed screenshot, mouse, keyboard, and typing actions to avoid shell interpretation and to return clear errors when tools are missing or invalid inputs (e.g., invalid mouse button) are provided.
  * Hardened general command execution and batch file operations to disallow shell metacharacter chaining.

### DIFF
--- a/tests/test_agent_bridge.py
+++ b/tests/test_agent_bridge.py
@@ -5,6 +5,8 @@ Uses httpx ASGITransport so no real server is started.
 X11/xdotool/scrot are not available in CI, so visual tests assert error status.
 """
 
+from unittest.mock import AsyncMock, MagicMock, patch
+
 import pytest
 import pytest_asyncio
 from httpx import ASGITransport, AsyncClient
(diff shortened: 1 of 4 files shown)


jaylfc:
@coderabbitai review

PR #668, opened 2026-06-07
fix(security): per-agent scoped LiteLLM keys instead of the master key (closes #645)

### LINKED ISSUE #645: Security: LiteLLM master key injected into all agent containers
## Context

When LiteLLM runs in routing-only mode, the shared master key is injected into every agent container environment (`deployer.py:180-181`) as `OPENAI_API_KEY`. An agent that discovers this key can access the LiteLLM admin API, create/delete keys for other agents, and consume unlimited LLM credits.

## Suggested Approach

Generate per-agent scoped keys via LiteLLM `/key/generate` at container deploy time:
- `key_alias` matching the agent name
- `max_budget` or `tpm_limit` if configured
- Only the models the agent is permitted to use

The `litellm_config.py` already has the admin API calls — just needs per-agent key generation wired into the deploy flow.

jaylfc:
## The exposure

When LiteLLM ran without a Postgres database (routing-only mode), `deployer.py` fell back to injecting the shared master key (`sk-taos-master`) as `OPENAI_API_KEY` and `LITELLM_API_KEY` into every agent container. Any agent that discovered this key could hit the LiteLLM admin API, create or delete keys for other agents, and consume unlimited LLM credits without restriction.

## Fix

The master-key fallback path in `deployer.py` is removed entirely. When `create_agent_key` returns `None` for any reason — no Postgres DB configured (routing-only mode), DB unreachable, or master key mismatch — the deploy now **fails with a clear error** rather than silently falling back. The error message directs operators to configure a Postgres database so LiteLLM can issue per-agent scoped virtual keys.

The existing path that issues scoped keys (DB configured, key mint succeeds) is unchanged. The per-agent key is still scoped to only the models the agent is permitted to use (primary + fallbacks), with the key alias set to the agent name.

## Tests

- `test_fallback_to_master_key_when_create_returns_none` — replaced with `test_deploy_fails_when_key_mint_returns_none_no_db`: asserts deploy returns `success=False` with an actionable error when routing-only mode prevents key minting
- `test_master_key_never_injected_into_container_env` — new test: iterates every env var passed to `create_container` and asserts none carries the master key value; the scoped virtual key is injected instead

73 tests pass (all existing deployer + llm_proxy tests green).

closes #645

## Summary by CodeRabbit

* **Bug Fixes**
  * Deployment now fails with a clear error message when required database configuration is missing, preventing fallback to shared master keys
  * Ensured shared master keys are never injected into container environments; only scoped virtual keys are used for agent deployments

### DIFF
--- a/tests/test_deployer.py
+++ b/tests/test_deployer.py
@@ -198,11 +198,12 @@ class TestDeployAgent:
             )
 
     @pytest.mark.asyncio
-    async def test_fallback_to_master_key_when_create_returns_none(self, tmp_path):
-        """When LiteLLM runs without a Postgres DB, create_agent_key
-        returns None. The deployer must fall back to the shared master
-        key so the container still gets a non-empty LITELLM_API_KEY —
-        otherwise openclaw's gateway crashes on boot."""
+    async def test_deploy_fails_when_key_mint_returns_none_no_db(self, tmp_path):
+        """When LiteLLM runs without a Postgres DB, create_agent_key returns None.
+        The deployer must refuse to fall back to the shared master key — injecting
+        it would give the agent full admin API access and access to every other
+        agent's key. Deploy must fail with a clear error directing operators to
+        configure a Postgres database."""
         mock_proxy = MagicMock()
         mock_proxy.is_running.return_value = True
         mock_proxy.url = "http://localhost:4000"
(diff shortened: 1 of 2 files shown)


jaylfc:
@coderabbitai review

jaylfc:
@coderabbitai review

PR #670, opened 2026-06-07
fix(security): SSRF + SQL injection in cluster mesh_sync (closes #644)

### LINKED ISSUE #644: Cluster: SSRF bypass + SQL injection in mesh_sync
## Context

Two security issues in `scheduling/mesh_sync.py`:

**1. SSRF protection bypassed by default (line 81)** — `is_safe_url` returns `True` when `allow_private=True` (the default). Any peer URL is accepted, including `http://169.254.169.254/` (cloud metadata).

**2. SQL injection via column names (line 203)** — `col_names` comes from `record.keys()` received from a peer. A malicious peer could inject SQL via crafted column names.

## Suggested Approach

1. Change `is_safe_url` default to `allow_private=False`. If LAN peers are needed, allow only specific private ranges rather than all private IPs
2. Validate column names against a whitelist of known schema columns, or use `cursor.description` from the source table for legitimate column names

jaylfc:
## What

Two security vulnerabilities in `tinyagentos/scheduling/mesh_sync.py`.

### 1. SSRF bypass (line 81)

`is_safe_url` returned `True` unconditionally when `allow_private=True`, accepting any URL including `http://169.254.169.254/` (cloud metadata). The default signature was `allow_private=False` but all LAN callers passed `True`, making the check a no-op.

**Fix:** Split the blocked list into two tiers:
- `BLOCKED_NETWORKS` — always rejected: loopback (`127/8`), cloud metadata (`169.254/16`), loopback IPv6, ULA (`fc00::/7`), link-local IPv6.
- `LAN_NETWORKS` — RFC1918 ranges (`10/8`, `172.16/12`, `192.168/16`), only allowed when the caller explicitly passes `allow_private=True` for known LAN peers.

Cloud metadata and loopback are now rejected even when `allow_private=True`.

### 2. SQL injection via peer-supplied column names (line 203)

`import_delta` interpolated `record.keys()` from untrusted peer data directly into an INSERT statement. A malicious peer could craft column names like `); DROP TABLE kg_triples; --`.

**Fix:** Fetch the local table schema with `PRAGMA table_info` and build an allowlist of known columns. Any peer record containing a column not in that allowlist is rejected entirely and logged as a warning. Table name is sourced only from the `SYNCABLE_TABLES` constant. All VALUES remain parameterized (`?` placeholders).

## Tests

`tests/scheduling/test_mesh_sync_security.py` — 17 new tests:
- `is_safe_url` rejects `169.254.169.254` with both `allow_private=False` and `allow_private=True`
- `is_safe_url` rejects loopback always; rejects RFC1918 by default; allows RFC1918 only when `allow_private=True`
- `import_delta` drops records with injected column names, imports clean records, handles mixed batches, and returns 0 for unknown tables or empty schemas

closes #644

## Release Notes

* **Bug Fixes**
  * Enhanced SSRF protection to always block loopback and cloud metadata addresses; RFC1918 private networks are now conditionally blocked based on configuration.
  * Improved SQL injection defense by validating record columns against local schema before import; records with unknown columns are skipped.

* **Tests**
  * Added security regression tests for mesh_sync SSRF and SQL injection protection.

### DIFF
--- a/tests/scheduling/test_mesh_sync_security.py
+++ b/tests/scheduling/test_mesh_sync_security.py
@@ -131,6 +131,29 @@ class TestIsSafeUrl:
             assert is_safe_url("http://lan-worker.local/", allow_private=True) is True
 
 
+
+    def test_dns_timeout_causes_rejection(self):
+        """A socket.timeout during getaddrinfo must cause is_safe_url to return False."""
+        import socket as _socket
+        def _raise_timeout(host, port, *args, **kwargs):
+            raise _socket.timeout("timed out")
+
+        with patch("tinyagentos.scheduling.mesh_sync.socket.getaddrinfo", _raise_timeout):
+            assert is_safe_url("http://slow-peer.example.com/") is False
+
+    def test_dns_timeout_restores_default(self):
+        """The global socket timeout is restored after is_safe_url, even on timeout."""
+        import socket as _socket
+        _socket.setdefaulttimeout(None)
+
+        def _raise_timeout(host, port, *args, **kwargs):
+            raise _socket.timeout("timed out")
+
+        with patch("tinyagentos.scheduling.mesh_sync.socket.getaddrinfo", _raise_timeout):
+            is_safe_url("http://slow-peer.example.com/")
+
+        assert _socket.getdefaulttimeout() is None
+
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jaylfc:
@coderabbitai review

jaylfc:
@coderabbitai review

Issue #639, opened 2026-06-06
Security: service runs as root


jaylfc:
## Context

The taOS systemd service (`install.sh:168` and `tinyagentos.service`) runs as `User=root`. The web server on port 6969 therefore runs with full root privileges, amplifying the impact of any vulnerability to full system compromise.

## Suggested Approach

1. Create a dedicated `tinyagentos` system user during install
2. Grant only needed capabilities: read/write to data dir, LXC/Incus management (add to `incus` group), port binding (`AmbientCapabilities=CAP_NET_BIND_SERVICE` or bind to a high port)
3. The terminal WebSocket (once authenticated per issue #1) can use `sudo` for privileged operations
4. The `os-build/` service definition should match

This significantly reduces blast radius for every other issue.

jaylfc:
Resolved. The controller no longer runs as root. install-server.sh provisions an unprivileged system user 'taos' and the systemd unit sets User=taos; the installer itself still elevates via sudo but the resulting service does not. Verified live on the controller: systemctl show tinyagentos.service -p User returns User=taos.

PR #677, opened 2026-06-07
fix(security): run controller as non-root 'taos' user (#639)

jaylfc:
## Summary

- **Installer** (`scripts/install-server.sh`): creates a dedicated `taos` system user (idempotent), adds it to the `incus` and `docker` groups (warns instead of failing if either group is absent), substitutes `User=taos`/`Group=taos` into the rendered system unit (was `$USER`), then `chown -R taos:taos data/` (mode 0700) and `chmod 0600` the sensitive credential files after the venv/data setup.
- **Installer service template** (`scripts/systemd/tinyagentos.service`): no change needed — `TAOS_USER`/`TAOS_GROUP` placeholders were already there; the `ExecStartPre=+chmod` debugfs lines already carry the `+` root-override prefix and continue to work as-is.
- **Armbian image overlay** (`os-build/userpatches/overlay/etc/systemd/system/tinyagentos.service`): `User=root` → `User=taos`, `Group=taos` added.
- **Armbian build extension** (`os-build/userpatches/extensions/tinyagentos.sh`): creates `taos` user in the image chroot, adds to `incus`/`docker` groups (creates them if absent so the user record is correct even before packages land at first-boot), `chown -R taos:taos /opt/tinyagentos/data` (0700).
- **`tinyagentos/deployer.py`**: one-line comment only — `os.getuid()` returning the `taos` UID is correct for `raw.idmap both {uid} 0` (maps container-root to the host `taos` UID for the trace bind-mount). No behaviour change.
- **`systemd/tinyagentos-host-firewall.service`**: left as-is — it is a separate root oneshot for iptables rules and does not need to change.
- **Migration script** (`scripts/pre-beta-to-beta.sh`): run-once helper for existing users upgrading from a root-based pre-beta install to this non-root beta layout (see below).

## What was not changed

- The installer itself still runs as root (via `sudo bash`) — the drop to `taos` is in the systemd unit only.
- The `install_linux_systemd_user` path (no-sudo fallback) still uses `$USER` as before — it runs as the calling user by design.
- No auth bypass; no weakening of any security boundary.

## Root-only assumptions checked

The only genuine concern was `os.getuid()` in `deployer.py` being passed as `host_uid` to incus `raw.idmap`. Verified: `raw.idmap both {uid} 0` maps container-root to whatever UID the controller runs as. When running as `taos` (a non-zero UID) this is correct and required — it ensures the trace bind-mount directory owned by `taos` is writable by the container. No code change needed.

## Migration script — pre-beta → beta (`scripts/pre-beta-to-beta.sh`)

Existing users who ran a root-based pre-beta install (typically at `/root/tinyagentos` on the Pi) need to copy their data to the new non-root layout. Run this **after** the beta installer has been applied:

```bash
sudo bash scripts/pre-beta-to-beta.sh
```

Or with explicit paths if auto-detection is ambiguous:

```bash
sudo bash scripts/pre-beta-to-beta.sh \
  OLD_TAOS_DIR=/root/tinyagentos \
  NEW_TAOS_DIR=/opt/tinyagentos
```

Pass `--yes` to skip the interactive confirmation prompt.

**What it does (in order):**
1. Stops `tinyagentos.service`
2. Backs up `NEW/data/` to a timestamped `.tgz` if it already has content (never overwrites without a backup)
3. Copies `OLD/data/` → `NEW/data/` (`cp -a`, preserving timestamps/perms)
4. Ensures the `taos` system user exists and has `incus`/`docker` group membership (same logic as the installer)
5. `chown -R taos:taos NEW/` (whole install dir) + `chmod 0700 data/` + `chmod 0600` credential files
6. Patches the systemd unit from `User=root` → `User=taos` if the old unit is still in place
7. `systemctl daemon-reload && systemctl start tinyagentos`
8. Verifies the service is `active` and running as `taos`; prints PASS/FAIL with next steps

The OLD install is **never modified or deleted** — it is left intact for rollback.

## Updater compatibility

Two additional fixes so the in-app updater (Settings → Updates) works correctly when running as non-root `taos`:

### Fix 1 — taos owns the whole install dir, not just data/

The updater needs to write to `.git/` (git pull), `.venv/` (pip install -e .), and `static/desktop/` (npm run build). `set_data_dir_ownership()` in `install-server.sh` previously only chowned `data/`; it now chowns the entire `INSTALL_DIR` first, then tightens `data/` to 0700 and secret files to 0600 on top (order matters — restrictive perms win). The same change is mirrored in `pre-beta-to-beta.sh` and the Armbian build extension.

**Security trade-off**: `taos` owning its own code directory is the minimum required for non-root in-app self-update without a privileged helper. Full update-privilege-separation (a signed updater suid binary that verifies integrity before writing) is a post-beta hardening task.

### Fix 2 — pending-restart flag moved out of ~/.config

`restart_orchestrator.py` previously hardcoded `~/.config/taos/pending-restart.json`. The `taos` service user is created with `-M` (no home directory), so `~` does not resolve to a writable path under systemd.

Replaced the module-level constant with a `_pending_restart_path()` helper. Resolution order:
1. `$TAOS_DATA_DIR/pending-restart.json` — when the env var is set (already used by `__main__.py` for the Mac app).
2. `<install_dir>/data/pending-restart.json` — derived from `Path(__file__).parent.parent / "data"`, which is the same `PROJECT_DIR / "data"` convention used throughout `app.py`. This is the path used on every standard Linux install.
3. `~/.config/taos/pending-restart.json` — backward-compatible fallback for root-based or developer installs where `~` resolves correctly.

The systemd unit template does not need a new `Environment=TAOS_DATA_DIR=…` line: the module-location fallback (`__file__` → `../../data`) resolves to the correct `<install_dir>/data` directory automatically, because the process is always launched with `WorkingDirectory=<install_dir>` and the module lives inside that tree.

All three functions (`write_pending_restart`, `read_pending_restart`, `clear_pending_restart`) now call `_pending_restart_path()` so the path is consistent across write/read/clear.

## Pi test plan (deploy-time — validate on Pi before relying on it)

This is a deploy-time change with no automated CI coverage. Validate on the Orange Pi 5 Plus:

- [ ] **Re-run the installer** (or `sudo bash scripts/install-server.sh`) on the Pi to create the `taos` user and update the unit.
- [ ] **Verify user and groups**: `id taos` — expect `uid=<N>(taos) gid=<N>(taos) groups=...,incus,docker`.
- [ ] **Verify unit user**: `systemctl show -p User tinyagentos` — expect `User=taos`.
- [ ] **Verify service starts**: `systemctl status tinyagentos` — Active: active (running).
- [ ] **Verify process UID**: `ps -o user= -p $(systemctl show -p MainPID tinyagentos | cut -d= -f2)` — expect `taos`.
- [ ] **Install dir ownership**: `ls -la /opt/tinyagentos/` — expect `taos:taos` owner throughout; `data/` should be `drwx------`.
- [ ] **Auth/login works**: open `http://taos.local:6969`, log in.
- [ ] **Agent deploy works**: deploy a test agent — verifies incus socket access via `incus` group.
- [ ] **Traces write**: check `data/trace/<slug>/` is created and populated by the container.
- [ ] **Host-firewall unit**: `systemctl status tinyagentos-host-firewall` — still applies iptables rules (runs as root, unaffected by this change).
- [ ] **RKNPU stats**: check the taOS hardware panel shows NPU load (the `ExecStartPre=+chmod` debugfs lines run as root regardless of `User=taos`).
- [ ] **In-app updater**: trigger an update from Settings → Updates and confirm it completes without permission errors.
- [ ] **Pending-restart flag**: after an update, confirm `data/pending-restart.json` is created (not `~/.config/taos/`).
- [ ] **Migration script (Pi)**: on the Pi with an existing `/root/tinyagentos` install, run `sudo bash scripts/pre-beta-to-beta.sh --yes` and verify the 8-step output ends with `Migration PASSED`, the service is active as `taos`, and data is intact.

> **Note**: deploy-time change only. Validate on Pi before merging to dev.

## Summary by CodeRabbit

* **New Features**
  * Service now runs under a dedicated unprivileged system user for improved safety.
  * Added a migration tool to move pre-beta installs into the new layout while preserving data.

* **Security Improvements**
  * Data and credential files now receive stricter ownership and permissions to limit access.

* **Install/Upgrade**
  * Installer and upgrade workflow log and enforce the new service user and data ownership.

### DIFF
--- a/os-build/userpatches/extensions/tinyagentos.sh
+++ b/os-build/userpatches/extensions/tinyagentos.sh
@@ -40,14 +40,18 @@ function post_customize_image__tinyagentos() {
         " || display_alert "TinyAgentOS" "could not add taos to '${grp}' group" "warn"
     done
 
-    # Set ownership of the install and data directories so the 'taos' user can
-    # write to them at runtime.  The source files stay owned by root (readable
-    # by taos).  The data directory gets 0700 so only taos can enter it.
+    # Set ownership of the entire install directory so the 'taos' user can
+    # write to .git/, .venv/, and static/desktop/ during non-root in-app
+    # self-updates (git pull, pip install -e ., npm run build).
+    # Security trade-off: taos owns its own code, which is required for
+    # self-update without a root helper.  Full update-privilege-separation
+    # (a signed updater suid helper) is a post-beta hardening task.
+    # The data directory is then tightened on top so secrets stay 0700/0600.
     if [[ -d "${SDCARD}/opt/tinyagentos" ]]; then
         chroot "${SDCARD}" /bin/bash -c "
-            chown -R taos:taos /opt/tinyagentos/data 2>/dev/null || true
+            chown -R taos:taos /opt/tinyagentos 2>/dev/null || true
             chmod 0700 /opt/tinyagentos/data 2>/dev/null || true
         "
-        display_alert "TinyAgentOS" "Set /opt/tinyagentos/data ownership to taos:taos (0700)" "info"
+        display_alert "TinyAgentOS" "Set /opt/tinyagentos ownership to taos:taos; data/ → 0700" "info"
     fi
 }
--- a/os-build/userpatches/extensions/tinyagentos.sh
+++ b/os-build/userpatches/extensions/tinyagentos.sh
@@ -13,4 +13,41 @@ function post_customize_image__tinyagentos() {
     # Verify critical paths exist in the image
     [[ -d "${SDCARD}/opt/tinyagentos" ]] || \
         display_alert "TinyAgentOS" "/opt/tinyagentos missing — customize-image.sh may have failed" "warn"
+
+    # Create the dedicated 'taos' system user that the controller service runs
+    # as.  The service unit in the overlay already references User=taos; without
+    # this user present at first boot the service fails to start.
+    #
+    # chroot into the image target so useradd writes to the image's /etc/passwd
+    # rather than the build host.  -r = system account, -M = no home dir,
+    # -s /usr/sbin/nologin = non-interactive, -d = home field in passwd (not
+    # an actual directory on disk).
+    display_alert "TinyAgentOS" "Creating 'taos' system user" "info"
+    chroot "${SDCARD}" /bin/bash -c "
+        id taos >/dev/null 2>&1 && exit 0
+        useradd -r -M -s /usr/sbin/nologin -d /opt/tinyagentos taos \
+            || useradd -r -M -s /sbin/nologin -d /opt/tinyagentos taos
+    " || display_alert "TinyAgentOS" "useradd taos failed — service will not start as non-root" "warn"
+
+    # Add 'taos' to the incus and docker groups so the controller can reach
+    # those sockets without root.  These groups may not exist at image-build
+    # time (the packages are installed at first-boot or post-flash), so we
+    # create them if absent and add the user idempotently.
+    for grp in incus docker; do
+        chroot "${SDCARD}" /bin/bash -c "
+            getent group ${grp} >/dev/null 2>&1 || groupadd -r ${grp}
+            usermod -aG ${grp} taos 2>/dev/null || true
+        " || display_alert "TinyAgentOS" "could not add taos to '${grp}' group" "warn"
+    done
+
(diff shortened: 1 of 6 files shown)


jaylfc:
@coderabbitai full review

Issue #637, opened 2026-06-06
Secrets: hard-coded LiteLLM master key + XOR encryption


jaylfc:
## Context

Two findings around secret management:

**1. Hard-coded LiteLLM master key** (`litellm_config.py:29`)

The static string `TAOS_LITELLM_MASTER_KEY` is used as the master key for the LiteLLM proxy across all installations. It is written into the YAML config, exported to subprocesses, and used as a Bearer token for admin API calls. It is also injected into every agent container environment as `OPENAI_API_KEY` (`deployer.py:180-181`).

**2. XOR-based secret "encryption"** (`secrets.py:37-58`)

The `_encrypt` function XORs data with a key derived from `/etc/machine-id`. On macOS (no `/etc/machine-id`), the fallback is always `"tinyagentos-default"` — all macOS installations share the same key. No IV, no MAC, no KDF beyond a single SHA256 hash.

## Suggested Approach

1. **LiteLLM key**: Generate a random key per installation during first boot (`first_boot.py` would be a natural place)
2. **Secrets encryption**: Replace XOR with `cryptography.fernet.Fernet` (symmetric, authenticated, already a transitive dependency via `argon2-cffi` -> `cryptography`). On macOS, generate and persist a random machine ID on first run
3. Stop injecting the master key into agent containers — use per-agent scoped keys via LiteLLM `/key/generate` instead

PR #678, opened 2026-06-07
fix(security): generate per-install LiteLLM master key + real encryption for secrets (#637)

jaylfc:
## Summary

- **Fix A — master key**: Replaces the hardcoded `sk-taos-master` constant with `get_litellm_master_key(data_dir)`. The key is persisted at `<data_dir>/.litellm_master_key` (mode 0600) and generated once per install using `secrets.token_urlsafe(32)`. `LLMProxy` stores `data_dir` and calls the loader in `write_config`, `start`, and every Bearer-auth'd admin call (`/key/generate`, `/key/update`, `/key/delete`, `/key/info`, `/v1/models`) so the YAML config, `LITELLM_MASTER_KEY` env var, and HTTP auth always agree.

- **Fix B — real encryption**: Replaces the XOR obfuscation in `secrets.py` with `cryptography.Fernet`. The 32-byte key is stored at `<data_dir>/.secrets_key` (mode 0600). Migration is transparent: on first `get()` of an old XOR-ciphered value, `SecretsStore._dec()` detects the missing `fernet:` prefix, decrypts with XOR, re-encrypts with Fernet, and persists the upgrade atomically.

## New dependency
None. `cryptography` is already a transitive dependency via `pywebpush>=1.14` (listed in `pyproject.toml`).

## Migration notes
- **Master key**: Existing installs have no `.litellm_master_key` file. On next start, a key is generated and written. All callers read the loader — no manual action needed. Per-agent virtual keys in the LiteLLM Postgres DB are independent tokens and are NOT invalidated.
- **Secrets**: Old XOR-encrypted values are re-encrypted with Fernet on first read. No secrets are lost. The `.secrets_key` file is created on first `SecretsStore` use.

## Test plan
- [ ] `tests/test_litellm_master_key.py` — new: key generation, file persistence, mode 0600, cache stability, proxy consistency (env == Bearer header)
- [ ] `tests/test_secrets_fernet.py` — new: Fernet key file, round-trip, XOR migration on `get()`, migration durability on second `get()`, agent secrets
- [ ] `tests/test_secrets.py` — extended: XOR legacy path, Fernet round-trip, migration via raw `_decrypt`
- [ ] `tests/test_llm_proxy.py` — updated: `test_config_emits_master_key` uses generated key; `test_start_passes_database_url_when_set` checks prefix not literal
- [ ] `tests/test_taos_agent_chat.py` — updated: fallback key asserts against `get_litellm_master_key(tmp_path)` not the old constant
- [ ] `tests/test_deployer.py` — updated: master-key leak check uses generated key value

## Release Notes

* **Tests**
  * Added comprehensive test coverage for master key generation, persistence, and consistency
  * Added tests for encryption with automatic migration from legacy format
  * Enhanced security-related test suite across authentication and secrets management

* **Refactor**
  * Improved security with per-installation master key generation and persistent storage
  * Upgraded encryption to use modern cryptographic standards with transparent migration from legacy format

### DIFF
--- a/tests/test_agent_permitted_models.py
+++ b/tests/test_agent_permitted_models.py
@@ -19,6 +19,7 @@ class _FakeProxy:
         self.database_url = "postgres://x" if db else None
         self._running = running
         self._capture = capture  # list to append (models,) tuples
+        self._data_dir = None  # in-memory master key (no disk I/O in tests)
 
     def is_running(self):
         return self._running
--- a/tests/test_deployer.py
+++ b/tests/test_deployer.py
@@ -228,14 +228,16 @@ class TestDeployAgent:
 
     @pytest.mark.asyncio
     async def test_master_key_never_injected_into_container_env(self, tmp_path):
-        """The shared master key (sk-taos-master) must never appear in the
-        container env — not as OPENAI_API_KEY, LITELLM_API_KEY, or any other
-        variable. Only scoped per-agent virtual keys are permitted."""
-        from tinyagentos.llm_proxy import TAOS_LITELLM_MASTER_KEY
+        """The per-install master key must never appear in the container env —
+        not as OPENAI_API_KEY, LITELLM_API_KEY, or any other variable.
+        Only scoped per-agent virtual keys are permitted."""
+        from tinyagentos.litellm_config import get_litellm_master_key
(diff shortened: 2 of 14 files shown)


jaylfc:
@coderabbitai full review

PR #709, opened 2026-06-09
feat(security): registry revocation feed (GET /api/agents/registry/revoked)

jaylfc:
Adds the revocation feed @taOSmd needs to wire its bus-side token verify (follow-up to the agent registry #705).

## What
- `GET /api/agents/registry/revoked` → `{"revoked": [{canonical_id, revoked_at}, ...]}`
- Backed by `AgentRegistryStore.list_revoked()` (selects only rows with `revoked_at` set)
- Route declared **before** `/{canonical_id}` so "revoked" isn't matched as an agent id

## Why
@taOSmd picked the poll model for revocation (locked decision: registry revocation-list). The bus caches this feed and rejects any token whose `sub` (canonical_id) is in the set. Revocation is per-canonical_id — revoking an agent kills all its tokens — so no per-jti tracking is needed.

## Tests
5 new (32 pass total): store-level `list_revoked` (empty default, only-revoked, exact key shape) + route-level (empty feed, lists after delete, route-ordering regression).

### DIFF
--- a/tests/test_agent_registry.py
+++ b/tests/test_agent_registry.py
@@ -169,6 +169,32 @@ class TestAgentRegistryStore:
         finally:
             await store.close()
 
+    # -- Revocation feed -----------------------------------------------------
+
+    async def test_list_revoked_empty_by_default(self, tmp_path):
+        store = await self._make_store(tmp_path / "reg.db")
+        try:
+            await store.register(framework="openclaw", display_name="Active")
+            assert await store.list_revoked() == []
+        finally:
+            await store.close()
+
+    async def test_list_revoked_returns_only_revoked(self, tmp_path):
+        store = await self._make_store(tmp_path / "reg.db")
+        try:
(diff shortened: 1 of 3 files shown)


jaylfc:
Holding this for review alongside the registry auth model. An automated security pass flagged two HIGH items on `agent_registry.py` (#705 code): `register` trusts a body-supplied `user_id` that gets signed into the token, and `revoke`/list/get have no per-owner scoping. Both are real for multi-user but not exploitable beyond an authenticated single user today (all routes are session-gated and no per-request user object is wired up app-wide yet). The proper fix belongs with the multi-user auth-context layer, not a registry-only patch, so this revocation-feed PR waits on that decision. The `/revoked` contract itself is stable for the bus to build against.

jaylfc:
Superseded by the multi-user auth-context PR, which folds in this `/revoked` feed with proper owner-or-admin authz and fixes the registry security findings in the same change.

Issue #11, opened 2026-04-05
Multi-host fallback for inference backends


jaylfc:
Config-driven priority chain for inference:

```yaml
backends:
  - name: fedora-gpu
    url: http://fedora:7832
    priority: 1
  - name: mac-mini  
    url: http://mac:7832
    priority: 2
  - name: local-npu
    url: http://localhost:8080
    priority: 3
```

Try highest priority first, fall back on failure. Already partially implemented in config — needs the actual failover logic.

PR #710, opened 2026-06-09
feat(security): multi-user auth-context layer + registry authz (supersedes #709)

jaylfc:
Fixes the two HIGH security findings on the agent registry and lands the foundational multi-user auth-context layer. Design doc: `docs/superpowers/specs/2026-06-09-multi-user-auth-context-design.md`.

## Security fixes
- **Spoofable identity** — `register` no longer accepts `user_id` from the request body; it's derived from the authenticated session and signed into the token. `origin` is constrained to an allowlist.
- **IDOR / cross-tenant leak** — `revoke`/`get`/`list` now enforce owner-or-admin; members see only their own agents; single-entry reads use existence-hiding 404.

## Auth-context layer (reusable)
- `AuthMiddleware` now sets `request.state.user_id` / `is_admin` / `via` on every path (session → that user; local token → primary/admin; exempt/anon → None). No change to the allow/redirect/401 flow.
- New `tinyagentos/auth_context.py`: `CurrentUser`, `current_user` dependency (401), `require_owner_or_admin` (403). This is the default pattern for user-scoped resources going forward.

## Registry / bus
- `GET /api/agents/registry/pubkey` is now exempt (public) — the bus fetches it with no token.
- `GET /api/agents/registry/revoked` (folds in #709) → `{"revoked":[{canonical_id,revoked_at}]}`, admin/local-token only, declared before `/{canonical_id}`.

## Tests
45 pass (38 registry + 7 auth_context); all 145 non-e2e auth/registry tests green. Covers: body user_id ignored (token carries session user), origin allowlist (422), owner-or-admin (owner/admin allowed, stranger 403), existence-hiding 404, revoked-feed shape + route-ordering, `current_user` 401.

Supersedes #709. Other stores get user-scoping in tracked follow-ups (task #11 family).

## Summary by CodeRabbit

* **New Features**
  * Requests now carry an authenticated user context; registry register/list/read/revoke endpoints respect caller identity.
  * Registration enforces an origin allowlist.
  * New revoked-feed endpoint exposed.

* **Security & Access Control**
  * Listings scoped to owner (admins see all); non-owners get existence-hiding 404s for others' records.
  * Revocation and revoked feed restricted to authorized callers; register tokens bind to the authenticated user.

* **Tests**
  * Expanded coverage for auth helpers, owner/admin checks, origin allowlist, listing and revoked-feed behavior.

* **Documentation**
  * Added design spec describing multi-user auth/context and rollout guidance.

### DIFF
--- a/docs/superpowers/specs/2026-06-09-multi-user-auth-context-design.md
+++ b/docs/superpowers/specs/2026-06-09-multi-user-auth-context-design.md
@@ -1,0 +1,112 @@
+# Multi-User Auth-Context Layer — Design
+
+**Date:** 2026-06-09
+**Status:** Approved (design decisions locked via review); ready for implementation
+**Branch:** `feat/multi-user-auth-context`
+**Supersedes:** PR #709 (registry revocation feed) — folded in here so a single PR owns the registry route changes.
+
+## Problem
+
+A background security review flagged two HIGH findings on the agent registry (`tinyagentos/routes/agent_registry.py`, merged in #705):
+
+1. **Spoofable identity in signed token** — `register_agent` reads `user_id` (and `origin`) from the request **body** and `mint_registry_token` signs that `user_id` into the EdDSA token. A caller can self-assert any `user_id`.
+2. **IDOR / cross-tenant exposure** — `DELETE /api/agents/registry/{id}` revokes any agent with no ownership check; `GET /api/agents/registry` and `/{id}` return every tenant's entries.
+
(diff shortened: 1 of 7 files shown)


jaylfc:
Re the CodeRabbit fail-closed suggestion on the local-token path: declined after it broke 9 chat/typing tests (and the legitimate pre-onboarding script/CLI flow). A valid local token is already valid auth — the token file is 0600, so possession means same-host trust; it's not a bypass. When no primary user exists yet the token now passes through with `user_id=None`, so `current_user`-gated routes still 401 while middleware-only routes behave as before. Kept the `via='exempt'` contract fix. (commit c026c5a5)

Issue #660, opened 2026-06-06
HTTP: reuse httpx clients in health probes


jaylfc:
## Context

Two locations create a new `httpx.AsyncClient` on every call:

1. **Health probes** (`lifecycle_manager.py:153`) — New client every 2 seconds during startup polling
2. **OpenCode health check** (`opencode_runtime.py:224`) — New client every 0.5 seconds during `ensure_running`

This creates unnecessary TCP connection churn.

## Suggested Approach

Reuse a shared httpx client from `app.state`:

```python
# In lifespan startup
app.state.http_client = httpx.AsyncClient(timeout=10.0)
```

Close in lifespan shutdown. httpx connection pooling handles the rest.

jaylfc:
Fixed by #752 (merged to dev). Shared httpx client implemented in health probes.

PR #752, opened 2026-06-10
fix(perf): evict approve-lock leak, index auth_requests, reuse shared httpx client (#660)

jaylfc:
Three audit safe-cleanups with zero runtime-behavior change:

- **Fix 1 (memory leak):** Evict the per-request `asyncio.Lock` from `_approve_locks` in a `try/finally` block after `approve_auth_request` reaches a terminal decision. Without this the dict grows for the process lifetime (one entry per unique `request_id` ever approved).
- **Fix 2 (missing indexes):** Add `idx_auth_requests_status` and `idx_auth_requests_identity` to the `auth_requests` SCHEMA. `list_pending` and `count_pending_for` previously did full-table scans. The `IF NOT EXISTS` guard means `BaseStore.init()` applies them idempotently on next startup with no migration step.
- **Fix 3 (connection churn):** Replace five per-call `async with httpx.AsyncClient(timeout=...)` blocks in `user_memory` routes with `request.app.state.http_client` (the shared client already owned by the app). Timeout is forwarded on each individual `.get`/`.post` call. Affected tests updated to inject mocks via `app.state.http_client = mock_client` instead of patching the constructor.

[label: perf, audit]

### DIFF
--- a/tests/test_auth_requests.py
+++ b/tests/test_auth_requests.py
@@ -593,3 +593,20 @@ class TestAuthRequestRoutes:
         assert resp.status_code == 200
         grants = resp.json()["grants"]
         assert all(g["canonical_id"] == canonical_id for g in grants)
+
+    async def test_approve_lock_is_evicted_after_decision(self, consent_client):
+        """After approval the per-request lock entry is removed from the dict."""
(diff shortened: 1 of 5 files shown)


jaylfc:
@coderabbitai full review

Issue #737, opened 2026-06-10
Security: cluster worker register + heartbeat are unauthenticated (worker hijack)


jaylfc:
**Fact** (audit 2026-06-10): `auth_middleware.EXEMPT_PATHS` contains the bare string `/api/cluster/workers`, matched method-insensitively, so `POST /api/cluster/workers` (register, upsert-by-name) and `POST /api/cluster/heartbeat` are fully unauthenticated (`tinyagentos/auth_middleware.py:8`; `routes/cluster.py:111-144,222-238`). register accepts an attacker-controlled `url`; heartbeat updates an existing worker's advertised models/backends/capabilities by name: no shared secret. **Consequence:** any host that can reach :6969 can inject a malicious worker or re-register an existing worker name to repoint its backend URL, after which the cluster routes LLM/task traffic (prompts, possibly secrets) there. `DELETE /api/cluster/workers/{name}` is NOT exempt, confirming the GET+POST collision is the specific hole. **Fix:** gate register+heartbeat with a cluster shared-secret/HMAC and split the EXEMPT entry by method. Severity: High. Related: #640.

jaylfc:
## Decided approach: device-pairing code flow

Admin-approve, but with a proof-of-possession code so a LAN host cannot register a worker it does not control.

Flow:
1. The worker node runs the service/worker script. The script generates a short pairing code (e.g. 6-8 chars) and prints it on the worker console, and starts advertising itself (mDNS discovery + a registration intent to the controller, status pending).
2. In taOS, the worker is discovered (mDNS) or manually entered (host/url). It shows as 'pending: enter pairing code'.
3. The admin enters the code shown on the worker. The controller verifies the code against what the worker advertised (the worker holds the code; the controller confirms the admin-entered code matches), then mints/persists the worker's HMAC signing_key (the field already exists on WorkerInfo).
4. From then on, register/heartbeat (and the controller->worker calls) are authenticated with that signing_key; unpaired register/heartbeat are rejected.

Why this shape: no long-lived secret distributed in scripts; the code proves the admin has physical/console access to the worker; reuses the existing signing_key field + the consent/notification UX; mDNS discovery keeps onboarding low-friction.

Build phases (each a PR):
1. Backend: pairing-code store + endpoints (worker announces intent + code hash; admin confirms code -> mint signing_key); split EXEMPT so only GET /api/cluster/workers is public; require signing_key HMAC on register/heartbeat (reject unpaired). Tests.
[...]

PR #762, opened 2026-06-10
feat(cluster): pairing-code auth on worker register and heartbeat (#737)

jaylfc:
## Summary

- Adds `ClusterPairingStore` (SQLite via BaseStore) tracking announce/confirm/claim state per worker name, with 15-minute expiry and 5-attempt cap
- Adds `require_worker_hmac` dependency validating `X-TAOS-Worker-Name`, `X-TAOS-Timestamp` (300s skew window), and `X-TAOS-Signature` (HMAC-SHA256 over `timestamp.METHOD.path.body_hash`)
- Four new pairing endpoints: `POST /pairing/announce` (unauthenticated), `GET /pairing/pending` (admin), `POST /pairing/confirm` (admin), `POST /pairing/claim` (unauthenticated, delivers key exactly once)
- HMAC gate applied to `POST /api/cluster/workers` and `POST /api/cluster/heartbeat`; `GET /api/cluster/workers` stays public
- Middleware split: removed blanket exemptions for the two POST endpoints, replaced with method-sensitive rules following the auth-requests pattern
- App wiring: `ClusterPairingStore` constructed and wired on both the lifespan and eager (test) paths, mirroring `AuthRequestsStore`

## Migration note

Workers registered before this change will receive `401 worker_not_paired` on their next heartbeat. This is intentional: the error code is explicit so Phase 4 UX can surface a clear re-pair prompt to the user. The pairing flow is available immediately. The local controller worker registers in-process and is unaffected.

## Test plan

- 25 new tests in `tests/test_routes_cluster_pairing.py` covering the full flow, edge cases (expiry, attempt cap, duplicate claim), and all HMAC failure modes
- All 18 existing tests in `tests/test_routes_cluster.py` updated to drive register/heartbeat through `pair_worker` + `sign_worker_request` helpers; none deleted
- `tests/test_routes_auth.py` updated to assert the new session-exempt/HMAC-gated behavior

Closes nothing. Phase 1 of #737.

## Summary by CodeRabbit

* **New Features**
  * Secure cluster worker pairing (announce → confirm → claim) with one-time signing keys and HMAC-signed worker registration/heartbeat; admin view to list/manage pending pairings.
  * Worker registration surfaces KV cache quantization capabilities and supports mixed-cluster queries.

* **Tests**
  * Expanded test coverage for pairing flow, HMAC auth, kv-quant behavior, concurrency of claims, and updated pairing-based registration helpers.

### DIFF
--- a/tests/cluster/test_worker_invariants.py
+++ b/tests/cluster/test_worker_invariants.py
@@ -1,7 +1,7 @@
 """Cluster invariants for the worker-as-LXC architecture.
 
 Two checks:
-  1. One worker LXC per bare host — enforced via host_lan_ip uniqueness.
+  1. One worker LXC per bare host -- enforced via host_lan_ip uniqueness.
   2. Privileged + nesting verification at incus-enroll (separate test
      coverage in test_local_worker_enroll.py once that exists).
 
@@ -9,15 +9,17 @@ This file covers (1).
 """
 import pytest
 
+from tests.conftest import pair_and_register_worker
+
 
 @pytest.mark.asyncio
-async def test_second_worker_with_same_host_lan_ip_returns_409(client):
+async def test_second_worker_with_same_host_lan_ip_returns_409(client, app):
     payload1 = {
         "name": "worker-a",
         "url": "http://192.168.1.50:6970",
         "host_lan_ip": "192.168.1.50",
     }
-    r1 = await client.post("/api/cluster/workers", json=payload1)
+    r1 = await pair_and_register_worker(client, app, payload1)
     assert r1.status_code == 200
 
     payload2 = {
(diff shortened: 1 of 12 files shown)


jaylfc:
Review feedback addressed in 9384d3d7:

- pending + confirm now require an admin session via _require_admin (was: any authenticated session could mint a key)
- claim() is atomic: the pending-clear UPDATE is gated on confirmed = 1 and the rowcount decides the winner, so concurrent double-claims deliver the key exactly once
- list_pending() excludes confirmed and attempt-capped rows
- claim on an expired entry returns 410 with a re-announce message, attempt-capped returns 404 (was: 202 forever)
- routes no longer touch store internals; new public pairing_state() method, and the inline __import__("re") is a normal import

One suggestion deliberately not taken: counting bad HMAC signatures toward the pairing attempt cap. That cap protects the pairing code from brute force. Wiring signature failures into it would let anyone on the LAN spam garbage signatures under a worker's name and invalidate that worker's legitimate pending re-pair flow, a denial of service with no offsetting benefit since brute-forcing a 256-bit HMAC key is not a realistic threat. Signature failures stay independent of pairing state.

PR #763, opened 2026-06-10
fix: apply knowledge user_id migration to existing DBs and exit on main-server startup failure

### LINKED ISSUE #755: critical: update bricks controller, knowledge_store user_id migration never applies to existing DBs
**Fact** (live failure on a beta install, 2026-06-10): `KnowledgeStore.SCHEMA` contains `CREATE INDEX IF NOT EXISTS idx_ki_user_id ON knowledge_items(user_id)`, but databases created before the multi-user keying lack the `user_id` column. `BaseStore.init` runs `SCHEMA` via `executescript` BEFORE `MIGRATIONS` (tinyagentos/base_store.py:26-31), so init throws `sqlite3.OperationalError: no such column: user_id` on every pre-existing knowledge.db and the controller never finishes startup.

The registered migration `(1, "ALTER TABLE knowledge_items ADD COLUMN user_id ...")` (tinyagentos/knowledge_store.py:103) can never save the situation, for two reasons: it runs after the schema script that crashes, and even if it ran, `run_migrations_async` BASELINES any existing DB with no migration record at the latest version without executing the SQL (tinyagentos/db_migrations.py:177-198). Baseline-at-latest semantics make retrofit migrations (ones written for DBs that predate them) silently skip on exactly the databases that need them.

**Consequence:** every install with a knowledge.db from before the user_id change bricks on update. Combined with the half-alive process bug (filed separately) the failure is silent: systemd shows active, port 6969 is dead.

**Fix:**
1. Move `idx_ki_user_id` out of `KNOWLEDGE_SCHEMA`. Apply the column + index in a guarded `_post_init` step following the proven pattern in `agent_registry_store._migration_v1_add_status` (PRAGMA table_info check, ALTER only if missing, then CREATE INDEX IF NOT EXISTS). Drop the MIGRATIONS list (knowledge_store is the runner's only user).
2. Self-healing: the guarded step repairs already-bricked DBs on the next update, no manual surgery needed.
3. Add loud docstring warnings to db_migrations.py and BaseStore: SCHEMA must never reference columns introduced by migrations, and baseline-at-latest cannot deliver retrofit migrations. Rework of the runner semantics can be a follow-up if a store ever needs it again.
4. Tests: init against a simulated old-shape DB (table without user_id), assert column + index exist after init and a second init is a no-op; fresh-DB path unchanged.

### LINKED ISSUE #756: startup failure of the main server leaves a half-alive process on the proxy port (systemd stays active)
**Fact**: `_serve_dual_port` (tinyagentos/__main__.py:103) drives the main app and the browser-proxy origin with `asyncio.gather(main_server.serve(), proxy_server.serve())`. When the main app's lifespan raises during startup, uvicorn logs "Application startup failed. Exiting." but `Server.serve()` RETURNS instead of raising. gather then waits on the proxy server forever.

**Consequence:** the process stays alive listening only on the proxy port, systemd reports the unit active, Restart=always never triggers, and the controller is silently dead on :6969. Observed live on a beta install: 2.5 hours of "active (running)" with zero listeners on the main port. A startup crash should be a crash.

**Fix:** replace the bare gather with FIRST_COMPLETED semantics: when either server's serve() completes, cancel and await the other; exit non-zero if the main server never reached started (so systemd surfaces the failure and Restart= does its job). Factor the wait logic into a small helper testable with stub servers.

jaylfc:
## Summary

- **#755 (knowledge_store):** `KNOWLEDGE_SCHEMA` contained `CREATE INDEX IF NOT EXISTS idx_ki_user_id ON knowledge_items(user_id)`, which crashes `executescript` on any pre-existing database that lacks the `user_id` column -- because SCHEMA runs before MIGRATIONS. The registered `MIGRATIONS` entry could never help: `run_migrations_async` uses baseline-at-latest semantics, stamping pre-existing DBs at the latest version without executing the SQL.

  Fix: remove `idx_ki_user_id` from SCHEMA, drop the `MIGRATIONS` list, and add a guarded `_migration_v1_add_user_id` module-level coroutine (PRAGMA table_info check + ALTER TABLE only when the column is absent + CREATE INDEX IF NOT EXISTS). It runs at the top of `_post_init`, before anything else, so it self-heals already-bricked installs on their next update -- no manual DB surgery needed. Adds loud docstring warnings to `db_migrations.py` and `BaseStore` documenting the two footguns (SCHEMA-runs-first; baseline-at-latest skips retrofit migrations).

- **#756 (__main__):** `asyncio.gather(main_server.serve(), proxy_server.serve())` waits forever when the main server's lifespan raises on startup -- uvicorn logs the error and returns from `serve()` instead of raising, leaving the process alive on the proxy port only. systemd sees the unit as active and `Restart=` never fires.

  Fix: replace the bare gather with a `_serve_until_first_exit(main_server, proxy_server)` helper that uses `asyncio.wait(FIRST_COMPLETED)`, cancels and awaits the survivor, re-raises any exception from the completed task, and returns `bool(main_server.started)`. If `False`, `_serve_dual_port` logs a clear error and raises `SystemExit(3)` so systemd surfaces the failure.

## Tests

- `tests/test_knowledge_migration.py`: init against an old-shape DB (knowledge_items without user_id), assert column+index exist after init and a second init is a no-op; fresh-DB path also asserted.
- `tests/test_dual_port_helper.py`: stub server objects (no real sockets) cover: main-fails-to-start returns False, proxy-exits-first-with-started-main returns True, both-exit returns True, serve-exception propagates.

Fixes #755, fixes #756

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for server startup scenarios and database migration behavior

* **Bug Fixes**
  * Enhanced server startup and shutdown handling to detect initialization failures and report errors
  * Improved database schema migration reliability and robustness

* **Chores**
  * Updated internal documentation on database initialization patterns and migration design considerations

### DIFF
--- a/tests/test_dual_port_helper.py
+++ b/tests/test_dual_port_helper.py
@@ -1,0 +1,93 @@
+"""Tests for _serve_until_first_exit helper in tinyagentos.__main__."""
+from __future__ import annotations
+
+import asyncio
+
+import pytest
+
+
+class _StubServer:
+    """Minimal stand-in for uvicorn.Server."""
+
+    def __init__(self, *, started: bool = True, raise_on_serve: Exception | None = None):
+        self.started = started
+        self._raise = raise_on_serve
+        self._cancelled = False
+
+    async def serve(self) -> None:
+        await asyncio.sleep(0)  # yield once to let the other task run
+        if self._raise is not None:
+            raise self._raise
+
+
+class _HangingServer:
+    """Server whose serve() never returns (simulates a running server)."""
+
+    def __init__(self, *, started: bool = True):
+        self.started = started
+        self._task: asyncio.Task | None = None
+
+    async def serve(self) -> None:
+        try:
+            await asyncio.sleep(3600)
+        except asyncio.CancelledError:
+            raise
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+@pytest.mark.asyncio
+async def test_main_fails_to_start_returns_false():
+    """When main server exits with started=False, helper returns False.
+
+    The proxy stub's serve task must be cancelled.
+    """
+    from tinyagentos.__main__ import _serve_until_first_exit
+
+    main_server = _StubServer(started=False)
+    proxy_server = _HangingServer(started=True)
+
+    result = await _serve_until_first_exit(main_server, proxy_server)
+
+    assert result is False
+
+
+@pytest.mark.asyncio
+async def test_main_starts_and_proxy_exits_returns_true():
+    """When proxy exits first but main started=True, helper returns True."""
+    from tinyagentos.__main__ import _serve_until_first_exit
+
+    main_server = _HangingServer(started=True)
+    proxy_server = _StubServer(started=True)
+
+    result = await _serve_until_first_exit(main_server, proxy_server)
+
+    assert result is True
+
+
+@pytest.mark.asyncio
+async def test_both_exit_main_started_returns_true():
+    """Both servers exit; main.started=True -> returns True."""
+    from tinyagentos.__main__ import _serve_until_first_exit
+
+    main_server = _StubServer(started=True)
+    proxy_server = _StubServer(started=True)
+
+    result = await _serve_until_first_exit(main_server, proxy_server)
+
+    assert result is True
+
+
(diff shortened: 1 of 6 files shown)


jaylfc:
@coderabbitai full review

PR #764, opened 2026-06-10
fix(cluster): name worker backends type:port instead of embedding localhost URLs

### LINKED ISSUE #759: cluster: worker backend names embed worker-local localhost URLs, misleading in controller UI
**Fact** (spotted on a live Activity app): a remote worker's backend chip reads `ollama@http://localhost:11434`. The name is composed by the worker agent as `f"{backend_type}@{base_url}"` (tinyagentos/worker/agent.py:121) from its OWN localhost probe address, then displayed verbatim by the controller's Activity and Cluster views (the Loaded Models row renders the same name via `b.name`).

**Why it matters:** on the controller's screen, "localhost" reads as the controller itself, so a model running on a remote worker looks like it runs on the controller. The URL inside the name is the worker's local view and is not (necessarily) reachable from anywhere else; the worker's default Ollama typically binds 127.0.0.1 on the worker box.

**Routing is unaffected:** cross-host calls go through the worker agent (`worker.url`, TaskRouter), never the backend URL. But the `url` field on aggregated backend entries is worker-local too, so any future consumer that dials `backend["url"]` cross-host (Model Browser cluster filter, scheduler dispatch) would silently hit the wrong machine. That semantic should be written down.

**Fix:**
1. Worker agent: name backends `f"{backend_type}:{port}"` (e.g. `ollama:11434`, `ollama:21434`). The port keeps the bundled-vs-system distinction the URL was providing; no consumer parses the old format (checked).
2. Controller UI (ActivityApp + ClusterApp chips): normalise legacy names from not-yet-updated workers for display, strip a `@http://localhost:PORT` / `@http://127.0.0.1:PORT` suffix down to `:PORT`.
3. Document on the backend dict (worker_protocol.py / aggregate_catalog docstring): `url` is worker-local; cross-host calls must go via the worker agent.

jaylfc:
## What was misleading

Worker agent built each backend name as `type@http://localhost:PORT` (e.g. `ollama@http://localhost:11434`). On the controller screen, the embedded `localhost` reads as the controller itself, not the remote worker where the model actually runs. The URL inside the name was the worker's own local probe address -- unreachable from other hosts.

## New name shape

Workers now emit `type:PORT` (e.g. `ollama:11434`, `ollama:21434`). The port retains the bundled-vs-system distinction the URL was providing. No consumer parsed the old format.

## Legacy-name display normalisation

Added `normalizeBackendName(name: string)` in `desktop/src/lib/cluster.ts`. It rewrites `type@http://localhost:PORT` and `type@http://127.0.0.1:PORT` to `type:PORT` so workers that have not yet updated still display correctly. All three backend-name render sites in ActivityApp.tsx and ClusterApp.tsx now pass names through this helper.

## url-is-worker-local documentation

Added one-line notes in `tinyagentos/cluster/worker_protocol.py` (backends field) and the `aggregate_catalog` docstring in `tinyagentos/cluster/manager.py`: the backend `url` is the worker-local probe address; cross-host callers must route through the worker agent (`worker.url`), never the backend `url` directly.

## Tests

- `tests/test_worker.py`: extended `test_ollama_running` to assert `name == "ollama:11434"`, and added `test_backend_name_is_type_colon_port` covering dual ollama ports and asserting no `@` in any name.
- `desktop/src/lib/__tests__/cluster.test.ts`: 4 unit tests for `normalizeBackendName` covering legacy localhost, legacy 127.0.0.1, already-new-format, and untouched shapes.

Python: 21 passed. Frontend cluster test: 4 passed. `npm run build`: clean. Ruff delta: 0 new issues (8 pre-existing in touched files, unchanged).

Fixes #759

## Summary by CodeRabbit

* **New Features**
  * Backend identifiers are now consistently displayed in a standardized format across the cluster management interface for improved clarity.

* **Bug Fixes**
  * Legacy backend identifier formats are automatically normalized to ensure uniform display throughout the application and prevent inconsistencies.

### DIFF
--- a/desktop/src/apps/ActivityApp.tsx
+++ b/desktop/src/apps/ActivityApp.tsx
@@ -5,7 +5,7 @@ import {
 } from "lucide-react";
 import { Card, CardContent, Button } from "@/components/ui";
 import type { ClusterWorker } from "@/lib/cluster";
-import { workerStatus, workerHardwareSummary, workerShortIp, STATUS_PILL_CLASS, STATUS_LABEL } from "@/lib/cluster";
+import { workerStatus, workerHardwareSummary, workerShortIp, normalizeBackendName, STATUS_PILL_CLASS, STATUS_LABEL } from "@/lib/cluster";
 
 interface ActivityData {
   timestamp: number;
@@ -270,7 +270,7 @@ export function ActivityApp({ windowId: _windowId }: { windowId: string }) {
     ...clusterWorkers.flatMap((w) => {
       const out: LoadedModel[] = [];
       for (const b of w.backends ?? []) {
-        const backendName = b.name ?? b.type ?? "backend";
+        const backendName = normalizeBackendName(b.name ?? b.type ?? "backend");
         const activeList = (b as { loaded_models?: unknown }).loaded_models;
         const models = Array.isArray(activeList) ? activeList : [];
         for (const model of models) {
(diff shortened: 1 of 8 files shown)


jaylfc:
@coderabbitai full review

PR #770, opened 2026-06-11
feat(cluster): worker pairing flow and signed register/heartbeat (#737)

jaylfc:
## Summary

- Adds `tinyagentos/worker/pairing.py`: announce -> claim polling loop, HMAC signing (`sign_request_headers`), key persistence with 0600 file permissions, unambiguous 8-char pairing codes, and `default_state_dir` that respects `TAOS_WORKER_STATE_DIR`.
- Adds `tinyagentos/worker/pair.py`: the `python -m tinyagentos.worker.pair` CLI that both install scripts invoke. `--register-after` does the first signed `POST /api/cluster/workers` so the worker is registered before the incus-enroll step (which 404s for unknown workers).
- Updates `WorkerAgent`: loads signing key from `state_dir` on init; `register()` and `heartbeat()` now serialise the body to bytes, sign with `sign_request_headers`, and send via `content=` + explicit headers instead of `json=`. Logs a clear actionable error when not paired rather than silently failing.
- Updates `scripts/install-worker.sh`: replaces the unauthenticated inline `curl` register block with the Python CLI call; threads `TAOS_WORKER_STATE_DIR` into both systemd unit `Environment=` stanzas so the running agent reads the same key the pairing step wrote.
- Updates `scripts/install-worker.ps1`: mirrors the same pairing CLI call and injects `TAOS_WORKER_STATE_DIR` into the scheduled task.

## Design decision: crypto in Python, not shell

Both scripts invoke the same `python -m tinyagentos.worker.pair` CLI rather than reimplementing HMAC-SHA256 and the announce/claim poll loop independently in bash and PowerShell. The one security-critical path is tested once (round-trip against `require_worker_hmac` using the real app) and the shell/PowerShell delta is a single subprocess call. This keeps bash -n and shellcheck clean and avoids the two scripts drifting.

## WorkerAgent signing change

`register()` and `heartbeat()` now produce a deterministic byte string (via `json.dumps(...).encode()`), compute the HMAC over that exact body, and send it with `content=body` plus explicit headers. Using `json=` would let httpx produce a different serialisation than the one that was signed, so the signature would be computed over bytes that never get sent.

## State dir / key persistence

The default state directory is `$TAOS_WORKER_STATE_DIR` (if set), then `$XDG_STATE_HOME/taos-worker` on POSIX, or `%LOCALAPPDATA%\taos-worker` on Windows. The install scripts set `TAOS_WORKER_STATE_DIR=$INSTALL_DIR/.taos-worker-state` in the service unit so the agent always reads the key written by the pairing step.

## Test plan

- `tests/worker/test_pairing.py` (26 tests, new): sign_request_headers round-trip against `require_worker_hmac`, tampered body -> 401 bad_signature, key save/load + 0600 permissions, pairing code alphabet, `run_pairing` happy/410/timeout/already-paired paths, WorkerAgent signed header assertions
- Existing worker/cluster tests updated to inject signing keys and capture `content=` instead of `json=`
- `bash -n scripts/install-worker.sh` passes; shellcheck --severity=warning reports 0 warnings (same as baseline on dev)
- Full suite: 288 cluster+worker tests pass

Phase 2 of #737

## Summary by CodeRabbit

* **New Features**
  * Workers now pair with the controller using a displayed pairing code and persist a signing key for authenticated requests.
  * Installer and service units set an explicit worker state directory so the runtime knows where to store pairing/signing material.

* **Bug Fixes**
  * Installer stops when pairing fails or requires admin approval, and clearly instructs the user.

* **Tests**
  * Expanded tests cover pairing, signing, and updated request/heartbeat behavior.

### DIFF
--- a/scripts/install-worker.ps1
+++ b/scripts/install-worker.ps1
@@ -233,14 +233,35 @@ Log "installing worker python deps into .venv"
 & $venvPython -m pip install --quiet --upgrade pip
 & $venvPython -m pip install --quiet httpx pydantic psutil fastapi uvicorn pyyaml pillow libtorrent
 
+# --- pair + register with controller ------------------------------------
+# Pairing acquires the HMAC signing key; --register-after does the first
+# signed POST /api/cluster/workers. Crypto lives in Python, not PowerShell.
+
+$stateDir = Join-Path $InstallDir '.taos-worker-state'
+Log "pairing worker '$WorkerName' with controller at $ControllerUrl"
+Log "  (the pairing code will be printed below -- enter it in taOS > Cluster)"
+$pairArgs = @(
+    '-m', 'tinyagentos.worker.pair',
+    $ControllerUrl,
+    '--name', $WorkerName,
+    '--state-dir', $stateDir,
+    '--register-after'
+)
+& $venvPython @pairArgs
+if ($LASTEXITCODE -ne 0) {
+    Warn "pairing requires admin approval in taOS > Cluster."
+    Warn "  The code was printed above. Once approved, re-run this installer to resume."
+    exit 1
+}
+
 # --- first-boot benchmark -----------------------------------------------
 
 if (-not $SkipBenchmark) {
-    Log "running initial worker benchmark (first-join only — subsequent runs are manual)"
+    Log "running initial worker benchmark (first-join only -- subsequent runs are manual)"
     try {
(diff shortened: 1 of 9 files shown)


jaylfc:
Caught one correctness bug in review and fixed it: the pairing code was printed as a dashed pair (e.g. AB23-CD45) but the controller hashes the exact string the admin types, so entering the displayed dashed form would never match the announced hash and confirm would fail. Now prints the literal 8-char code.

PR #827, opened 2026-06-12
feat(messages): full emoji picker for reactions

jaylfc:
In `desktop/src/apps/MessagesApp.tsx` the reaction popover now hosts both the original 8-emoji quick-react row and a full picker inside the same container. The container is `w-[300px] h-[360px]` flex column with the quick row pinned to the top and the picker filling the rest. The full picker is the bare `Picker` from `emoji-picker-react` (already installed), wrapped in a div that targets the library's root class `.EmojiPickerReact` with `h-full w-full` so the picker fills the column. Selection calls the same `toggleReaction(msg.id, d.emoji)` the strip uses, so toggling on and off works the same way as before.

Used the bare `Picker` import (`Theme.DARK`, `onEmojiClick`) rather than the existing `EmojiPickerField` wrapper because the wrapper is a button plus popover and this popover is already opened by the message hover actions row. The library-level `Picker` is what the existing component renders internally.

Verified: `npx tsc -b` is clean, `npm run build` from `desktop/` succeeds (an `emoji-picker-react.esm-*.js` chunk now shows up in the bundle). No tests reference this popover.

## Summary by CodeRabbit

* **New Features**
  * Added emoji reactions to messages with an integrated emoji picker interface.

* **Style**
  * Updated emoji picker UI layout and styling for improved presentation.

### DIFF
--- a/desktop/src/apps/MessagesApp.tsx
+++ b/desktop/src/apps/MessagesApp.tsx
@@ -1916,8 +1916,11 @@ export function MessagesApp({
                       const vw = window.innerWidth;
                       const vh = window.innerHeight;
                       const r = showEmoji.rect;
-                      const top = Math.max(8, Math.min(r.top, vh - POPOVER_H - 8));
-                      const left = Math.max(8, Math.min(r.right - POPOVER_W, vw - POPOVER_W - 8));
+                      // Upper bounds are clamped to >=8 so a viewport smaller
+                      // than the popover (with margins) cannot produce a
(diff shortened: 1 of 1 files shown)


jaylfc:
Landed on dev via #833. These four Messages jobs all edited MessagesApp.tsx in overlapping regions, so a single squash-merge kept re-conflicting; they were integrated together in #833 instead. This PR's content is in commit 385ae32c (review fixes applied for 828 and 832). Closing as superseded.

PR #828, opened 2026-06-12
feat(messages): date separators in the message list

jaylfc:
In `desktop/src/apps/MessagesApp.tsx` the message render loop now inserts a centered "Today / Yesterday / weekday-day-month" divider whenever the calendar day changes between two consecutive messages. Added a module-scope `dayLabel(ts)` helper (string or number, same shape as `Message.created_at`) and a `showDaySeparator` check that compares `new Date(...).toDateString()` of the current message against the previous one. The separator is rendered before the existing message row inside a `React.Fragment` keyed by `msg.id` so React doesn't see an unkeyed child. `showAuthor` logic and the row markup are untouched.

Verified: `npx tsc -b` is clean, `npm run build` from `desktop/` succeeds. No existing tests touch this loop.

### DIFF
--- a/desktop/src/apps/MessagesApp.tsx
+++ b/desktop/src/apps/MessagesApp.tsx
@@ -255,6 +255,17 @@ function renderInline(text: string, keyPrefix: string) {
 
 const EMOJI_PICKER = ["👍", "❤️", "😂", "🎉", "🤔", "👀", "🚀", "✅"];
 
+function dayLabel(ts: string | number): string {
+  const d = new Date(ts);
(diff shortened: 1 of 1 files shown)


jaylfc:
Landed on dev via #833. These four Messages jobs all edited MessagesApp.tsx in overlapping regions, so a single squash-merge kept re-conflicting; they were integrated together in #833 instead. This PR's content is in commit cdf32cc4 (review fixes applied for 828 and 832). Closing as superseded.

PR #831, opened 2026-06-12
feat(messages): unread divider

jaylfc:
In `desktop/src/apps/MessagesApp.tsx` a red "New" divider now appears at the first unread message when a channel is opened with unread count greater than 0. The unread value is captured into a `pendingNewCountRef` immediately before `markRead` runs (so it survives the clear), and `fetchMessages` consumes that ref after the message list loads: it computes `list.length - pending`, clamps to 0, and stores the message id at that index in `newDividerAtId`. The render loop wraps each message in a `React.Fragment` and renders the divider markup (red-tinted lines, "New" label) right before the message whose id matches. `newDividerAtId` is cleared on every channel switch and after a successful send, so an outgoing message always dismisses the marker; incoming messages do not clear it.

Edge case: when the unread count exceeds the 50-message window the divider is placed before the first loaded message (clamped to index 0). When unread is 0, no divider is set.

Verified: `npx tsc -b` clean, `npm run build` from `desktop/` succeeds. Manual test plan: with two sessions, receive 2 messages in a background channel, open it, the divider sits above the 2 new messages. Switching away and back (now read) shows no divider.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a "New" divider in message lists to separate unread messages from previously read ones.
  * Divider correctly positions at the first new message after switching channels and after loading latest messages.
  * Divider is cleared after sending messages (including attachments and commands) and stays synchronized with unread counts.

### DIFF
--- a/desktop/src/apps/MessagesApp.tsx
+++ b/desktop/src/apps/MessagesApp.tsx
@@ -1728,7 +1728,11 @@ export function MessagesApp({
               return (
                 <React.Fragment key={msg.id}>
                 {newDividerAtId === msg.id && (
-                  <div className="flex items-center gap-3 my-3 select-none">
+                  <div
+                    role="separator"
+                    aria-label="New messages"
+                    className="flex items-center gap-3 my-3 select-none"
+                  >
                     <div className="flex-1 h-px bg-red-400/40" />
                     <span className="text-[11px] text-red-400 font-semibold">New</span>
                     <div className="flex-1 h-px bg-red-400/40" />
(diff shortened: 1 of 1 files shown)


jaylfc:
Landed on dev via #833. These four Messages jobs all edited MessagesApp.tsx in overlapping regions, so a single squash-merge kept re-conflicting; they were integrated together in #833 instead. This PR's content is in commit 18dd1495 (review fixes applied for 828 and 832). Closing as superseded.

PR #832, opened 2026-06-12
feat(messages): chat search UI

jaylfc:
New `SearchPanel` component and a header toolbar search button.

Search endpoint: `GET /api/chat/search?q=<query>` (also accepts an optional `channel_id` and `limit`; defaults to 20). Handler in `tinyagentos/routes/chat.py:606` returns `{ results, query }`. Each `results` row is a `chat_messages` row from `message_store.search`; fields used: `id`, `channel_id`, `author_id`, `author_type`, `content`, `created_at`.

New `SearchPanel` (`desktop/src/apps/chat/SearchPanel.tsx`): autofocused text input, 300ms debounce via `setTimeout`/`clearTimeout` in a `useEffect`, queries under 2 characters are skipped (matching the server-side guard), and an `AbortController` cancels the previous in-flight request when the query changes. `fetch` uses `credentials: "include"`. Results are grouped by `channel_id` with a small uppercase channel header (name resolved via the `channels` prop, falls back to the raw id) and each row shows the author (via `displayAuthor`), a 140-char snippet, and a relative timestamp. The toggle button uses `aria-expanded` + `aria-controls"search-panel"`; the panel renders only when `showSearch` is true and no thread is open (one-right-panel rule). `onJump` selects the channel, then scrolls the target message into view and applies the same `data-highlight` class used by the pin and deep-link paths; if the target is not in the loaded 50 messages the scroll silently no-ops (no pagination in this job). Escape closes the panel via a keydown listener inside `SearchPanel`.

Verified: `npx tsc -b` clean, `npm run build` from `desktop/` succeeds.

### DIFF
--- a/desktop/src/apps/MessagesApp.tsx
+++ b/desktop/src/apps/MessagesApp.tsx
@@ -16,6 +16,7 @@ import {
   Archive,
   Trash2,
   RotateCcw,
+  Search,
 } from "lucide-react";
 import {
   Button,
@@ -68,6 +69,7 @@ import { displayAuthor } from "./chat/format-author";
 import { useProcessStore } from "@/stores/process-store";
 import { getApp } from "@/registry/app-registry";
 import { CodeBlock } from "@/components/CodeBlock";
+import { SearchPanel } from "./chat/SearchPanel";
 
 /* ------------------------------------------------------------------ */
 /*  Types                                                              */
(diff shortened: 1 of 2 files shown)


jaylfc:
Landed on dev via #833. These four Messages jobs all edited MessagesApp.tsx in overlapping regions, so a single squash-merge kept re-conflicting; they were integrated together in #833 instead. This PR's content is in commit bebafacf (review fixes applied for 828 and 832). Closing as superseded.

PR #872, opened 2026-06-13
feat(wallpaper): live particle-network wallpaper (tsParticles, selectable)

jaylfc:
Adds a **Neural (Live)** animated wallpaper using **tsParticles** (the maintained particles.js successor) — the configurable foundation you asked for so users/agents can author wallpapers.

- Drifting node + proximity-link mesh, canvas-rendered at native resolution (any aspect ratio).
- **Theme-aware**: node/link/background colours derive from the active scheme, so it inverts with the theme.
- Pauses when the desktop is hidden (Pi perf).
- **Selectable** from the wallpaper picker; the static graphite image stays the default, so nothing breaks.
- Density / speed / colour are the next step (sliders, #21) and become the package payload for shareable live wallpapers.
- Removes the earlier rejected flat-canvas NeuralWallpaper.

## Verification
tsc + build + wallpaper tests pass. **The look needs a live check on the Pi** — tsParticles' animated canvas does not rasterize in a headless screenshot, so I couldn't preview it here (confirmed the engine runs: 200 particles, correct canvas).


jaylfc:
@coderabbitai full review

PR #874, opened 2026-06-14
feat(desktop): programmatic desktop control API (window.taosDesktop)

jaylfc:
Exposes a programmatic control surface for the desktop — the agent-tool side of the deep-navigation API (#836), and the foundation for agent-arranged layouts + deterministic screenshots/tests.

## API
- `window.taosDesktop.getLayout()` → screen size + ratio + every window's bounds/state
- `window.taosDesktop.run({ action, ... })` and a `taos:window` CustomEvent
- Actions: open, close, focus, minimize, restore, maximize, move, resize, snap, **arrange** (tile-2 / tile-3 / center / cascade)
- Targeting: explicit windowId → first window for appId → focused/topmost

All wired to the existing `process-store` actions; presets respect the work area. Docs: `docs/desktop-control.md`.

## Why
Lets the taOS agent drive the desktop, and gives me deterministic multi-window screenshots (verified: opened 3 apps + `arrange tile-3` via the API, `getLayout` returns clean tiled bounds) instead of fragile Playwright clicking.

## Verification
tsc + build clean. Drove the API live in build+preview: opened Agents/Files/Settings + tiled them; `getLayout()` returned the 3 windows at correct thirds.

## Follow-ups (tracked)
- Controller agent-tool wrapper (`desktop_get_layout` / `desktop_arrange`) + agent-manual entry so the agent can call this over the bridge.
- UI: investigate a faint double-header strip seen above the title bar on tiled windows.

## Summary by CodeRabbit

* **New Features**
  * Added programmatic desktop window management API enabling dynamic window control operations (open, close, focus, minimize, maximize, move, resize, snap, arrange).
  * Support for window arrangement presets (tile layouts, center, cascade).

* **Documentation**
  * Added comprehensive guide for the desktop control API with usage examples.

PR #877, opened 2026-06-14
feat(desktop): controller→browser command channel for agent OS control (transport)

jaylfc:
## What

Phase 1 of the **agent-OS-control transport** — the linchpin for the offline agent-driven storybook demo. Lets the controller (and through it the taOS agent) open apps and drive windows on a specific user's desktop.

The desktop subscribes to an SSE command stream and re-dispatches each command to the **existing** `taos:open-app` / `taos:window` receivers (from #874 / deep-nav), so there's no new app logic — just the missing backend→browser hop.

## How
- **`DesktopCommandBroker`** (`tinyagentos/desktop_control/broker.py`): per-`user_id` pub/sub, mirrors the canvas broker but with **no replay** — a command is a one-shot side effect; replaying buffered commands to a reconnecting desktop would re-open closed apps.
- **Routes** (`tinyagentos/routes/desktop_control.py`):
  - `GET /api/desktop/stream` — SSE, scoped to the caller's `user_id`.
  - `POST /api/desktop/command` — body `{kind, payload}`; emits to the caller's **own** desktop(s) only. Returns `{delivered: N}`.
- **Browser receiver** (`use-desktop-command-stream.ts`, mounted in `Desktop.tsx`): re-dispatches `open-app`→`taos:open-app`, `window`→`taos:window`.

## Security
`POST /command` emits only to `_user_id(request)` — a caller can only ever drive **their own** desktop; no cross-user vector. Sits behind the existing API auth/onboarding middleware.

## Tests
- 8 backend (broker fan-out, per-user isolation, **no-replay**, unsubscribe, route validation, real SSE delivery over ASGI)
- 5 frontend (subscribe, re-dispatch both kinds, malformed payload, unmount close)
- `tsc` + `vite build` clean; app constructs with both routes registered.

## Not in this PR (next)
Agent MCP tools (`open_app`/`arrange_windows` + data tools wrapping existing project/canvas/image routes) that POST commands, plus the agent-manual entry — they land **together** so the manual only advertises capabilities the agent can actually invoke. The live agent-drives-desktop **visual** is a Pi-verify (same bar as #874).

Spec: agent-OS-control + storybook demo (private).

## Release Notes

* **New Features**
  * Added agent-driven desktop control to trigger opening apps and managing windows through a live backend-to-browser command stream.
  * Introduced REST endpoints to submit commands and to subscribe to an SSE stream for real-time delivery.
  * Browser now converts streamed commands into the existing desktop event handlers.

* **Bug Fixes**
  * Malformed incoming stream messages are safely ignored.

* **Tests**
  * Added coverage for per-user isolation, no-replay behavior, bounded buffering, unsubscribe behavior, and route validation.

* **Documentation**
  * Documented the controller-to-browser command channel and provided a curl usage example.

### DIFF
--- a/desktop/src/components/Desktop.tsx
+++ b/desktop/src/components/Desktop.tsx
@@ -6,6 +6,7 @@ import { useWidgetStore } from "@/stores/widget-store";
 import { useSnapZones } from "@/hooks/use-snap-zones";
 import { useDeepNavigation } from "@/hooks/use-deep-navigation";
 import { useDesktopControl } from "@/hooks/use-desktop-control";
+import { useDesktopCommandStream } from "@/hooks/use-desktop-command-stream";
 import { getApp } from "@/registry/app-registry";
 import { Window } from "./Window";
 import { SnapOverlay } from "./SnapOverlay";
@@ -72,6 +73,10 @@ export function Desktop() {
   // runtime (lets the taOS agent drive the desktop). See the hook for details.
   useDeepNavigation(openWindow);
   useDesktopControl();
+  // Backend -> browser command channel: lets the taOS agent drive this desktop
+  // (open apps, move/arrange windows) by pushing commands the controller streams
+  // here. Re-dispatches to the taos:open-app / taos:window receivers above.
+  useDesktopCommandStream();
 
   const menuItems: MenuItem[] = [
     {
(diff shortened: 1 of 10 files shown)


jaylfc:
Addressed the review findings in 1e3bffc2:

- **CRITICAL user scoping** (Kilo + Gitar): `_user_id` now reads `request.state.user_id` (the field `AuthMiddleware` actually sets) instead of the never-populated `request.state.user`. Without this every caller collapsed to one "system" channel — a real cross-user desktop-control hole. Good catch.
- **Null JSON guard** (Kilo): the stream hook now bails on `null`/non-object before reading `.kind`.
- **Unbounded queue** (Gitar): per-subscriber queue is bounded at 128 with drop-oldest; added a test.
- **SSE buffering** (Kilo): added `Cache-Control: no-cache` + `X-Accel-Buffering: no`.
- **Docs curl** (CodeRabbit): example now shows the host + session-cookie scoping.

9 backend + 5 frontend tests green; tsc + build clean.

PR #884, opened 2026-06-14
feat(agent): agent-controlled image generation (canvas wiring + cluster tier awareness)

jaylfc:
The image half of the storybook demo: the agent generates art, picks the right tool for the cluster, and the art lands live on the project's ideas board.

**1. generate_image → canvas (the contract fix):**
- `generate_image` was b64-encoding the route's JSON as PNG bytes (garbage); now returns `image_ref` (filename) + `url`. No consumer used image_b64.
- `canvas_add_image` now takes `image_ref` and copies the workspace PNG into the project's canvas files (where `ImageShape` renders it), then creates the element. Ownership-checked.

**2. `describe_image_capabilities` (cluster tier awareness):**
- Read-only tool: the agent sees the hardware tiers (this host's NPU/GPU/CPU + cluster workers like the 3060 box) and which image tools/models each has loaded, so it picks the best model by intent (npu draft vs gpu cover). The system owns load/unload/queue.

**Split (the right separation):** agent knows the menu + chooses; system manages the plumbing. Not in scope (separate greenlight): cross-worker routing so the Pi can offload generation to the 3060 (the `TaskRouter` exists but isn't auto-invoked).

Manual + skill schemas updated; project_tools/image_tool/cluster_tools tests green (26).

## Summary by CodeRabbit

* **New Features**
  * Added `describe_image_capabilities` skill to list available image-generation backends and hardware tiers across local and cluster resources.

* **Bug Fixes**
  * Updated `canvas_add_image` to use `image_ref` (from `generate_image`) instead of `file_id` for placing generated images on project boards.

* **Documentation**
  * Updated image placement workflow documentation to reflect the new `image_ref` parameter usage.

### DIFF
--- a/docs/agent-manual/09-os-control.md
+++ b/docs/agent-manual/09-os-control.md
@@ -21,6 +21,11 @@ update the open Projects app in real time):
 - **add_task** — add a to-do task to a project's board. Args: `project_id`, `title`.
 - **canvas_add_image** — place a generated image on a project's ideas board. Args:
   `project_id`, `image_ref` (the `image_ref` returned by `generate_image`), optional `alt`.
+- **describe_image_capabilities** — see the hardware tiers (this host + any cluster
+  workers, e.g. an NVIDIA box) and which image tools/models each has loaded. Use it
+  to pick the right model before `generate_image`: an NPU model for a fast draft, a
+  GPU model for a quality cover. The system loads/unloads and queues for you — you
+  just choose the model.
 
 A typical flow: open the Projects app, create_project, add a few tasks, call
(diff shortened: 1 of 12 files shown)


jaylfc:
Addressed the review:
- **Gitar (hardware 500):** describe_image_capabilities now JSON-coerces hardware_profile fields (`_json_safe`) so a real object profile can't 500; added an object-profile test.
- **CodeRabbit/Kilo (false success):** generate_image fails instead of returning success when the scheduler response has no filename.
- **Kilo (broken file_id fallback):** dropped it; require `image_ref` (a workspace file).
- **Gitar (parity):** worker image_backends now report `loaded` too.
24 tests green.

PR #937, opened 2026-06-15
feat(providers): DeepSeek provider

jaylfc:
Adds DeepSeek as a first-class cloud model provider.

Base URL: the official https://api.deepseek.com (no /v1 suffix; DeepSeek serves the OpenAI-compatible API at the bare host). This is the official-DeepSeek path, not the ark/Volcengine gateway that Hermes happens to use locally.

Model ids seeded: deepseek-v4-pro and deepseek-v4-flash (the V4 generation), plus deepseek-chat and deepseek-reasoner (compatibility aliases, scheduled for deprecation 2026/07/24). DeepSeek's /models endpoint requires an API key, so a fresh add without a working key will not auto-discover; the seed list guarantees the provider registers routable models. With a valid key, live /models discovery replaces the seed as it does for every other cloud provider.

API key: no key is hardcoded. The provider references the user-supplied key the same way every other cloud provider does, through api_key_secret resolved to os.environ/<secret_name> in the LiteLLM config and injected from the taOS secrets store on reload. The conventional env var name is DEEPSEEK_API_KEY.

Registration: deepseek is a native LiteLLM provider, so it maps to the deepseek/ prefix in BACKEND_TYPE_MAP / CHAT_BACKEND_TYPE_MAP and LiteLLM emits deepseek/<model> entries with api_base set to the official base. Changes mirror the existing openrouter/kilocode entries exactly:

- tinyagentos/providers/__init__.py: deepseek added to ALL_TYPES, CLOUD_TYPES, and the two routing maps (prefix deepseek).
- tinyagentos/routes/providers.py: base URL default https://api.deepseek.com and a seed model list (mirrors the kilocode seed pattern).
- tinyagentos/backend_adapters.py: deepseek wired to CloudAPIAdapter (probes /models; 401 without a key reads as online).
- desktop/src/apps/ProvidersApp.tsx: DeepSeek added to FALLBACK_CLOUD_TYPES, DEFAULT_URLS, and CLOUD_PROVIDER_META so it appears in the Add Provider wizard.
- desktop/src/lib/models.ts: deepseek added to CLOUD_PROVIDER_TYPES so its models flatten into the picker.

Verification: create_app() imports clean; provider test suites (test_routes_providers, test_provider_refresh, test_provider_lifecycle_api) pass 39/39.

## Release Notes

* **New Features**
  * Added DeepSeek as a supported cloud provider with automatic model discovery and pre-configured API defaults for seamless integration.

### DIFF
--- a/desktop/src/apps/ProvidersApp.tsx
+++ b/desktop/src/apps/ProvidersApp.tsx
@@ -18,7 +18,7 @@ import { useIsMobile } from "@/hooks/use-is-mobile";
 /* ------------------------------------------------------------------ */
 
 /** Fallback constants used before the API call completes. */
-const FALLBACK_CLOUD_TYPES = ["openai", "anthropic", "openrouter", "kilocode", "openai-compatible"] as const;
+const FALLBACK_CLOUD_TYPES = ["openai", "anthropic", "openrouter", "kilocode", "deepseek", "openai-compatible"] as const;
 const FALLBACK_LOCAL_TYPES = ["rkllama", "ollama", "llama-cpp", "vllm", "exo", "mlx", "sd-cpp"] as const;
 
 /** Active cloud types — seeded with fallback, updated from /api/providers/types.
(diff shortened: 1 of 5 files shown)


jaylfc:
Re gitar's note on V4 model id routability: verified against DeepSeek's official API docs. `deepseek-v4-pro` and `deepseek-v4-flash` are the current primary model ids; the legacy `deepseek-chat`/`deepseek-reasoner` are scheduled for deprecation on 2026-07-24 (they now alias v4-flash non-thinking/thinking). Leading with the V4 ids is correct and future-proof, so keeping the order as-is. Legacy ids retained for back-compat until the deprecation date.

Issue #971, opened 2026-06-17
Userspace install source_url fetch: DNS-rebinding/TOCTOU SSRF hardening (pin validated IP)


jaylfc:
Background security review (MEDIUM). is_safe_public_url (tinyagentos/userspace/url_guard.py) resolves + validates a hostname's IPs, but install_app then re-resolves the hostname when httpx connects, leaving a DNS-rebinding / TOCTOU window where the connected IP can differ from the validated one (and bounce to an internal address).

## Fix
Resolve once in is_safe_public_url, return the validated IP(s), and pin the httpx connection to that IP (a custom transport / pinned resolver) while preserving the original Host header and SNI for TLS.

## Scheduling
This lives in install_app's source_url fetch, which the P2 distribution phase rebuilds into a proper download broker (signed-bundle fetch + sha256 + signature + min_os verification). The DNS-pinned fetch belongs there, so it is folded into P2 rather than fixed in isolation, and to avoid colliding with the in-flight P3b branch that edits the same install_app function.

Note: is_safe_public_url already blocks the obvious case (private / loopback / reserved IPs at check time); this closes the rebinding residual. Not yet on master (master is gated), so not shipping.

Part of #89 (P2 download broker).

simonlpaige:
Looking at this from a practical standpoint, the approach of folding this into P2 makes sense to avoid merge conflicts with P3b, but there's a simpler intermediate hardening you could apply right now without touching `install_app` at all.

In `url_guard.py`, after resolving and validating the IPs, you could cache the resolved IP-to-hostname mapping with a TTL equal to the DNS TTL from the original query (or a conservative default like 30s). Then add a lightweight middleware in the httpx client factory that intercepts connections and rejects any that don't match the cached mapping. This would live entirely outside `install_app` and wouldn't conflict with P3b's edits.

For the pinned resolver approach you mentioned, `httpx` supports custom transports - specifically `httpx.AsyncHTTPTransport` with a custom `resolve` callback. The key detail: you need to set `verify_mode` and pass the original hostname as the SNI hint via `ssl.SSLContext.server_hostname`, otherwise TLS verification will fail on the IP address. The `tls_extensions` callback in Python 3.12+ can also handle this more cleanly.

If you want a reference implementation, `aiohttp`'s `TCPConnector` with `resolve=async_dns_resolver` and custom `ssl_context` has been used in production for this exact pattern.

jaylfc:
[...]

PR #984, opened 2026-06-17
fix(userspace): pin install fetch to validated IP (DNS-rebind SSRF #971)

jaylfc:
## What

Closes the DNS-rebind SSRF (#971) on the install-from-URL path.

`POST /api/userspace-apps/install` validated `source_url` against private/
loopback/link-local ranges, then let httpx **re-resolve** the hostname when it
actually fetched. A host that resolved to a public address during the check
could resolve to an internal address at connect time (a classic DNS-rebind
TOCTOU), so the guard could be bypassed to hit cloud metadata, localhost, or
LAN services.

## How

- `url_guard.resolve_safe_public_ip(url)` resolves the host **once** and returns
  a validated public IP (rejects private, loopback, link-local, reserved,
  unspecified, multicast; rejects if **any** resolved address is non-public).
- The install fetch connects to that **pinned IP** and carries the original
  `Host` header + TLS `sni_hostname`, so vhost routing and certificate
  validation still work, but the client never re-resolves the name.
- `follow_redirects` stays off so a 3xx cannot bounce to a blocked host.
- `is_safe_public_url` is kept as a thin wrapper over the resolver.

Also centralises required-arg validation in the userspace broker: capabilities
that index `args` directly now return `{"error": "missing_arg", "arg": ...}`
instead of raising an uncaught `KeyError` (a 500).

## Tests

- `resolve_safe_public_ip`: public literal -> pinned IP; private/loopback/
  metadata/non-http -> None; public hostname -> validated IP; **hostname that
  resolves to a private IP -> None** (the rebind case); mixed public+private ->
  None.
- Route-level: the install fetch connects to the pinned IP (not the hostname)
  with the original Host + SNI preserved; connect failure still returns 502.
- Broker: missing required arg returns a clean `missing_arg` error.

77 userspace tests pass locally.

### DIFF
--- a/tests/userspace/test_broker.py
+++ b/tests/userspace/test_broker.py
@@ -43,6 +43,19 @@ async def test_table_capabilities(tmp_path):
 
 
 @pytest.mark.asyncio
+async def test_missing_required_arg_returns_clean_error(tmp_path):
+    # a capability that indexes args directly must return a clean error, not a 500
+    s = await _store(tmp_path)
+    out = await handle_capability("todo", "app.kv.get", {},
+                                  granted=[], data_store=s, app_dir=tmp_path / "todo", services={})
+    assert out == {"error": "missing_arg", "arg": "key"}
+    out2 = await handle_capability("todo", "app.table.delete", {"table": "t"},
+                                   granted=[], data_store=s, app_dir=tmp_path / "todo", services={})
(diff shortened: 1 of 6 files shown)


jaylfc:
@coderabbitai full review

PR #985, opened 2026-06-17
fix(update): harden docs-only classification (skip docs-commit update banner)

jaylfc:
## What

Hardens documentation-path classification for the docs-only update-suppression
feature (skips the "Update available" banner when incoming commits touch only
docs). Original logic classified anything under `docs/` (including scripts) as
docs, and only `.md` files elsewhere.

## Changes

- A `.py`/config file under `docs/` is no longer misclassified as docs-only
  (it would have suppressed a real code update). `docs/` paths are docs only
  when they do not carry a code/config extension.
- Doc extensions recognised anywhere: `.md`, `.markdown`, `.rst`.
- `.txt` is deliberately NOT a global doc extension: a root-level `.txt`
  (requirements.txt, constraints.txt) is ambiguous, and treating it as docs
  would silently suppress a real dependency update. A `.txt` under `docs/`
  still counts as docs.

## Tests

`test_is_documentation_path` covers `docs/STATUS.md`, `README.md`, `guide.rst`,
`docs/notes.txt` (docs) and `tinyagentos/foo.py`, `docs/scripts/foo.py`,
`docs/config.yaml`, `notes.txt`, `requirements.txt` (not docs). 7 pass.

Includes a reviewer commit on top of the original (gate correction).

### DIFF
--- a/tests/test_docs_only_update.py
+++ b/tests/test_docs_only_update.py
@@ -49,11 +49,15 @@ def repo(tmp_path):
 def test_is_documentation_path():
     assert is_documentation_path("docs/STATUS.md") is True
     assert is_documentation_path("README.md") is True
-    assert is_documentation_path("notes.txt") is True
     assert is_documentation_path("guide.rst") is True
+    assert is_documentation_path("docs/notes.txt") is True
     assert is_documentation_path("tinyagentos/foo.py") is False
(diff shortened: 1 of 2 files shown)


jaylfc:
Superseded: the hardening this built on is already on dev via #983. Re-landing only the .txt safety correction as a fresh PR off current dev.

Issue #112, opened 2026-04-11
Cluster-wide scheduler aggregation (v2)


jaylfc:
The cluster scheduler routes tasks based on individual worker heartbeats today — first available worker wins, no global view. Aggregating the full cluster catalog for capacity planning, bin-packing, and priority preemption across all workers is deferred to v2.

**Today's behaviour:**
- Each worker reports its live backend catalog in the heartbeat payload
- The controller stores the most recent heartbeat per worker
- When a task arrives, the scheduler walks workers in registration order and picks the first one that can serve the request
- Under load this means the cluster is rarely optimally packed; a big request can land on a worker that was about to evict a model it could have used

**What this deferral covers:**
- [ ] Global view of all workers' current memory / NPU / GPU core headroom
- [ ] Bin-packing model placement so cache locality is maximised
- [ ] Priority preemption across workers (high-priority agents can bump low-priority ones on another box)
- [ ] Cluster-wide LRU eviction driven by aggregate pressure, not per-worker pressure
- [ ] Routing hints from agent manifests (prefer worker X, avoid GPU, etc.)

**Dependencies:** Sits on top of #62 (Phase 1.5 sequential loading) and #61 (core-aware resource model). Both need to land locally first.

**Spec:** \`docs/design/resource-scheduler.md\`, plus the Phase 2 section of \`docs/superpowers/specs/2026-04-11-taos-framework-integration-bridge-design.md\`.

PR #1138, opened 2026-06-20
feat(desktop): multi-window Projects (dock New Window + per-window project)

jaylfc:
Implements Jay's ask: "if I have a project open I should be able to open a new window to view another project" + "right click on dock icon and open new window". Consumer of the forceNew `openWindow` primitive (#1109).

## Changes
- **Registry**: Projects is now `singleton: false` so multiple windows can coexist.
- **Dock**: right-clicking a multi-window app's dock icon shows **New Window** (uses `openWindow(appId, size, undefined, { forceNew: true })`). Only shown for non-singleton apps; singleton apps are unchanged.
- **ProjectsApp**: accepts a per-window `projectId` prop (threaded through `WindowContent` props) and pins that window to the project. The first/default window keeps the list behaviour.
- **Project list**: a hover affordance on each row opens that project in its own window, the natural way to bring a second project up beside the current one. Touch devices (no hover) keep it visible.

## Tests
- `ProjectList.empty-state.test.tsx`: new affordance calls `onOpenInNewWindow(id)`; absent when no handler; row select still fires `onSelect`. (6 passed)
- `tsc -b` clean; `npm run build` green; existing process-store/WindowContent tests pass.

Note: a follow-up (#112) adds the macOS-style per-app menu bar for File > New Window parity. Visual polish to be confirmed on-device.


jaylfc:
Kilo Code Review failed with 'Could not connect to the sandbox' (infra/sandbox connectivity, not a code finding). Gitar and CodeRabbit both reviewed clean. Re-requested Kilo; if it stays infra-failed, this is safe to admin-merge once the python test jobs pass.

PR #1139, opened 2026-06-20
test(project-canvas): fix DB isolation (real db_path attr)

jaylfc:
Folds the #1119 gitar must-fix finding (completes task #113 with #1133).

The autouse fixture in tests/test_routes_project_canvas.py set `store._db_path` (a non-existent attribute) as a string, but `BaseStore.init()` reads `self.db_path` (a `Path`). So the override was a silent no-op and the suite ran against the **production** canvas DB.

Fix: assign `store.db_path = tmp_dir / "test_projects.db"` (a Path), matching BaseStore.

Verified: 18 passed / 1 skipped; a probe confirms init() now targets the temp DB and never creates the production path.

### DIFF
--- a/tests/test_routes_project_canvas.py
+++ b/tests/test_routes_project_canvas.py
@@ -19,9 +19,11 @@ def _ensure_canvas_store(client, tmp_path_factory):
             asyncio.get_event_loop().run_until_complete(store.close())
         except Exception:
             pass
-    # Use a fresh DB per test session via tmp_path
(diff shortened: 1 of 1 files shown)


jaylfc:
Kilo Code Review is failing platform-wide right now with 'Could not connect to the sandbox' (infra, not a code finding) -- it is red on every open PR this cycle. Gitar and CodeRabbit reviewed clean. Safe to admin-merge once the python test jobs pass.

PR #1140, opened 2026-06-20
fix(account-proxy): trust-gate X-Forwarded-Proto + harden header relay

jaylfc:
Folds two post-merge gitar findings on #1133 (per the bot-review severity gate).

**Security** - X-Forwarded-Proto is client-spoofable unless the controller sits behind a trusted proxy. Previously it was honored unconditionally, so a client on a direct (non-relay) connection could send `X-Forwarded-Proto: https` and force the session cookie's `Secure` attr over plain http. Now gated behind `TAOS_TRUST_FORWARDED_PROTO` (the TLS-terminating taOSgo relay deployment sets it); otherwise Secure derives from the real connection scheme only.

**Edge case** - relayed header values are `encode("latin-1")`d; a non-latin-1 value would raise `UnicodeEncodeError` and break the whole response. Added `_append_header` which skips non-encodable values, applied to both the rescoped cookie and the allowlist (Location / Cache-Control / WWW-Authenticate).

Tests: trusted X-Forwarded-Proto keeps Secure; untrusted ignores it (Secure dropped); existing relay/cookie tests still pass (7 passed).

### DIFF
--- a/tests/test_routes_account_proxy.py
+++ b/tests/test_routes_account_proxy.py
@@ -99,10 +99,12 @@ async def test_account_me_503_when_upstream_unreachable(client, monkeypatch):
 
 
 @pytest.mark.asyncio
-async def test_secure_kept_when_x_forwarded_proto_https(client, monkeypatch):
+async def test_secure_kept_when_x_forwarded_proto_https_and_trusted(client, monkeypatch):
     """Behind a TLS-terminating proxy the request scheme is http but the browser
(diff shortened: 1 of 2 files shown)


jaylfc:
Kilo Code Review is failing platform-wide right now with 'Could not connect to the sandbox' (infra, not a code finding) -- it is red on every open PR this cycle. Gitar and CodeRabbit reviewed clean. Safe to admin-merge once the python test jobs pass.

PR #1298, opened 2026-06-21
feat(cluster): manual free-tier worker pairing (controller + UI + worker)

jaylfc:
## Free-tier manual worker pairing

Adds the LAN/free-tier path to add a cluster worker with **no network discovery** and **no announce** — the user manually enters the worker's IP and a PIN. Automated discovery and one-tap install stay taOSgo-only (the paid convenience tier).

### Flow (option A — worker polls for key keyed on PIN)
1. User installs the worker with `TAOS_PAIR_MANUAL=1`. The worker prints its LAN address + an 8-char PIN and starts polling.
2. In **taOS > Cluster > Add worker**, the user enters that IP + PIN.
3. The controller authorises the `{url, code}` pair and mints a signing key.
4. The worker's next poll returns the key; it persists it and registers. Done.

### Slices
- **1/3 controller** (`pairing_store.py`, `routes/cluster.py`): `cluster_manual_pairings` table (code-hash keyed, single-use, 15-min TTL); `manual_authorize` / `manual_claim`; `POST /api/cluster/pairing/manual` (admin) + `/manual-claim` (worker poll).
- **2/3 UI** (`ClusterApp.tsx`): "Add worker" button (toolbar + empty state) opens a modal for IP + PIN, POSTs to the manual endpoint.
- **3/3 worker** (`worker/pairing.py`, `worker/pair.py`, `install-worker.sh`): `run_manual_pairing` (no announce), `--manual` flag, `TAOS_PAIR_MANUAL` env plumb.

### Tests
- pairing_store: authorize/claim, single-use, wrong code, re-authorize, expiry (6).
- routes: manual + manual-claim happy/await/validation.
- worker: manual happy path, already-paired short-circuit, timeout (no announce asserted).

Security note: the manual table has no per-code attempt cap, so the PIN reuses the strong 8-char unambiguous alphabet (~7e13 space) rather than a short numeric PIN; brute force within the 15-min TTL is infeasible on a LAN.


jaylfc:
Thanks — all three findings addressed in 13944c6b:

**🚨 Bug (urlparse not imported):** confirmed and fixed. `pairing_manual` now does a local `from urllib.parse import urlparse` (matching `incus_enroll`'s style). Added a route-level test (`test_manual_authorize_then_claim_happy_path`) that drives a bare-host url through the endpoint, so this path is now covered, not just the store layer.

**⚠️ Security (no brute-force cap on manual-claim):** the subtlety is that `manual_claim` matches purely by `sha256(code)`, so a *wrong* code is indistinguishable from a worker polling *before* the admin authorises — both return "awaiting". A failure-counting cap would therefore throttle legitimate polling. Instead I added a per-IP fixed-window rate limit (20 req / 10s) on the endpoint: a real worker polls every ~3s (0.3 req/s, far under the cap), while an attacker is held to a few attempts per window. On top of the high-entropy PIN (8-char unambiguous alphabet, ~7.2e13 space) and the 15-min TTL, guessing the code is infeasible. Covered by `test_manual_claim_rate_limited`.

**💡 Cleanup (expired/abandoned rows accumulate):** switched single-use from a `consumed` flag to DELETE-on-claim, and the expired branch now deletes the row on access. No claimed or expired authorisation leaves a stale signing key at rest.

PR #1306, opened 2026-06-21
llm-proxy: in-house per-agent key store (custom_auth, no Postgres)

jaylfc:
## What

Opt-in key mode that mints per-agent virtual keys in a local SQLite store and authorizes them through LiteLLM's \`custom_auth\` hook, instead of LiteLLM's Postgres/prisma virtual-key table.

This is the durable fix for the ARM / no-Postgres deploy regression: \`prisma-client-py\` 0.15.0's query engine won't start on the Pi, so \`/key/generate\` fails and per-agent keys can't be minted. With this mode the router keeps working and per-agent keys are minted locally, with **no DATABASE_URL and no prisma**.

## How

- \`litellm_keystore.py\` — SQLite \`mint\` / \`lookup\` / \`set_models\` / \`delete\` / \`delete_agent\`. WAL + busy_timeout for cross-process controller-write / hook-read.
- \`litellm_auth.py\` — the \`custom_auth\` hook. Master key passes through (admin); otherwise looks up the token and enforces the per-agent model allowlist in-hook (401 unknown key, 403 out-of-scope model), returning \`UserAPIKeyAuth\` with the allowlist + metadata so the usage callback still attributes spend per agent.
- \`llm_proxy.py\` — \`inhouse_keys\` flag routes \`create/update/delete_agent_key\` through the keystore, writes the \`taos_auth.py\` shim next to the config, exports \`TAOS_LITELLM_KEYSTORE\`, and skips \`DATABASE_URL\` so prisma never starts.
- \`litellm_config.py\` — emits \`general_settings.custom_auth\` when in-house.
- \`app.py\` — gates the flag on a \`.litellm_inhouse_keys\` marker file.

## Safety

Behind a flag, **default off**. Current behavior (master-key fallback for routing-only installs) is unchanged unless the marker file is present. Not yet enabled on any host.

## Tests

- \`test_litellm_keystore.py\` — store round-trips + the hook (master passthrough, 401 unknown, 403 out-of-scope, in-scope success). litellm stubbed so the hook runs in CI without the container-only package.
- \`test_llm_proxy.py\` — config emits \`custom_auth\`, shim written, and create/update/delete mint locally with no DB and no running proxy.

187 of 187 targeted tests green (keystore + proxy + config + deployer).

## Release Notes

* **New Features**
  * Added an opt-in “in-house key mode” for LiteLLM authorization using TinyAgentOS’s SQLite-backed keystore.
  * Introduced a LiteLLM `custom_auth` hook that supports master-key passthrough and enforces per-agent model scoping (denies requests when a token’s scope is empty or missing the requested model).
  * Enables local per-agent API key management with model scoping; activated by creating `.litellm_inhouse_keys` in the data directory.

* **Tests**
  * Added keystore and auth-hook test coverage, plus proxy tests for in-house key mode behavior (token minting, scoping, updates, deletions, and persistence).

### DIFF
--- a/tests/test_litellm_keystore.py
+++ b/tests/test_litellm_keystore.py
@@ -139,3 +139,16 @@ async def test_hook_out_of_scope_model_403(monkeypatch, tmp_path):
     with pytest.raises(HTTPException) as exc:
         await auth.user_api_key_auth(_FakeRequest({"model": "claude-3"}), token)
     assert exc.value.status_code == 403
+
+
+@pytest.mark.asyncio
+async def test_hook_empty_allowlist_is_deny_all(monkeypatch, tmp_path):
+    """An empty allowlist must 403 every model, not fall through to allow-all
+    (which is how LiteLLM reads UserAPIKeyAuth(models=[]))."""
+    from fastapi import HTTPException
+    path = tmp_path / "keys.db"
+    token = LiteLLMKeyStore(path).mint("agent-a", None)  # stored as []
+    auth = _reload_auth(monkeypatch, path)
+    with pytest.raises(HTTPException) as exc:
+        await auth.user_api_key_auth(_FakeRequest({"model": "anything"}), token)
+    assert exc.value.status_code == 403
--- a/tests/test_litellm_keystore.py
+++ b/tests/test_litellm_keystore.py
@@ -1,0 +1,141 @@
+"""Tests for the in-house LiteLLM key store + custom_auth hook."""
+import importlib
+import sys
+import types
+
+import pytest
+
+from tinyagentos.litellm_keystore import (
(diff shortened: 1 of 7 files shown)


jaylfc:
Addressed the empty-allowlist Security finding in fcf403d0: the auth hook now treats an empty allowlist as deny-all (403s every model) instead of short-circuiting to allow-all, and the in-house mint path mirrors the Postgres `models or ["default"]` fallback so a model-less agent gets a usable scoped key rather than an empty allowlist. Follow-up 7fee9f75 reads `inhouse_keys` via getattr so the duck-typed test doubles in test_update_agent_key / test_agent_permitted_models stop failing. Regression tests added for both.

PR #1307, opened 2026-06-22
feat(agents): add DeerFlow (bytedance/deer-flow) framework support

jaylfc:
## What this adds

Support for the **deer-flow** agent framework (github.com/bytedance/deer-flow, MIT) to taOS, following the exact pattern of the existing `hermes` framework. Mechanical mirror, no redesign.

Files added:
- `app-catalog/agents/deer-flow/manifest.yaml` (agent-framework, alpha, v2.1.0, MIT, full hardware-tier matrix)
- `app-catalog/agents/deer-flow/scripts/install.sh` (idempotent Debian bookworm LXC installer)
- `app-catalog/agents/deer-flow/framework-integration.yaml` (bridge design spec, `langgraph-runs`)
- `tinyagentos/adapters/deer_flow_adapter.py` (FastAPI bridge adapter)
- registered `deer-flow` in `tinyagentos/adapters/__init__.py`
- `tests/test_deer_flow_adapter.py`

## Runtime facts (verified against deer-flow)

- Python 3.12 backend under `backend/`, provisioned with `uv` (`uv sync`).
- Backend runs API served via `uvicorn app.gateway.app:app` on **127.0.0.1:8001**.
- Chat surface is NOT OpenAI-compatible. The adapter POSTs to `http://localhost:8001/api/runs/wait` with `{"input": {"messages": [...]}, "context": {"model_name": "taos-default"}}` (600s timeout for long-horizon runs) and extracts the final assistant reply from the serialized LangGraph channel-values.
- `config.yaml` at the repo root points deer-flow's `taos-default` model at the taOS LiteLLM proxy via `$TAOS_MODEL` / `$OPENAI_API_KEY` / `$OPENAI_BASE_URL`. Sets `memory.token_counting: char` to skip the startup tiktoken cache network fetch in restricted containers. Default `LocalSandboxProvider` (no Docker) left as-is.
- systemd unit runs as root (the uv-managed venv under `/opt` keeps permissions simple; the container is the isolation boundary). Service is enabled but not started, matching hermes (deployer starts it).

## Status

This is **alpha**: happy path implemented, adapter response parsing is defensive across the plausible channel-values shapes. Needs a **live Pi deploy-test** to confirm the real `/api/runs/wait` response shape, the `uv sync` build on aarch64, and end-to-end chat through the proxy.

## Tests

Adapter and registry tests pass locally (`_extract_text` covering string content, list-of-text-blocks content, and empty/missing messages; plus health and no-backend message tests).

## Summary by CodeRabbit

* **New Features**
  * Added DeerFlow SuperAgent integration to the app catalog (alpha, v2.1.0) with model selection and capability metadata.
  * Introduced a DeerFlow adapter service with `POST /message` (bridges requests, extracts structured assistant text) and `GET /health` (framework status).
  * Added Debian-based installation and automated deployment using a systemd service with generated config and environment setup.
* **Tests**
  * Added unit and API tests validating text extraction behavior and `/health`/`/message` responses, including graceful handling when the backend is unavailable.

### DIFF
--- a/app-catalog/agents/deer-flow/framework-integration.yaml
+++ b/app-catalog/agents/deer-flow/framework-integration.yaml
@@ -1,0 +1,107 @@
+# DeerFlow SuperAgent — framework integration manifest
+# See docs/superpowers/specs/2026-04-11-taos-framework-integration-bridge-design.md
+#
+# NOTE: This is a *design specification*, not machine-processed runtime config.
+# ${VAR} tokens (e.g. ${TAOS_HOST}) are documentation placeholders indicating
+# values resolved by the TAOS deployer at provision time. They mirror the same
+# env vars the deployer injects into the container after install.sh.
+
+framework: deer-flow
+display_name: DeerFlow SuperAgent
+version: "2.1.0"
+
+# -- Protocol endpoints the adapter speaks -----------------------------------
+endpoints:
+  llm: openai-chat              # DeerFlow calls the taOS LiteLLM proxy (OpenAI-compatible)
+  memory: none                  # DeerFlow uses its own SQLite-backed persistence
+  skills: none                  # DeerFlow has built-in tools via its LangGraph nodes
+  channels: sse-bridge          # Connects to TAOS channels via the adapter bridge
+  approval: none                # Not implemented yet
+  sandbox: host-enforced        # Container-level sandbox (default LocalSandboxProvider)
+
+# -- Persistence contract -----------------------------------------------------
+persistence:
+  memory:
+    kind: sqlite-volume
+    mount: /opt/deer-flow/backend
+    host_subdir: agent-memory/{name}/deer-flow/
+    writable: true
+    note: DeerFlow persists run state to SQLite by default
+  workspace:
+    mount: /workspace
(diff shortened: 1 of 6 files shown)


jaylfc:
Folded the gitar findings:
- Bug (systemd never receives OPENAI/TAOS_MODEL env): fixed in a7c8ea3e — install.sh now snapshots the deployer-injected vars into /opt/deer-flow/deer-flow.env and the unit reads it via EnvironmentFile (systemd does not inherit the container's incus environment.* automatically). Same pattern as Hermes.
- Edge case (600s timeout retried 7x re-runs the job): 100f6a2a — retry only on connection-level errors (ConnectError/RemoteProtocolError), not ReadTimeout, and cap attempts at 3. A read timeout means the long-horizon run is in progress, so it is never re-fired.
- Edge case (_extract_text assumed messages[-1] is the assistant reply): 100f6a2a — now scans in reverse for the last assistant/ai message; the final channel-values message can be a tool/human turn. Falls back to the last message's text.
Tests added for the trailing-tool-message and no-role-fallback cases. Live Pi deploy-test of the full install + chat is in progress.

PR #1309, opened 2026-06-22
litellm: default to the in-house per-agent key store

jaylfc:
Per Jay: make the in-house LiteLLM key store the default for all installs (was opt-in behind a marker).

The controller now mints per-agent virtual keys in the in-house SQLite store and authorizes them via the custom_auth hook by default, so every install gets uniform per-agent isolation that works with no DATABASE_URL and no prisma (the ARM fix proven on the Pi). Opt out with a `.litellm_disable_inhouse_keys` marker to keep LiteLLM's Postgres virtual-key table on a Postgres-backed install. The deployer master-key fallback stays as a last-resort safety net.

One-line gate flip in app.py. 100 keystore/proxy/deployer tests green. Already validated end-to-end on the Pi (aarch64) earlier this session.

## Summary by CodeRabbit

* **Chores**
  * Modified default configuration for per-agent virtual key authorization. Users requiring the previous behavior can disable this feature via a configuration marker.

### DIFF
--- a/tinyagentos/app.py
+++ b/tinyagentos/app.py
@@ -311,11 +311,20 @@ def create_app(data_dir: Path | None = None, catalog_dir: Path | None = None) ->
     # Authorize per-agent virtual keys against taOS's own SQLite key store via
     # LiteLLM's custom_auth hook, instead of its Postgres/prisma virtual-key
     # table. Lets per-agent keys work with NO DATABASE_URL and no prisma (the
-    # ARM / no-Postgres fix) and gives every install real per-agent isolation
-    # uniformly. ON by default now; create a ``.litellm_disable_inhouse_keys``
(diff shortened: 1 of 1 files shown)


jaylfc:
Folded the gitar edge case (silent switch on Postgres installs) in 3b87ae27: in-house is now default ON only when there is NO DATABASE_URL configured. A Postgres-backed install keeps LiteLLM's native virtual-key table (it already has per-agent isolation), so an upgrade no longer silently orphans its minted keys / 401s running agents. Force in-house even with Postgres via .litellm_force_inhouse_keys; .litellm_disable_inhouse_keys still disables entirely. This keeps the intent (in-house default for every ARM/no-Postgres install) while being safe on upgrade.

PR #1384, opened 2026-06-23
add a vitest test for desktop/src/components/AppShell

jaylfc:
Autonomous build of board card tsk-czv5js.

Files:
 desktop/src/components/AppShell.test.tsx | 77 ++++++++++++++++++++++++++++++++
 1 file changed, 77 insertions(+)


jaylfc:
Closing: the added AppShell test asserts the build version as rendered text content, but AppShell passes BUILD_VERSION as a prop to UpdateAvailableToast (a build-time constant), so the assertion is wrong and build-env-fragile. Re-queue a fresh card if AppShell coverage is still wanted; a correct test should mock the child and assert the prop, not text.

Issue #106, opened 2026-04-11
tinyagentos.com: community gallery at community.tinyagentos.com (phase 7)


jaylfc:
Public static site at `community.tinyagentos.com` listing user-submitted customisation bundles. Browsing the gallery requires no account — the full catalogue is publicly visible. Bundles are displayed with a thumbnail (generated server-side from the bundle's theme colours and wallpaper at upload time), a download count, a star rating, author badges, and a report button for moderation.

Uploading a bundle requires a valid TAOS account session. The upload endpoint accepts a `.taosbundle` file, verifies the embedded signature against the author's published public key (server-side re-verification — client claims are not trusted), extracts the thumbnail, and stores the bundle in the gallery database. The gallery database lives on the platform LXC in Postgres, separate from the identity database. The Caddy `community.tinyagentos.com` vhost serves a statically-generated or SSR'd HTML front end.

A moderation queue exists for reported bundles. Moderators (initially jaylfc) can approve, reject, or feature bundles. Featured bundles appear in a curated section on the gallery homepage. Report button is available to all visitors; submitting a report requires a brief text reason but no account.

**Acceptance criteria:**
- [ ] Gallery homepage loads without login; bundles browsable by newest / top-rated / featured
- [ ] Bundle upload requires a valid TAOS session; signature re-verified server-side before storage
- [ ] Invalid or missing signature → rejected at upload; no unsigned bundles in the gallery
- [ ] Thumbnail generated at upload and served as a static asset (not re-generated on each page load)
- [ ] Report button submits a report to the moderation queue; confirmation shown to reporter
- [ ] Moderation queue UI accessible to admin accounts; approve/reject/feature actions available
- [ ] Download action serves the original `.taosbundle` file; download count incremented atomically
- [ ] Gallery database schema migrated cleanly alongside the identity database in CI

**Depends on:** #105

PR #1437, opened 2026-06-25
feat(logging): front-end crash capture + in-OS Logs viewer (#106)

jaylfc:
## What
The front-end half of #106: the desktop now ships browser-side errors to the controller so a PWA crash (where no one can open a console) is captured server-side.

- `lib/client-log.ts`: `reportClientLog(level, message, extra)` -- best-effort (never throws/blocks), CSRF-aware via withCsrf, length-capped, with a short client-side dedupe so a render loop does not hammer the endpoint (the server also rate-limits per user). `installGlobalErrorReporting()` hooks `window.onerror` + `unhandledrejection`.
- `main.tsx`: installs the global handlers at startup (next to installAuthGuard).
- `AppErrorBoundary`: the generic-crash branch now also reports the error + componentStack as level `fatal` (in addition to console.error).

## Why
Jay: a PWA has no console the user or we can read. This is the capture loop on top of the backend store/endpoint (#1436) -- once both land and the Pi updates, a crash like the Messages app failure records itself and is readable via GET /api/client-logs.

## Tests
client-log: posts to /api/client-logs, never throws on fetch reject, drops empty message, dedupes within window. AppErrorBoundary: a generic crash posts a level-fatal report. 11/11 green.

## Summary by CodeRabbit

* **New Features**
  * Added a new Logs section in Settings to view captured client logs.
  * Added log filtering by severity, refresh controls, and expanded details for messages, timestamps, sources, and stack traces.

* **Bug Fixes**
  * Improved handling of app crashes and unhandled errors by sending fatal logs for better troubleshooting.
  * Added safeguards so logging is best-effort and won’t interrupt normal app use.

### DIFF
--- a/desktop/src/apps/SettingsApp.tsx
+++ b/desktop/src/apps/SettingsApp.tsx
@@ -15,6 +15,7 @@ import {
   Users,
   Palette,
   UserCircle,
+  ScrollText,
 } from "lucide-react";
 import {
   Button,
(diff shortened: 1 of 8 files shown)


jaylfc:
Added in a follow-up commit on this branch: an in-OS **Settings -> Logs** viewer (LogsPanel) that reads GET /api/client-logs (admin) and lists captured errors/crashes newest-first with a level filter + refresh. This closes the #106 loop end to end: capture (error-boundary + global handlers) -> store (#1436) -> view in the UI, so a crash is chased in Settings rather than by curling the controller. 3 LogsPanel tests added; AccountPanel + Settings suites still green.

Issue #176, opened 2026-04-12
Agent Browsers — container manager + cookie export (Knowledge Pipeline Step 7)


jaylfc:
## Summary
Build the Agent Browsers app — create/destroy persistent Chromium browser containers, assign to agents, VNC preview, cookie export for platform auth.

## Status
Implementation plan written at `docs/superpowers/plans/2026-04-12-agent-browsers.md`.

## Dependencies
- Container runtime (LXC/Docker)

## Unblocks
- X Monitor (cookie auth)
- YouTube Library (cookie fallback)

## Spec
`docs/superpowers/specs/2026-04-12-knowledge-capture-pipeline-design.md` — Agent Browsers section

jaylfc:
Implemented — AgentBrowsersManager + routes + AgentBrowsersApp

PR #1467, opened 2026-06-28
feat: deploy agents onto cluster workers (remote deploy, #176/#150)

jaylfc:
Completes the worker-deploy path so an agent container is created ON a cluster worker's nested incus, not just on the controller. Builds on the worker-LXC foundation (#1466).

## Slices (each tested)
1. `create_container(remote=)` creates on an enrolled remote's nested incus (qualified image + instance, host mounts skipped).
2. `ensure_image_present(remote=)` prefetches the base onto a worker (incus image import/list with a `<remote>:` positional; local-only bake skipped).
3. `deploy_agent` threads `remote`: creates on the worker, qualifies every incus op with `<remote>:`, skips the localhost proxy devices, and points the agent's callback at the controller (`taos_host`). Record name stays unqualified.
4. Routing: an explicit `target_worker` pin deploys the container on that worker (was a 202 stub). `configure_remote_deploy` resolves the pin (409 if unknown/offline), sets the controller Tailscale IP as the callback host, and best-effort prefetches the correct-arch base.

## Networking (per Jay: manual Tailscale, not headscale)
The agent reaches the controller over the tailnet (`controller_callback_host` -> `tailscale ip -4`). The live on-worker test is held pending a decision on nested-container -> tailnet routing (see pending-decisions); same-LAN workers can use the LAN IP in the meantime.

## Tests
configure_remote_deploy unit tests + routing fall-through + create_container/ensure_image_present/deploy_agent remote paths. Full deployer/containers/agent_image/agent_deploy suites green (126).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for deploying agents to a specified remote worker, including remote base-image prefetch and correct worker callback wiring.
  * Remote deployments now use networked LiteLLM endpoints and honor an optional controller callback host override.
* **Bug Fixes**
  * Remote image import/presence checks now target enrolled remotes; remote-imported images are not baked.
  * Remote container creation skips local-only mounts and related proxy/device setup, while still applying resource limits and environment on the remote target.
  * Container-related responses now consistently return the bare container name.
* **Tests**
  * Expanded coverage for remote image importing, remote container creation, remote deploy execution, and remote routing edge cases.
* **Chores**
  * Improved nested incus initialization logic during worker setup.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -699,11 +699,15 @@ phase2_inside_lxc() {
         warn "  to enable later: apt install bees && systemctl enable --now bees.service"
     fi
 
-    # Gate init on a storage pool existing, NOT on `incus list` succeeding:
-    # `incus list` returns 0 even on a fresh daemon with no storage pool and an
-    # empty default profile, which would skip init and leave the nested incus
-    # unable to create agent containers ("No root device could be found").
-    if incus storage list --format csv 2>/dev/null | grep -q .; then
+    # Gate init on the default storage pool existing, NOT on `incus list`
+    # succeeding: `incus list` returns 0 even on a fresh daemon with no storage
+    # pool and an empty default profile, which would skip init and leave the
+    # nested incus unable to create agent containers ("No root device could be
+    # found"). Match the pool name exactly, as create_btrfs_loopback does at the
+    # host level: `--format=csv` still emits a header row on some incus builds,
+    # so a bare `grep -q .` matches the header and wrongly skips init on a fresh
+    # daemon. `incus admin init --minimal` creates a pool named `default`.
+    if incus storage list --format=csv 2>/dev/null | awk -F',' '{print $1}' | grep -q '^default$'; then
         log "nested incus already initialised"
     else
         incus admin init --minimal < /dev/null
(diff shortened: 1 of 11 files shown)


jaylfc:
Reviewed the three Major findings against the current head (e3ac9f60 + tailscale-log commit):

1. **deployer.py LiteLLM/embedding URLs** — already rewritten for remote. The proxy block sets `_llm_base = http://{taos_host}:{port}` when `req.remote`, and both `OPENAI_BASE_URL` and `TAOS_EMBEDDING_URL` derive from it (deployer.py 283-291); the fallback default also uses `taos_host` (341-345). No localhost leaks to a remote agent.

2. **containers launch syntax for public images** — the `":" not in image` guard means a remote-image-server ref like `images:debian/bookworm` is left unqualified, producing the valid `incus launch images:debian/bookworm <remote>:<name>`; only bare prefetched aliases get `<remote>:` prefixed. Verified working on the live wkrlan cold-path deploy.

3. **prefetch out of the request path** — already moved. `configure_remote_deploy` only validates the worker; the 300-500MB import runs in the background `prefetch_base_onto_worker` called from the deploy task, not inline.

Folded the one remaining quick win (log the tailscale lookup fallback instead of swallowing it).

PR #1474, opened 2026-06-28
feat(notes): agent write tool for shared docs (option C, slice 2)

jaylfc:
Option C, slice 2: agents can write to shared notes/lists they belong to, completing the member-writes feature.

## What
- Store: `docs_for_agent(agent_name)` returns the non-archived docs where the agent is an agent-type member.
- New agent tools (`tinyagentos/tools/notes_tools.py`), dispatched in-process with the calling agent's identity injected as `agent_name` (the standard skill_exec convention):
  - `notes_list_shared_docs` -- the docs the agent is shared on, so it can discover what it may write to.
  - `notes_add_entry(doc_id, text)` -- appends an entry. Authorization is membership: the call is rejected unless the agent is an agent-member of the doc. The write is attributed to the agent, and the existing `_trigger_agent_notifications(..., skip_agent=agent_name)` notifies the OTHER agent members but never the writer (the self-notify loop guard from slice 1).
- Registered both in `SKILL_IMPLEMENTATIONS` mirroring `notify_user` / `list_projects`, with `agent_name` injected by dispatch (not model-facing).
- Agent manual: documented both tools in `docs/agent-manual/09-os-control.md`.

## Why
This is the agent half of option C (Jay: user and agent members can write). With slice 1 (human member writes) it means a user can share an "app ideas" note with an agent, the agent gets pinged on new entries, and now the agent can append its research or critique straight back into the doc without looping on its own write.

## Tests
41 notes tests pass (8 new): agent member can add an entry, non-member agent rejected, the write notifies other agents but not the writer, plus list/archived/validation cases. compileall clean.

## Follow-up
The two app UIs (Notes + Todo) to the impeccable/taste bar, live screenshot-verified.

## Summary by CodeRabbit

* **New Features**
  * Added support for shared notes/lists so agents can view shared documents they belong to and append new entries to them.
  * Shared documents now return a simple list with recent updates first.

* **Documentation**
  * Updated the agent manual with guidance for using the new shared-notes capabilities.

* **Tests**
  * Added coverage for shared-doc access, entry creation, membership checks, and required-field validation.

### DIFF
--- a/docs/agent-manual/09-os-control.md
+++ b/docs/agent-manual/09-os-control.md
@@ -30,19 +30,15 @@ update the open Projects app in real time):
   GPU model for a quality cover. The system loads/unloads and queues for you — you
   just choose the model.
 
-A typical flow: open the Projects app, create_project, add a few tasks, call
-generate_image and keep its `image_ref`, then canvas_add_image(project_id, image_ref)
-to drop it on the board. To finish a storybook, call export_storybook(project_id,
-title, pages) to produce the illustrated PDF in the project's Files.
+A typical flow: open Projects, create_project, add tasks, generate_image then
+canvas_add_image(project_id, image_ref) to place it; export_storybook(project_id,
+title, pages) writes the illustrated PDF to the project's Files.
 
-These drive the user's own desktop in their session. Use them to make your work
-visible: open the relevant app so the user can watch, then carry out the task with
-that app's own tools and your other skills.
+These drive the user's own desktop. Make your work visible: open the relevant app
+so the user can watch, then carry out the task with that app's tools. Open only
+what you need, leave their windows alone, say what you do.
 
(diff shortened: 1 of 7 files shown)


jaylfc:
Folded the three CodeRabbit Major findings:
- **docs_for_agent payload**: now selects only `id, kind, title, updated_at`, so internal columns (e.g. `owner_user_id`) are not exposed to the agent.
- **archived docs**: `notes_add_entry` now fetches the doc and rejects it when `archived_at` is set, before appending.
- **writer attribution**: `add_entry` gained an `editor_type` param and the tool passes `'agent'`, so the create revision is attributed to the agent in history.

Also logged the best-effort trigger failure instead of swallowing it. Added tests for all three; 44 notes tests pass.

PR #1517, opened 2026-07-01
fix(updater): install the proxy extra so updates never strip litellm

jaylfc:
## Problem

The Settings-update dependency install (`_install_dependencies`) ran a bare `uv sync --frozen`. `uv sync` prunes the venv to the locked **default** dependency set, so it silently **uninstalls `litellm[proxy]` + `prisma`** — which live in the optional `proxy` extra — on every update.

`install-server.sh` installs the extra (`pip install -e ".[proxy]"`), so a **fresh** box has the LLM proxy, but **every subsequent Settings update disabled it**. The controller then logs `LiteLLM not installed — proxy disabled` and the proxy never binds. The LLM proxy is a **core** dependency — basic agent functionality fails without it.

Discovered on the live Pi: after a clean reconcile + Settings update, `:4000` wasn't listening and litellm wasn't importable in the venv, despite being present in `uv.lock`. Restored on the box with `uv sync --frozen --extra proxy`.

## Fix

Carry the `proxy` extra on both updater paths, matching install-server.sh:
- uv path: `uv sync --frozen --extra proxy`
- pip fallback: `pip install -e ".[proxy]"`

## Tests

`tests/test_updater_dep_install.py` updated to pin the new commands; all 8 pass locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the in-app “Install Update” flow to install the complete dependency set needed for proxy functionality.
  * Made the dependency set consistent across both update install methods to prevent missing capabilities after an update.
  * Preserved existing handling for failed installs and pending-restart behavior.
* **Tests**
  * Strengthened update-install coverage to verify the installed extras match the official install script, including the proxy extra.

### DIFF
--- a/tests/test_updater_dep_install.py
+++ b/tests/test_updater_dep_install.py
@@ -5,8 +5,10 @@ Settings "Install Update" flow:
 
   * `_find_uv` resolves uv robustly (uv is not always on PATH; on the Pi it
     lives at <install_dir>/.local/bin/uv).
-  * `_install_dependencies` prefers a lockfile-pinned `uv sync --frozen` and
-    falls back to `pip install -e .` only when uv is absent.
+  * `_install_dependencies` prefers a lockfile-pinned `uv sync --frozen
+    --extra proxy` and falls back to `pip install -e .[proxy]` only when uv is
+    absent. The proxy extra (litellm + prisma) is a core dep, so both paths
+    carry it to match install-server.sh.
   * a non-zero install return code aborts the update WITHOUT writing the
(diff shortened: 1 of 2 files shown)


jaylfc:
Folded the review: the `proxy` extra was hardcoded in two places. Now single-sourced as `settings.UPDATE_EXTRAS` (both updater commands build from it), plus `test_updater_extras_match_install_server` parses `install-server.sh`'s pip extras and asserts parity, so any future drift (e.g. adding `.[proxy,gpu]` to one side only) fails CI. 9/9 updater tests pass.

PR #1519, opened 2026-07-01
fix(llm-proxy): self-heal the litellm proxy extra on the bootstrap update

jaylfc:
## Why

#1517 stops *future* updates from stripping litellm, but the transitional update that **delivers** that fix is executed by the **old** updater code (bare `uv sync --frozen`), which prunes the `proxy` extra before the new code takes over. Verified live on the Pi: updating to the fix commit left `:4000` down and litellm not importable. Without this, every user updating to the fix-release loses their LLM proxy once.

## Fix

When `LLMProxy.start()` finds litellm missing, self-install the proxy extra once:
- `uv sync --frozen --extra proxy` (pip fallback `-e .[proxy]`), extras single-sourced from `settings.UPDATE_EXTRAS`.
- Guarded to run **once per process** (`_selfheal_attempted`), **bounded** (900s timeout), **non-fatal** (any failure just leaves the proxy disabled and the box running).
- Runs inside the existing **background** litellm-bringup task, so a slow install never blocks controller startup - the proxy simply arrives a minute or two after boot.

## Tests

3 new tests in `tests/test_llm_proxy.py` (command construction, non-zero handling, missing-root guard); full file 41/41 pass.

## Verification plan

After merge + Settings-update to the Pi: deliberately strip litellm, restart, and confirm the self-heal reinstalls it and the proxy binds - the definitive proof before the release.

## Summary by CodeRabbit

* **New Features**
  * Added automatic recovery during startup when a required proxy component is missing, helping the app continue without manual repair.

* **Bug Fixes**
  * Improved startup behavior so failed recovery no longer blocks the app; it now safely stays disabled if recovery cannot complete.

* **Tests**
  * Added coverage for successful recovery, failed recovery, and cases where recovery is skipped.

### DIFF
--- a/tests/test_llm_proxy.py
+++ b/tests/test_llm_proxy.py
@@ -537,3 +537,71 @@ class TestInhouseKeys:
         assert proxy._keystore().lookup(key)["allowed_models"] == ["b", "c"]
         assert await proxy.delete_agent_key(key) is True
         assert proxy._keystore().lookup(key) is None
+
+
+class TestProxySelfHeal:
+    """The proxy self-installs the litellm extra once if a pre-fix update
+    stripped it (bounded + non-fatal), so agents are not left without a route."""
+
+    @pytest.mark.asyncio
+    async def test_selfheal_uses_uv_sync_extra_proxy(self, tmp_path, monkeypatch):
+        import sys
+        import tinyagentos.llm_proxy as mod
+
+        (tmp_path / "pyproject.toml").write_text("[project]\n")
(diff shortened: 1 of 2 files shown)


jaylfc:
Folded the CRITICAL: the self-heal install subprocess is now killed + bounded-reaped on `wait_for` timeout (mirrors the updater's `_run_capture` pattern), so a stuck `uv`/`pip` cannot leak past the 900s bound. Added `test_selfheal_kills_subprocess_on_timeout`; 42/42 proxy tests pass.

PR #1520, opened 2026-07-01
fix(doc-gate): fold Qodo findings (glob precision, CONTRIBUTING drift, single-source trailer)

jaylfc:
Folds all three Qodo findings on the merged doc-gate (#1518):

1. **Glob crossed directory separators (correctness).** `_match_any` used `fnmatch`, where `*` matches `/`, so `tinyagentos/routes/*.py` wrongly matched nested files like `tinyagentos/routes/desktop_browser/__init__.py`, over-triggering the routes rule. Added `_glob_match` (path-segment-aware: `**`→across `/`, `*`→within a segment, `?`→one non-sep char); `_match_any` uses it.
2. **CONTRIBUTING drift (correctness).** The routes-rule row listed `docs/AGENT_HANDOFF.md` as a satisfying doc, but it is gitignored/untracked and not in `doc-gate.toml`, so following CONTRIBUTING would still fail CI. Now lists only `docs/agent-coordination.md`.
3. **Trailer hardcoded in hooks (maintainability).** `.githooks/commit-msg` now sources the trailer from config via a new `print-trailer` subcommand (fallback preserved), so it cannot desync from CI.

Tests: `tests/test_doc_gate.py` gains glob-precision + evaluate_rules + trailer tests; 20 pass. `invariants` + `diff-gate --base origin/dev` clean; hooks `bash -n` + shellcheck clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved documentation gate checks so path matching is more precise and nested files are handled correctly.
  * Commit messages now pick up the required documentation trailer automatically when available, with a safe fallback if not.

* **Documentation**
  * Updated contribution guidance for route changes: only the main agent coordination docs are now required.

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -8,10 +8,11 @@
 # configures for CI (default: "Docs-Reviewed:").
 set -euo pipefail
 
-trailer="Docs-Reviewed:"
 msg_file="$1"
 repo_root="$(git rev-parse --show-toplevel)"
 cd "$repo_root"
+
+trailer="$(python3 "$repo_root/scripts/check_doc_gate.py" print-trailer 2>/dev/null || echo 'Docs-Reviewed:')"
 
 if python3 scripts/check_doc_gate.py diff-gate --staged >/dev/null 2>&1; then
     exit 0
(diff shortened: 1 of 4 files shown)


jaylfc:
Folded both Kilo suggestions: evaluate_rules now uses get_trailer(config) (true single source), and _glob_match maps a trailing /** to zero-or-more segments so app-catalog/** matches the bare parent too. Added test_glob_double_star_matches_bare_parent; 21/21 pass.

PR #1522, opened 2026-07-01
fix(sw): network-first for the SPA shell (stops redeploys breaking open PWA sessions)

jaylfc:
## Problem

The service worker served the SPA shell HTML **stale-while-revalidate** (cached index first). After a redeploy, the cached index references old hashed chunk URLs that 404, so lazy imports throw `ChunkLoadError`, the error boundary auto-reloads, and it lands on the default pane. Repro (Jay): Settings > Account after several rebuilds -> `fatal: Importing a module script failed` in client logs -> reload loop back to System Info.

## Fix

Serve the shell **network-first**: fetch the current index when the backend is healthy so its chunk refs match the deployed assets. Fall back to the cached shell only on a **non-OK response** (503 mid-restart after Install Update) or a **network failure** (offline) - preserving the reconnect UX the SW cache exists for. Immutable hashed assets stay cache-first; icons/manifests stay stale-while-revalidate.

## Notes
- HTTP-layer was already correct (`index.html` is `no-cache`); the SW cache was the remaining stale layer.
- Existing open sessions need one refresh to pick up the new SW (skipWaiting + clients.claim then take over); after that, redeploys no longer strip sessions.
- `desktop/tests/e2e/sw-and-reconnect.spec.ts` covers precache + offline-shell; the non-OK/offline fallback preserves both.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app loading reliability by reducing stale cached shell pages after updates.
  * The app now prefers the latest online shell for smoother navigation and fewer reload loops caused by outdated cached files.
  * When offline or during restarts, it falls back to the last known cached shell to keep the UI accessible.

### DIFF
--- a/desktop/src/sw.ts
+++ b/desktop/src/sw.ts
@@ -107,19 +107,26 @@ self.addEventListener("fetch", (event: FetchEvent) => {
     return;
   }
 
+  // isShellHTML is checked BEFORE isPrecachedStatic on purpose: /desktop/,
+  // /desktop/index.html and /chat-pwa are in PRECACHE_URLS but must use
+  // network-first, not the SWR branch below -- a stale cached index
+  // reintroduces the ChunkLoadError loop. Keep this branch first.
   if (isShellHTML(url)) {
-    // Network-first for the SPA shell: a stale cached index references old
-    // hashed chunk URLs that 404 after a redeploy -> ChunkLoadError -> reload
-    // loop. Always try the network so the served index matches the deployed
-    // assets; fall back to cache only when the network fails (offline /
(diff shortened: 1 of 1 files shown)


jaylfc:
Folded all findings: shell fetch now uses `cache: 'no-store'` (no 304-empty-body slip-through), a 4s `AbortController` timeout so a backend stall falls back to the cached shell instead of hanging, and a comment documenting that `isShellHTML` must be checked before `isPrecachedStatic` (overlapping precache list). The no-hit 503 path already returns `hit || fresh` as suggested.

PR #1523, opened 2026-07-01
fix(sw-register): reload on new SW control so redeploys swap open sessions

jaylfc:
Completes the SW cache fix. #1522 made the SW network-first (so a reload gets the fresh index), but nothing reloaded an already-open session, so the page kept running old code with now-404 chunk refs -> ChunkLoadError loop on opening an app (repro: Settings > Account after a deploy; client log: `Importing a module script failed`, stack shows a lazy app chunk failing under Suspense).

`registerServiceWorker` now, on a returning session (an existing controller), listens for `controllerchange` and reloads once when the new SW takes control (sw.ts already does skipWaiting + clients.claim), and calls `registration.update()` on boot to pick up a fresh deploy promptly. First-ever visits skip the reload. 6/6 sw-register tests pass; tsc clean.

Note: an already-open session still needs one manual reset to get ONTO this new code; after that, future deploys swap seamlessly.

## Summary by CodeRabbit

* **New Features**
  * Returning sessions now auto-reload when service worker control changes, helping users access the latest version sooner.
  * Service worker registration now checks for updates immediately after setup.
* **Bug Fixes**
  * Prevents repeated reloads from happening multiple times for the same control change.
  * Avoids enabling reload behavior on first-time visits when there is no existing service worker controller.
* **Tests**
  * Expanded service worker registration tests to cover update calls, reload wiring, and “reload only once” behavior.

### DIFF
--- a/desktop/src/lib/__tests__/sw-register.test.ts
+++ b/desktop/src/lib/__tests__/sw-register.test.ts
@@ -86,4 +86,20 @@ describe("registerServiceWorker", () => {
     await registerServiceWorker();
     expect(addEventListener).not.toHaveBeenCalled();
   });
+
+  it("wires the controllerchange listener at most once across calls", async () => {
+    const addEventListener = vi.fn();
+    const register = vi.fn().mockResolvedValue({ update: vi.fn() });
+    // Same container instance across both calls (StrictMode / remount).
+    const swMock = { register, addEventListener, controller: {} };
+    Object.defineProperty(navigator, "serviceWorker", {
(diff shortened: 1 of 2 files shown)


jaylfc:
Folded all three: controllerchange listener now wired at most once per SW container (instance-marker guard, idempotent across StrictMode/HMR re-invokes, with a test); update() errors logged via console.debug instead of swallowed; and a comment noting controllerchange's broader scope (we never unregister, so in practice it means a new version took control). 7/7 sw-register tests pass.

PR #1524, opened 2026-07-01
chore(release): 1.0.0-beta.14

jaylfc:
Version bump + CHANGELOG for the beta.14 release (dev -> master promotion).

**Why now:** master is still v1.0.0-beta.13, which ships the broken updater that silently uninstalls litellm on every Settings update. This release replaces it.

Bumps the 4 version files (pyproject.toml, desktop/package.json, tinyagentos/__init__.py, uv.lock) and dates the CHANGELOG section. `test_version_lock_sync` passes.

**Headline fixes in beta.14:** updater no longer strips the litellm proxy extra (#1517); proxy self-heals if missing (#1519/#1521); PWA reliably swaps to the new build after a deploy instead of crashing/looping (#1522/#1523). Full set in the CHANGELOG.

After this merges to dev, follow-up PR dev -> master, then tag v1.0.0-beta.14 + GitHub Release.

## Summary by CodeRabbit

* **New Features**
  * Added live desktop notification streaming, live Agents panel updates, and per-session approval-mode (“Observatory”).
  * Introduced append-only, content-hashed action receipts/audit trail and improved A2A bus agent authentication.
  * Added a documentation-drift gate to keep changes aligned.

* **Changed**
  * Updated agent consent to be non-blocking (bell/toast) with archived history.
  * In-app updates now install exactly lockfile-pinned dependency versions.
  * Moved the install directory to a new path while preserving existing installs.

* **Bug Fixes**
  * Improved reliability/security across update/startup, including LLM proxy preservation/self-healing, fresh desktop reloads, and boot crash-loop recovery.
  * Fixed Settings > Account login behavior to prevent session-expired bounces.

* **Tests**
  * Added coverage to ensure session-expired is not dispatched for `401` responses from account endpoints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
 - The LLM proxy now self-heals: if litellm is missing at startup (e.g. left over from an update before the fix above), it reinstalls itself once automatically and comes back online.
 - The desktop no longer breaks after an update: it reliably loads the new version after a redeploy instead of getting stuck on stale cached code or crashing when opening an app, on Chrome, Safari, and Firefox alike.
 - A failed background rebuild during boot no longer crash-loops the controller; it now falls back to the last working build and logs a warning instead.
+- Opening Settings > Account no longer flashes the login screen and bounces you back to System Info when you are not signed into a taOS cloud account; the account service's expected "not signed in" response is no longer mistaken for your device session expiring.
 
 ## [1.0.0-beta.13] - 2026-06-28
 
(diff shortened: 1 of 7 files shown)


jaylfc:
Bundled an additional user-facing fix into this release (63301813):

**fix(auth-guard): Settings > Account no longer flashes the login screen and bounces to System Info**

Root cause (reproduced live as the signed-in user via the control API): the global auth guard wraps `window.fetch` and fires `taos-session-expired` on any 401 from an `/api/*` path, which LoginGate turns into a re-auth flash + desktop remount (Settings resets to its default section, System Info). The account proxy `/api/account/me` returns 401 whenever the user is **not signed into the taos.my cloud account** — a normal signed-out state that account-client already maps correctly — so opening the Account pane misfired the guard. Fix excludes `/api/account/*` from the guard (same rationale as the existing `/auth/*` carve-out); genuine controller-session expiry is still caught by every other `/api/*` call. Added a regression test.

PR #1526, opened 2026-07-01
fix: fold the four Kilo findings from the beta.14 promotion review

jaylfc:
Immediate fix-forward for the four inline findings Kilo raised on the #1525 promotion diff (all pre-existing on dev, none introduced by the release):

1. **llm_proxy self-heal, pip fallback** (`llm_proxy.py`): no longer runs `pip install -e .[proxy]` (which re-resolves every project dependency, the exact churn the self-heal exists to undo). It now installs only the extras' pinned requirement strings read from `pyproject.toml`, keeping the pins single-sourced. Deliberately NOT the reviewer's `--no-deps` variant: that would skip the extra's dependencies entirely and install nothing useful.
2. **llm_proxy self-heal, root discovery**: walks up from `sys.executable` to the first ancestor holding `pyproject.toml` instead of hard-coding `parents[2]`, so `python3.x`-named binaries and nested venv layouts still self-heal instead of silently no-opping.
3. **agent registry, seed-internal adoption**: `adopt` is now an explicit comma-separated handle list (`?adopt=@taOS-dev`); a bare `adopt=true` returns 400. A driver handle claimed by someone else via the consent flow can no longer be vouched for as a side effect of seeding the others. New tests cover the 400 and that unlisted pre-existing handles still 409.
4. **channel hub, reconnect leak**: `connect_bot` stops a prior connector under the same `platform:agent` key before replacing it (single fix above the platform dispatch, covering every token-based branch), so reconnecting after a token rotation no longer orphans the old background task/client.

163 tests pass across the affected suites; the 4 new tests pin the changed behaviours. Ruff findings in these files are pre-existing on dev (identical before/after).

## Summary by CodeRabbit

* **Bug Fixes**
  * Internal agent seeding now supports handle-specific adoption via a comma-separated `adopt` list, rejecting blanket/invalid values with HTTP 400 and ensuring conflicts occur for unlisted handles.
  * Channel hub reconnects/disconnects are now synchronized per connector key, ensuring the previous connector is cleanly stopped/replaced and preventing race conditions.
  * LLM proxy self-healing more reliably finds the correct project root and improves the non-`uv` fallback by installing only the needed extras.
* **Tests**
  * Expanded self-heal and internal seeding scenarios, and updated fixture metadata to match expected project naming.

### DIFF
--- a/tests/test_agent_internal_mint.py
+++ b/tests/test_agent_internal_mint.py
@@ -245,14 +245,16 @@ class TestMintInternalRoute:
         assert kwargs["actor_user_id"]
 
     async def test_seed_internal_adopt_handles_preexisting(self, mint_client):
-        """seed-internal?adopt=true vouches for a pre-existing driver handle and
-        still seeds the rest."""
+        """seed-internal?adopt=@handle vouches for THAT pre-existing driver
+        handle explicitly and still seeds the rest."""
         rec = await mint_client._app.state.agent_registry.register(
             framework="claude-code", display_name="taos-dev",
             origin="external-selfjoin", handle="@taOS-dev",
         )
         await mint_client._app.state.agent_registry.set_status(rec["canonical_id"], "active")
-        resp = await mint_client.post("/api/agents/registry/seed-internal?adopt=true")
+        resp = await mint_client.post(
+            "/api/agents/registry/seed-internal?adopt=@taOS-dev"
+        )
         assert resp.status_code == 200, resp.text
         seeded = {s["handle"]: s for s in resp.json()["seeded"]}
         assert seeded["@taOS-dev"]["adopted"] is True and seeded["@taOS-dev"]["created"] is False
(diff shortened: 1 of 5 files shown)


jaylfc:
Folded Kilo's second-round findings in 2b75b134:

- Root walk now requires `project.name == "tinyagentos"` before trusting an ancestor's pyproject, so a foreign pyproject higher up can never have its pins installed into our venv (new test).
- Requirement strings are stripped, and the pip fallback runs with `--no-input --disable-pip-version-check` (kept stderr un-quieted so failures stay diagnosable in the logs).
- The seed-internal 400 no longer enumerates the internal handle set; it echoes only the caller's bad input.
- Stop-prior now covers the webchat branch too, and every connect is serialized through a per-key `asyncio.Lock` held across stop + start + map-assign, closing the concurrent-reconnect orphan window.
- Added the suggested multi-handle adopt test (spacing + duplicate). One correction: `str.strip()` already removes tabs and newlines, so the handle-parsing concern needed only the test, not a code change.

170 tests pass across the affected suites.

jaylfc:
Folded the round-3 findings in a59cf66e:

- Matrix homeserver is validated before the destructive stop, so a malformed reconnect can no longer tear down the live connector and then 400 (this one was a real regression introduced by the stop-prior change on this branch, good catch).
- disconnect_bot runs under the same per-key lock as connect_bot, closing the disconnect-vs-reconnect race.
- The lock map is refcounted and evicts entries once no task holds or awaits them, so it no longer grows for every key ever connected; the bookkeeping has no awaits, making it atomic under asyncio's single-threaded scheduling.

165 tests pass across the affected suites.

PR #1528, opened 2026-07-02
fix(rknpu): re-pin rkllama to a reachable commit + verify the pin before checkout

### LINKED ISSUE #1527: install-rknpu.sh fails when the pinned rkllama commit no longer exists
## Bug report

The `install-rknpu.sh` installer currently pins `rkllama` to the following commit:

```bash
RKLLAMA_REF="${TAOS_RKLLAMA_REF:-06cf874d8b29767729ec06547cf02fc92acd875c}"
```

During installation it executes:

```bash
git -C "$RKLLAMA_DIR" checkout "$RKLLAMA_REF"
```

However, this commit no longer exists in the `jaylfc/rkllama` repository, causing the installer to fail on a fresh installation with:

```text
fatal: reference is not a tree: 06cf874d8b29767729ec06547cf02fc92acd875c
```

As a result, the installation cannot complete successfully.

---

## Expected behavior

The installer should either:

- use a valid pinned reference;
- detect that the pinned commit no longer exists and emit a descriptive error message; or
- gracefully fall back to a stable branch/tag (if that behavior is acceptable).

---

## Suggested improvement

Before attempting the checkout, verify that the commit exists after fetching:

```bash
git -C "$RKLLAMA_DIR" fetch origin --tags --prune

if git -C "$RKLLAMA_DIR" cat-file -e "${RKLLAMA_REF}^{commit}" 2>/dev/null; then
    git -C "$RKLLAMA_DIR" checkout "$RKLLAMA_REF"
else
    echo "Pinned rkllama commit no longer exists."

    DEFAULT_BRANCH="$(git -C "$RKLLAMA_DIR" remote show origin \
        | awk '/HEAD branch/ {print $NF}')"

    echo "Falling back to $DEFAULT_BRANCH"

    git -C "$RKLLAMA_DIR" checkout "$DEFAULT_BRANCH"
    git -C "$RKLLAMA_DIR" pull origin "$DEFAULT_BRANCH"
fi
```

Alternatively, if reproducibility is the goal, pinning to a Git tag instead of a raw commit hash would make the installer much more resilient to history rewrites.

---

## Environment

- OS: Ubuntu (ARM64)
- Installer: `scripts/install-rknpu.sh`
- Failure occurs on a clean installation.

### Additional information

As a quick test, I changed the default value of `RKLLAMA_REF` from the pinned commit to `main`:

```bash
RKLLAMA_REF="${TAOS_RKLLAMA_REF:-main}"
```

This bypassed the `fatal: reference is not a tree` error, and the installer completed successfully on my system without any further installation errors.

I'm not suggesting that using `main` is the ideal long-term solution, since pinning a specific revision provides better reproducibility. However, this does indicate that the current failure is caused solely by the pinned commit no longer existing in the upstream repository, rather than by another issue in the installer itself.

Thanks for maintaining the project!

jaylfc:
Fixes #1527.

The rkllama pin `06cf874d` still exists on GitHub as an API-visible object but is no longer reachable from any branch of the jaylfc/rkllama fork (history was rebased onto upstream), so a fresh `git clone` + `git checkout` fails with the reported `fatal: reference is not a tree`. Verified via the compare API: main is diverged from the pin and no branch head contains it.

Changes:
- Re-pin `RKLLAMA_REF` to `4f518e97d114` (current main head: the rerank endpoint plus the GET_LOGITS 1-token cap fix), which fresh clones can check out.
- After clone/fetch, verify the pinned ref resolves to a commit (`git cat-file -e <ref>^{commit}`) and fail with a descriptive error naming the orphaned-pin cause and the `TAOS_RKLLAMA_REF` override, instead of git's opaque message.
- Deliberately no auto-fallback to a moving branch: installs stay reproducible; a stale pin should fail loudly, not silently drift.

`bash -n` passes; the one shellcheck warning in the file (SC2034, line 511) is pre-existing and untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Rockchip NPU installer reliability on fresh installs by updating the default pinned `rkllama` ref.
  * Added safer validation of the pinned ref and a clearer, actionable error when the ref can’t be resolved, including override instructions.
* **Changelog**
  * Updated the Unreleased → Fixed entry to reflect the installer behavior change and guidance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
 
 ## [Unreleased]
 
+### Fixed
+- The Rockchip NPU installer no longer fails on fresh installs with "reference is not a tree": the rkllama pin now points at a commit that fresh clones can actually check out, and a stale pin fails with a clear explanation and override instructions instead of a raw git error.
+
 ## [1.0.0-beta.14] - 2026-07-01
 
 ### Added
--- a/scripts/install-rknpu.sh
+++ b/scripts/install-rknpu.sh
@@ -336,6 +336,13 @@ install_rkllama() {
         log "cloning $RKLLAMA_REPO -> $RKLLAMA_DIR"
         run_as_user mkdir -p "$(dirname "$RKLLAMA_DIR")"
         run_as_user git clone --quiet "$RKLLAMA_REPO" "$RKLLAMA_DIR"
+        # A clone brings down all branches and tags, but an overridden
+        # TAOS_RKLLAMA_REF can be a SHA reachable from none of them. Try to
+        # fetch the configured ref explicitly (best-effort: some servers
+        # refuse unadvertised objects) so the verification below sees the
+        # same coverage as the re-install path; on failure fall through to
+        # the curated error instead of dying on git's raw message.
+        run_as_user git -C "$RKLLAMA_DIR" fetch --quiet origin "$RKLLAMA_REF" 2>/dev/null || true
     fi
     # Verify the pinned ref is actually fetchable before checking it out. A
     # fresh clone only has branch/tag refs, so a pin orphaned by a history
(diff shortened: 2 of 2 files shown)


jaylfc:
Folded in 908d7eca: after a fresh clone, the installer now does a best-effort explicit fetch of the configured ref before verification, so an overridden TAOS_RKLLAMA_REF that is a valid-but-unreferenced SHA gets the same coverage as the re-install path; a genuinely bad ref still fails through the curated error rather than a raw git message.

One correction to the premise: a plain git clone does fetch all branches and tags (it is not default-branch-only), so the real gap was only SHAs reachable from no ref; the error message now says "branch or tag".

PR #1532, opened 2026-07-02
feat(install): environment banner at the top of both installers

jaylfc:
First output of the installer hardening audit (vendor-image support push, context in #2/#1527):

Both install-server.sh and install-rknpu.sh now log an environment banner before doing anything: distro (os-release PRETTY_NAME), kernel, board (device-tree model), python3 version, free root disk, and (rknpu) the currently installed librknnrt version.

Why: community testers are starting to run fresh installs on OSes we have not verified (official Orange Pi Debian Bookworm on an Orange Pi 5 Pro is in flight). Their pasted logs should self-describe the machine so failures are diagnosable without a round-trip asking for uname -a and /etc/os-release.

bash -n passes on both scripts; banner logic smoke-tested for the no-os-release and no-device-tree paths.

## Summary by CodeRabbit

* **New Features**
  * Installer output now includes a clearer environment summary before setup begins.
  * Displays detected OS, kernel version, architecture, board model, Python 3 version, disk space, and available runtime details when present.

PR #1533, opened 2026-07-02
fix(rknpu): re-pin rkllama to the proven production lineage (fixes #1529)

### LINKED ISSUE #1529: install-rknpu.sh generates a systemd service with an unsupported --preload argument
## Bug report

After working around the previous issue with the missing pinned commit and completing the installation, the generated `rkllama.service` fails to start.

The service exits immediately with the following error:

```text
Jul 01 21:34:12 taos-orange5pro python[13123]: usage: rkllama_server [-h] [--processor PROCESSOR] [--port PORT] [--debug]
Jul 01 21:34:12 taos-orange5pro python[13123]:                       [--models MODELS] [--llamacpp LLAMACPP]
Jul 01 21:34:12 taos-orange5pro python[13123]: rkllama_server: error: unrecognized arguments: --preload qwen3-embedding-0.6b,qwen3-reranker-0.6b,qmd-query-expansion
Jul 01 21:34:13 taos-orange5pro systemd[1]: rkllama.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
```

The generated systemd unit starts `rkllama_server` with:

```bash
--preload qwen3-embedding-0.6b,qwen3-reranker-0.6b,qmd-query-expansion
```

However, the current version of `rkllama_server` no longer recognizes the `--preload` argument.

---

## Expected behavior

The installer should generate a systemd service that is compatible with the version of `rkllama` it installs.

If the CLI has changed, the generated `ExecStart` command should be updated accordingly.

---

## Notes

This appears to be a compatibility issue between `install-rknpu.sh` and the current `rkllama` repository. The installer completes successfully, but the generated service cannot start because it passes a command-line argument that is no longer supported.

**Environment:**

- OS: Debian Bookworm (ARM64)
- SBC: Orange Pi 5 Pro 16GB
- Fresh installation
- `rkllama` cloned from the current default branch

jaylfc:
Fixes #1529 (found by @mandresve minutes after working around #1527 manually).

Root cause of #1529: the #1528 re-pin moved `RKLLAMA_REF` to the fork's rewritten main head, which carries the rerank endpoint but silently dropped the `--preload` server option; the systemd unit the installer generates uses `--preload`, so the service exits with `unrecognized arguments` on every fresh install.

Fix, done at the source:
- The original production pin `06cf874d` (preload + rerank + per-model locking + KV-cache fix + worker daemonize fix) is reachable again: `feat/rebase-onto-upstream` in jaylfc/rkllama was fast-forwarded one commit onto it (no history rewrite; that branch's head was the pin's direct parent). Verified with a fresh clone: `git checkout 06cf874d` works and `server.py` has `--preload`.
- `install-rknpu.sh` re-pins to `06cf874d`. The #1528 reachability guard stays, so a future orphaned pin still fails with the curated error.
- CHANGELOG beta.15 entry updated to tell the corrected story (release not yet promoted; #1531 is held for this).

Note for the fork backlog: rkllama main is missing taOS-required patches (`--preload`, and the daemonize fix exists only on the old lineage). Reconciling main is follow-up work in the rkllama repo, not this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the NPU/RKLLAMA installer’s default pinned version for fresh installs.
  * Stale version pins now fail with a clearer, more actionable error and override instructions; refreshed wording also ensures the `rkllama` service remains startable as expected.
  * Improved setup reliability by tightening installed runtime version detection to match `X.Y.Z`.
  * Enhanced server disk free logging to report `unknown` when the disk check fails.
* **Tests**
  * Stabilized an agent registry collision test by freezing time during the run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
 ## [1.0.0-beta.15] - 2026-07-02
 
 ### Fixed
-- The Rockchip NPU installer no longer fails on fresh installs with "reference is not a tree": the rkllama pin now points at a commit that fresh clones can actually check out, and a stale pin fails with a clear explanation and override instructions instead of a raw git error. Reported by @mandresve (#1527).
+- The Rockchip NPU installer no longer fails on fresh installs with "reference is not a tree": the rkllama pin points at the proven production commit (made reachable for fresh clones again), and a stale pin fails with a clear explanation instead of a raw git error. This also keeps the generated rkllama service startable: the newer rkllama default branch dropped the preload option the service relies on. Reported by @mandresve (#1527, #1529).
 - Reconnecting a comms channel (Telegram/Slack/Discord/email/Matrix/webchat) after a token rotation now stops the previous connector instead of silently leaking its background task, concurrent reconnects for the same agent can no longer orphan a connector, and a malformed Matrix reconnect fails cleanly without tearing down the working connector.
 - The LLM proxy self-heal is more robust: it locates the install root at any venv layout, only trusts our own pyproject when doing so, and its pip fallback installs just the proxy requirements instead of re-resolving every dependency.
 - Seeding the internal driver agents now requires naming each pre-existing handle explicitly to adopt it; a blanket adopt flag is rejected so a handle claimed by someone else can never be vouched for as a side effect.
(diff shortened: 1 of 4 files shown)


jaylfc:
This branch now also carries the fix-forward for the three Kilo findings on the merged #1532 (f0b499e9), so the promotion picks up the banner in its hardened form.

jaylfc:
Also carries 1318ce3a: deterministic fix for the flaky test_collision_appends_two_char_suffix, which failed the #1531 promotion re-run (the test and register() minted timestamps on opposite sides of a second boundary on a slow runner, so no collision occurred). The store clock is now frozen in the test so the collision always happens.

Issue #97, opened 2026-04-11
tinyagentos.com: remote desktop proxy at desktop.tinyagentos.com (phase 5)


jaylfc:
Single public URL — `desktop.tinyagentos.com` — that routes each authenticated web session to that user's linked TAOS controller over the Headscale mesh. The proxy is a thin Go service that: (1) terminates TLS at Caddy, (2) validates the incoming Kratos session cookie via `GET /sessions/whoami`, (3) looks up the account's bound controller node and its mesh IP in the orchestrator's `devices` table, and (4) forwards the WebSocket connection to `<mesh-ip>:7070` using the `platform-admin` Headscale identity.

The proxy supports switching between multiple linked controllers via a statusbar dropdown rendered client-side. The dropdown is populated from the orchestrator's `/api/v1/devices` endpoint, filtered to the authenticated account. Switching sends a new routing hint in the session; the proxy re-targets the WebSocket without a full page reload.

The proxy itself carries no session state — it is stateless and can be restarted without dropping in-flight connections (WebSocket hijacking hands off to the forwarded connection). TLS termination is Caddy's responsibility; the proxy listens on a Unix socket behind Caddy.

**Acceptance criteria:**
- [ ] `desktop.tinyagentos.com` requires a valid Kratos session; unauthenticated requests redirect to `/login`
- [ ] Authenticated session routes to the correct controller's `<mesh-ip>:7070` for that account
- [ ] Cross-account routing is impossible: session identity checked against device ownership before forwarding
- [ ] WebSocket connections survive a proxy process restart within the Caddy keepalive window
- [ ] Controller switcher dropdown populates from live orchestrator data, not a stale client cache
- [ ] Proxy runs as a systemd service on the platform LXC, behind the Caddy Unix socket
- [ ] End-to-end test: login → link → open desktop URL → TAOS UI loads from controller

**Depends on:** #95, #96

PR #1536, opened 2026-07-02
fix(lifecycle): resume every paused agent after a controller restart (#97)

jaylfc:
Fixes task #97 (controller update/restart silently pauses running agents), reproduced live minutes after the beta.15 Settings update: agent `naira` (container running) and both workers stuck `paused: true` with zero resume notes on disk.

Root cause: `prepare()` (systemd stop hook) marks EVERY agent `paused=True`, but the boot-time `resume_agents_from_notes` was note-gated, host-gated, and single-shot:
1. A framework that answers `/prepare-for-shutdown` itself (200) leaves no controller-side note, so the resume skipped exactly the agents that implemented the protocol correctly.
2. Hostless agents were never unpaused (nothing to call, flag never cleared).
3. An agent container that boots slower than the controller failed the one unretried attempt and stayed paused, silently.

Fix, all in `restart_orchestrator.py`:
- Resume iterates every `paused=True` agent; a missing note is synthesized (the framework keeps its own state in that case).
- Hostless agents are unpaused directly.
- Unreachable agents get a supervised background retry loop (30s interval, 10 minute window) with a per-agent resumed notification on late success.
- Anything still paused when the window closes raises a WARNING naming the agents, so the failure mode can never be silent again.

Six new tests pin each previously-silent path; 69 pass across the orchestrator-adjacent suites. CHANGELOG Unreleased entry included.

After merge I will unstick the three currently-paused agents on the Pi.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented paused agents from remaining stuck after controller updates/restarts.
  * On boot, automatically resumes paused agents, using any existing resume instructions when available and synthesizing safe defaults when not.
  * Retries resuming agents that aren’t reachable yet, and issues a visible warning if an agent still can’t be resumed after the retry window.
* **Tests**
  * Added coverage for boot-time resume behavior, retry handling, and warning emission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
 
 ## [Unreleased]
 
+### Fixed
+- A controller update or restart no longer silently strands agents in a paused state: agents whose framework handled the shutdown protocol itself, hostless agents, and agents whose containers boot slower than the controller are all resumed at boot (with background retries), and anything that still cannot be resumed raises a visible warning instead of staying paused quietly.
+
 ## [1.0.0-beta.15] - 2026-07-02
 
 ### Fixed
--- a/tests/test_restart_orchestrator_resume.py
+++ b/tests/test_restart_orchestrator_resume.py
@@ -61,12 +61,19 @@ class TestResumeAgentsFromNotes:
 
     @pytest.mark.asyncio
     async def test_unpauses_hostless_agent_directly(self, tmp_path):
+        """Hostless agents unpause without a /resume call, and any note on
+        disk is preserved (nothing consumed it)."""
         agent = {"name": "wkrlan1", "host": "", "paused": True}
         state = _app_state(tmp_path, [agent])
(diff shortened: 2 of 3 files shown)


jaylfc:
Folded all seven findings in the latest commit: hostless notes preserved (with a test), batched single config write on the boot pass, save-before-unlink ordering everywhere, int paused_at matching the controller-note contract, per-attempt exception guard in the retry loop, and pending-only indexing per tick.

One deliberate non-change: the cleartext-HTTP finding. /resume uses the same private controller-to-agent channel as every other agent call, including the /prepare-for-shutdown that delivered the note in the first place; hardening that transport is a real but cross-cutting concern that belongs to its own work, not this fix.

jaylfc:
Round-2 findings folded: failed unpause persistence now warns loudly and preserves the recovery notes (in-memory flags stay flipped on purpose, the agents were genuinely resumed), and the final still-paused warning posts inside its own guard so a notification-store error cannot swallow it. The two remaining suggestions are documented as deliberate in place: per-agent persistence at retry cadence is rare and isolated (the boot pass is the hot path and batches), and the per-tick index rebuild is correct because agents can be added or removed during the 10-minute window.

Issue #844, opened 2026-06-13
Store install chain for rkllama (backend not running) is broken: wrong script + non-interactive false-success


jaylfc:
## Scope
Only the **install-from-scratch** path: a store model install when rkllama is NOT already running, so the resolver takes `action="install_chain"` and installs the backend first. The common case (rkllama running) is handled by the live-probe in #843.

## Two problems in the chain
1. `app-catalog/services/rkllama/manifest.yaml` `install.script` points at `scripts/install-rkllama.sh`, which does not exist. `ScriptInstaller` returns a loud "script not found" (so the chain fails clearly, but never installs rkllama).
2. The real script is `scripts/install-rknpu.sh`, but it **exits 0 without doing anything** in a non-interactive shell unless `TAOS_RKNPU_SETUP=1` (or `--yes`) is set (`confirm_or_exit()`). And `ScriptInstaller.install` runs only `bash <script> <project_dir>` with **no env or extra args**, so it cannot pass the opt-in. Repointing the manifest to `install-rknpu.sh` without that would make `install_chain` mark rkllama installed when it did nothing (a silent false-success, worse than the loud failure). This is why #843 reverted the repoint.

## Fix
- Add env/args support to `ScriptInstaller` (e.g. honor an `install.env: {...}` / `install.args: [...]` block in the manifest), or set the opt-in for this script.
- Repoint the rkllama manifest to `scripts/install-rknpu.sh` and pass `TAOS_RKNPU_SETUP=1`.
- After the script runs, **verify rkllama actually answers** (reuse `rkllama_is_running()`) before reporting success, so a no-op never counts as installed.

## Priority
Low. #843 fixes the real user-facing bug (rkllama running but unregistered). This only matters for a clean Pi where rkllama is not yet installed and the user installs an rkllama model from the store.

Files: `tinyagentos/installers/script_installer.py`, `app-catalog/services/rkllama/manifest.yaml`, `scripts/install-rknpu.sh`, `tinyagentos/catalog/resolver.py`.

Issue #332, opened 2026-05-07
Store: hardware-tier install templates (curated bundles per device class)


jaylfc:
First-time users on a Pi-16GB or a 12GB CUDA box shouldn't have to research and install models one-by-one. A "Templates" section in the Store would offer curated bundles per hardware tier:

- **Pi 16GB starter pack** — chat (Qwen 2.5 3B rkllm), embedding (BGE small), reranker (Qwen 3 Reranker), STT (whisper-tiny), TTS (kokoro), vision (Florence-2)
- **Pi 32GB starter pack** — same shape, larger variants
- **NVIDIA 12GB starter pack** — Qwen 2.5 14B AWQ on vLLM, BGE-M3, etc.
- **Apple Silicon 16GB starter pack** — Qwen 2.5 7B on MLX, etc.
- **CPU-only starter pack** — small quants on llama.cpp

A template references multiple model + service manifests by ID. Click → all install via the resolver chain.

## Setup-wizard integration

The taOS Setup Agent should be able to: detect host + connected workers, infer the right tier, recommend (or auto-apply) a template. The wizard fronts the template store; the agent does the auto-detect.

Builds on PR #325's resolver infrastructure (the chain installer is what makes one-click bundles feasible).

Related: project_setup_agent_research, project_help_app

PR #1537, opened 2026-07-02
feat(setup): NPU-conditional checklist step for the rkllama backend (#1535)

### LINKED ISSUE #1535: Setup wizard never detects the NPU or offers the rkllama backend install
Traced end to end on dev (context: the fresh-install test wave in #2, #1527, #1529).

## Current state

A non-terminal user on a fresh Orange Pi install cannot get the NPU backend without finding it themselves:

- `OnboardingScreen.tsx` is account creation only; `SetupChecklist.tsx` has five hardcoded steps (account, provider, model, agent, memory) and never calls `/api/hardware`.
- `/api/setup/status` (`routes/setup.py`) has no NPU awareness.
- Hardware detection itself works (`hardware.py` `_detect_npu()` correctly detects RK3588 via five probes) and `/api/hardware` exposes it; nothing in the setup flow consumes it.
- The ONLY install path is manual: Store app > find "rkllama (RK3588 NPU LLM)" > Install, which runs `scripts/install-rkllama.sh` > `install-rknpu.sh` (this entry itself only started working on June 14, #844; before that there was no UI path at all, which matches the May report in the #2 thread).
- The cluster deploy endpoint supports `install-rknpu` for remote workers only; there is no local-controller equivalent.

So the original #2-thread gap ("the setup wizard's NPU detection should have queued the rkllama backend install but didn't fire it") was only half-fixed: the Store entry now works, the wizard still does nothing.

## Options (design decision needed)

1. Add an NPU-conditional sixth step to `SetupChecklist.tsx`: call `/api/hardware` on mount, and when `npu.type == "rknpu"`, show "Install the NPU backend" with a click-to-install that triggers the existing Store service install.
2. Auto-queue the rkllama service install in the background when the NPU is detected during setup (zero clicks, but ~2.8 GB of downloads without an explicit yes).
3. A post-onboarding "hardware-optimized setup" surface that recommends per-hardware installs (generalizes to Vulkan/CUDA tiers; overlaps the hardware-tier install templates idea, #332).

Option 1 is the smallest honest fix and keeps consent explicit; option 3 is where this probably wants to end up long-term.

jaylfc:
Implements the recommended option from #1535 (the smallest honest fix, explicit consent kept).

- `/api/setup/status` gains `npu_present` (cached hardware profile, `npu.type == rknpu`) and `npu_backend_running` (live rkllama probe via `rkllama_is_running`, run in a thread so the socket probes stay off the event loop). Probe only runs when an NPU is present.
- `SetupChecklist` appends a sixth step on NPU boards: "Install the NPU backend", detail pointing at the Store (where the existing one-click rkllama install lives, #844), completing once the backend answers. Step count adapts (0/6 etc.). Non-NPU boards render exactly as before.
- Tests: 3 new backend cases (absent by default, present+running, present+not-running) and 3 new component cases (hidden without NPU, shown as sixth item, complete state). 22 + 15 pass; tsc clean.

Closes #1535. The longer-term hardware-optimized setup surface (#332) can grow out of this step later.

## Summary by CodeRabbit

* **New Features**
  * Added an optional “Install the NPU backend” step for compatible Rockchip devices, with a one-click Store flow.
  * Updated the setup status response to report NPU presence and whether the NPU backend is running.
* **Bug Fixes**
  * Improved checklist visibility/completion so the checklist remains until the NPU backend is actually running on NPU devices.
* **Tests**
  * Expanded checklist tests for NPU step visibility, progress counters, and completion behavior.
  * Added API tests validating NPU fields for default and Rockchip scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
 
 ## [Unreleased]
 
+### Added
+- On Rockchip boards the setup checklist now includes an "Install the NPU backend" step: it appears only when an NPU is detected, opens the Store where the rkllama backend installs with one click, and ticks itself once the backend is running. Previously nothing in the setup flow ever surfaced the NPU install.
+
 ## [1.0.0-beta.15] - 2026-07-02
 
 ### Fixed
--- a/desktop/src/components/SetupChecklist.test.tsx
+++ b/desktop/src/components/SetupChecklist.test.tsx
@@ -175,4 +175,32 @@ describe("SetupChecklist", () => {
     render(<SetupChecklist />);
     expect(await screen.findByRole("list")).toBeInTheDocument();
   });
+
+  it("hides the NPU step on boards without an NPU", async () => {
+    mockFetchStatus(baseStatus);
+    render(<SetupChecklist />);
+    await screen.findByText("Get started");
+    expect(screen.queryByText("Install the NPU backend")).toBeNull();
+    expect(screen.getByText("0/5")).toBeInTheDocument();
+  });
+
+  it("shows the NPU step as a sixth item when an NPU is present", async () => {
+    mockFetchStatus({ ...baseStatus, npu_present: true, npu_backend_running: false });
+    render(<SetupChecklist />);
+    expect(await screen.findByText("Install the NPU backend")).toBeInTheDocument();
+    expect(
+      screen.getByText("Install rkllama from the Store to run models on this device's NPU"),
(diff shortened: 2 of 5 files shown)


jaylfc:
Folded all four findings: the checklist now stays visible when core steps are complete but the NPU step is outstanding (this was a real gap, thanks CodeRabbit; dismiss still hides it and it disappears once the backend runs, both pinned by component tests), the status docstring documents the new fields, the absent-case test pins hardware_profile so class order cannot leak the Rockchip profile, and the to_thread comment now states the blocking rationale precisely. 22 + 17 tests pass.

Issue #1548, opened 2026-07-02
taOS Desktop cannot install models after a fresh installation (downloads complete instantly, models are not available)


mandresve:
## Environment

- TinyAgentOS freshly installed
- Debian Bookworm
- Orange Pi 5 Pro (RK3588S)

The installation completed successfully after resolving the installer issues reported in previous issues.

## Problem

After opening taOS Desktop for the first time, I attempted to install a model (Qwen) from the Models section.

The installation does not appear to complete successfully.

### Observed behavior

- Clicking **Install** immediately advances the progress bar to completion.
- The download appears to finish almost instantly.
- However, the model is never actually available afterwards.
- Reopening the Models page still shows the model as not installed.

It looks as if the UI reports a successful installation even though no model has actually been downloaded or registered.

## Expected behavior

- Installing a model should display actual download progress.
- The model should appear as installed once the download completes.
- Providers should initialize correctly after a successful installation.
- The built-in log viewer should expose any backend/frontend errors to aid troubleshooting.

At the moment it is difficult to determine the root cause because the UI provides no visible errors and the integrated log viewer does not contain any diagnostic information.

mandresve:
## Summary

This issue is still present in **Beta 17**.

When attempting to download a model from the **Models** dialog in taOS Desktop, the download progress bar fills almost immediately, giving the impression that the model has been successfully installed.

However, no actual download seems to occur.

## Steps to reproduce

1. Perform a clean installation.
2. Open **taOS Desktop**.
3. Open the **Models** dialog.
4. Select any model (tested with Qwen).
5. Click **Download**.

## Expected behavior

- The progress bar should reflect the real download progress.
- The model should be downloaded to disk.
- The model should remain installed after closing and reopening the Models dialog.

## Actual behavior

- The progress bar reaches 100% almost instantly.
- The model is shown as if it were installed.
- Closing and reopening the Models dialog shows that **no models are actually installed**.
- No files appear to have been downloaded.

## Screenshot

Attached is a screenshot showing the model appearing as successfully downloaded even though no download actually took place and nothing is persisted after reopening the dialog.

<img width="905" height="324" alt="Image" src="https://github.com/user-attachments/assets/dbf44a31-cbc7-40e1-b333-11d1c04f5ddb" />

mandresve:
I have now confirmed the **same behavior on beta.18** after another completely fresh installation.

The model download issue is still present and behaves exactly as in previous builds:

[...]

PR #1549, opened 2026-07-02
fix(models): surface download failures instead of silently clearing them (#1548)

jaylfc:
Addresses the silent-failure half of #1548 (clean Debian Bookworm vendor image, Orange Pi 5 Pro): clicking Install appeared to complete instantly but the model never installed.

Root cause of the confusing UX: in ModelBrowser the download poll loop deleted the progress entry when the task reported status=error, and discarded the backend's task.error string; a rejected download-start did the same. An immediate failure therefore rendered exactly like an instant success.

Now:
- a failed download stays on the variant row with the backend's real error text (checksum mismatch, unreachable host, permission error, and so on) and a Retry button
- a download that cannot even be started shows the API's error instead of vanishing
- two vitest cases pin both paths

This intentionally does not guess at the underlying download failure on the reporter's board: with the error finally visible, their next attempt tells us the exact cause. The Providers-page error states and the empty log viewer from #1548 are being handled separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Failed model downloads now remain on the model card with the backend-provided error reason, instead of disappearing.
  * Added clearer failure states, including an inline error message and a **Retry** button when downloads fail or can’t be started.
  * Improved resilience when polling download status to avoid leaving silent or confusing states.
* **Tests**
  * Added UI tests covering model download error handling and **Retry** behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
 
 ## [Unreleased]
 
+### Fixed
+- Models: a failed model download no longer looks like an instant successful install. The failure now stays on the model card with the actual cause from the backend (for example a checksum mismatch or an unreachable download host) and a Retry button, instead of the progress UI silently disappearing (#1548).
+
 ## [1.0.0-beta.16] - 2026-07-02
 
 ### Added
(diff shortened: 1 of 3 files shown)


mandresve:
At this point, I think the installation is in a good enough state to start debugging the system itself rather than the installer.

I'm happy to continue testing on this existing installation and help isolate runtime issues.

If at any point it's useful to validate the complete workflow again, I have no problem performing a second clean installation from scratch and re-testing everything end-to-end, from the installer through the initial user experience and desktop usability.

Issue #1584, opened 2026-07-03
TextEditor is non-functional: cannot create notes or documents


mandresve:
## Description

The **TextEditor** application appears to be completely non-functional.

None of the available actions for creating a new document respond when clicked, making it impossible to create or edit notes.

## Environment

- TinyAgentOS: **beta.20**
- OS: **Debian 12 (Bookworm)**
- Hardware: **Orange Pi 5 Pro (RK3588S)**

## Steps to reproduce

1. Open **TextEditor**.
2. Attempt to create a new note or document using any of the available **New** buttons.
3. Repeat with each available creation option.

## Observed behavior

- None of the creation buttons perform any action.
- No new note or document is created.
- No dialog opens.
- No error message is displayed.
- The application remains idle regardless of which button is pressed.

From the user's perspective, the application appears completely unresponsive.

## Expected behavior

Clicking any of the available **New** actions should create a new document (or present the appropriate dialog) so the user can begin editing.

jaylfc:
Fixed in beta.21. The editor used a browser API that only exists over HTTPS or localhost, so over plain http it threw before creating anything, with no visible error. That is fixed, and the same class of bug was hardened across a few other spots. Creating notes and documents works again.

PR #1607, opened 2026-07-03
fix(texteditor): stop focus loss per keystroke (#1596)

### LINKED ISSUE #1596: TextEditor loses focus after every keystroke (crypto.randomUUID is not a function in logs)
## Description

After updating to **beta.21**, the previous issue preventing new file creation has been resolved. However, a new issue remains that makes the TextEditor unusable for editing.

## Steps to reproduce

1. Launch **TextEditor**.
2. Create a new note.
3. Click inside the editor and begin typing.

## Observed behavior

- The editor accepts **only a single character**.
- Immediately after typing one character, the editor loses focus (or the editing session is interrupted).
- No additional characters can be entered until the user clicks inside the text area again.
- After clicking, exactly **one more character** can be typed before the same behavior repeats.
- This behavior is **100% reproducible**.

## Expected behavior

The editor should retain keyboard focus and allow continuous typing without interruption.

## Additional information

The application logs consistently report the following JavaScript exception while this issue occurs:

```text
Uncaught TypeError: crypto.randomUUID is not a function
```

This may be related to the problem, although I cannot confirm whether it is the root cause.

jaylfc:
## Summary
- Fixes #1596: TextEditor lost keyboard focus after every keystroke, forcing a re-click to type each character.
- Root cause: `MarkdownEditor`'s CodeMirror mount `useEffect` in `desktop/src/apps/TextEditorApp.tsx` depended on the `content` prop. Every keystroke flows through `onChange -> setNotes -> activeNote.content` back into that prop, so the effect re-fired on every character, destroying and rebuilding the entire `EditorView`. The freshly created view is never focused, so the editor dropped focus after each character typed.
- Note switching is already handled by the stable `key={activeNote.id}` on `<MarkdownEditor>` at the call site (it fully remounts the component), so the internal effect only needs to run once per mount — it no longer needs (or should) depend on `content`.
- I could not reproduce the `crypto.randomUUID is not a function` error from the issue logs against current `dev`: the only `crypto.randomUUID()` call in this file is already guarded (`newNoteId`, used only by `createNote`), and a repo-wide grep found no other unguarded call site reachable from typing. Prior PRs #1584/#1585 (create path) and #1587 (assistant/push/webstudio) already closed those. The logged error was most likely from a stale cached bundle predating those fixes.

## Test plan
- [x] Added a regression test that shadows `crypto.randomUUID` to `undefined` (matching taOS's plain-http deployment), creates a note, then dispatches five keystrokes one at a time against the live CodeMirror view — asserting no throw and that the exact same `EditorView` instance/DOM node stays mounted and focused after every character.
- [x] Verified the new test fails against the pre-fix code (reverted the dependency array locally) and passes after the fix, confirming it captures the real regression.
- [x] `cd desktop && npm run build` — clean, exit 0.
- [x] `npx vitest run src/apps/__tests__/TextEditorApp.test.tsx` — 3/3 passed.
- [x] `npx vitest run src/lib/uid.test.ts` — 3/3 passed (unaffected, sanity check).

https://github.com/jaylfc/taOS/issues/1596

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the text editor experience so the editor stays mounted and keeps focus while typing, preventing interruptions between keystrokes.
  * Preserved note content more reliably during editing, with changes continuing to save correctly as you type.

* **Tests**
  * Added coverage for continuous typing behavior to ensure the editor remains stable and focused across updates.

### DIFF
--- a/desktop/src/apps/TextEditorApp.tsx
+++ b/desktop/src/apps/TextEditorApp.tsx
@@ -180,7 +180,16 @@ const MarkdownEditor = React.forwardRef<
       view.destroy();
       viewRef.current = null;
     };
-  }, [content]); // re-create when content identity changes (note switch)
+    // Mount once per component instance. The parent already remounts this
+    // component (fresh `key={activeNote.id}`) when the active note changes,
+    // so a new view with the right initial doc is created on note switch.
+    // Depending on `content` here as well made the effect re-fire on every
+    // keystroke (onChange -> setNotes -> content prop changes), destroying
+    // and recreating the CodeMirror view mid-edit. The new view isn't
+    // focused, so the editor dropped keyboard focus after each character
+    // typed (#1596).
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   return <div ref={containerRef} className="h-full w-full" />;
 });
--- a/desktop/src/apps/__tests__/TextEditorApp.test.tsx
+++ b/desktop/src/apps/__tests__/TextEditorApp.test.tsx
@@ -1,6 +1,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
-import { render, screen, fireEvent } from "@testing-library/react";
+import { render, screen, fireEvent, act } from "@testing-library/react";
 import React from "react";
+import { EditorView } from "@codemirror/view";
 
 import { TextEditorApp } from "../TextEditorApp";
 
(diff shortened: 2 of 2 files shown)


jaylfc:
Duplicate of #1606 (same root cause and fix for #1596: the MarkdownEditor mount effect keyed on [content] remounted CodeMirror per keystroke). Keeping #1606, which also consolidates the duplicate note-id helper onto the shared randomId. Closing this one.

Issue #160, opened 2026-04-12
Gaming PC GPU/AI co-scheduling: opt-in per-worker mode that drains AI jobs when gaming load spikes


jaylfc:
Gaming PCs on the cluster contribute significant GPU compute when idle, but become useless for AI inference mid-game when GPU utilization spikes. Without awareness of gaming workloads, TAOS may route expensive AI jobs to a machine that is about to become fully occupied by a game, causing timeouts and degraded quality.

This issue adds an opt-in gaming-detection mode for individual workers. When enabled on a specific worker, TAOS monitors GPU utilization and drains AI jobs from that node when a gaming workload is detected. Drained jobs are rerouted to other available nodes in the cluster.

**This feature must be opt-in per worker, configured explicitly in the worker settings. It must never be auto-enabled or auto-detected across all machines.** A worker with gaming-detection off behaves exactly as today with no background monitoring.

The detection heuristic can start simple: sustained GPU utilization above a configurable threshold for a configurable window (e.g. above 80% for 30 seconds). A user can also manually declare specific process names as gaming triggers (e.g. `steam.exe`, `GameOverlayRenderer.exe`, or a specific game binary).

**Acceptance criteria:**
- [ ] Per-worker setting `gaming_detection_enabled: bool` (default: false) in worker config
- [ ] Setting is only configurable per-worker in that worker's settings page; there is no global toggle
- [ ] When enabled: worker monitors GPU utilization on a configurable polling interval (default 10s)
- [ ] Configurable threshold: `gaming_drain_gpu_pct` (default 80) and `gaming_drain_window_s` (default 30)
- [ ] Optional: `gaming_process_names` list of process names that immediately trigger drain regardless of GPU %
- [ ] When drain triggers: worker signals controller to stop routing new jobs to it; in-flight jobs continue to completion or are migrated if they exceed a timeout
- [ ] When GPU utilization drops below threshold for `gaming_drain_window_s`, worker re-advertises capacity and resumes accepting jobs
- [ ] Dashboard shows gaming-detection state per worker: `active (gaming)`, `active (idle)`, `disabled`
- [ ] Workers without gaming-detection enabled show no monitoring overhead whatsoever

jaylfc:
Implemented in the scheduling package:

- `GamingDetector` — scans `/proc` for known game engine processes, checks X11 fullscreen state, monitors GPU utilisation via nvidia-smi. Auto-yields worker resources when gaming detected.
[...]

PR #1631, opened 2026-07-04
feat(governance): execution policies + approval gates for agent tool calls (#160 slice 1)

jaylfc:
## Summary

Slice 1 of agent governance: a policy layer that decides whether a deployed agent's tool call is **allowed**, **denied**, or **needs human approval**, wired into the existing skill-exec + Decisions plumbing.

- **`tinyagentos/governance/policy_store.py`** — `ExecutionPolicyStore` (follows `decision_store.py`'s shape). Two tables:
  - `execution_policies`: `(scope['global'|'agent'], agent_name, action_class, effect['allow'|'deny'|'require_approval'])`. A per-agent row overrides the global row for the same action class; absent either, a hardcoded conservative fallback applies.
  - `execution_grants`: a short-lived `(agent_name, action_class, expires_at, decision_id)` row. A live (unexpired) grant satisfies a `require_approval` gate so the agent's retry passes without asking again. TTL is 15 minutes.
  - `_post_init` seeds the **conservative default posture** once: global `require_approval` for `code-exec`, `external-network`, `deploy`, `file-write-external`, `spend`. Everything else is allowed by the *absence* of a policy row.
- **`tinyagentos/governance/action_classes.py`** — a small dict mapping today's actual `SKILL_IMPLEMENTATIONS` ids to action classes: `code_exec -> code-exec`, `http_request -> external-network`, `web_search -> external-network`. Unclassified skills (file_read/file_write/etc., which are already sandboxed to the agent's own workspace) return `None` and are never gated.
- **Enforcement hook in `skill_exec.py::execute_skill`** — runs strictly *after* `_is_admin_or_local_token` (the RCE-fix auth gate) and *before* `impl(...)`:
  - `allow` → proceeds unchanged.
  - `deny` → `403 forbidden_by_policy`, a fail-soft receipt (`stop_reason=policy_deny`) and a `board_audit` record, tool never runs.
  - `require_approval` with no live grant → creates a blocking `approve_deny` Decision (`metadata.kind = "execution_gate"`), emits a receipt (`human_approval=pending`, `decision_id`, `stop_reason=awaiting_approval`), returns `202 pending_approval` — tool never runs.
  - `require_approval` with a live grant → proceeds.
- **Approval side effect in `decisions.py::answer_decision`** — new `_apply_execution_grant` (mirrors `_apply_app_grant`'s structure): when an `execution_gate` decision is answered `"approve"`, writes the grant via `execution_policies.add_grant(...)`; either way it notifies the agent (reusing `_route_answer_to_agent`) that it may retry or was denied. Best-effort — a grant-store hiccup never fails the already-persisted answer.

### Placement relative to the auth gate (unchanged, first check)

```python
if not _is_admin_or_local_token(request):
    return JSONResponse({"error": "forbidden"}, status_code=403)
...
impl = SKILL_IMPLEMENTATIONS.get(skill_id)
if not impl:
    return JSONResponse({"error": f"No implementation for {skill_id}"}, status_code=501)

# Agent governance (#160 slice 1): runs after the auth gate above, before
# the skill actually executes.
agent_name = (args.get("agent_name") or "").strip()
gate_response = await _check_execution_policy(request, skill_id, agent_name)
if gate_response is not None:
    return gate_response

try:
    result = await impl(args, request)
```

### Deferred to later slices

- Userspace-app enforcement (this slice only covers the agent-tool / skill-exec path).
- Budgets / spend limits.
- Agent lifecycle (pause/kill on repeated denial, etc.).
- A management UI for policies (CRUD methods exist on the store; no route/UI yet).

### Behavioral note

Because the conservative default applies to *every* caller (not just non-admins), two pre-existing `code_exec` tests in `tests/test_skill_exec_authz.py` (which assert immediate `200` execution for an admin session / valid local token) now need a `set_policy("code-exec", "allow")` in their setup to neutralize the new, orthogonal governance gate — the auth-gate assertions themselves are untouched, and new regression tests (`TestAuthGateStillFirst` in `tests/test_skill_exec_governance.py`) independently verify the auth gate still runs first and blocks non-admins even with a live grant present.

## Test plan

- [x] `python -m pytest tests/governance/ tests/test_skill_exec_governance.py tests/test_routes_decisions.py -q` — 48 passed
- [x] `python -m pytest tests/ -k "skill_exec or decision or governance or auth" -q --ignore=tests/e2e` — 564 passed, 1 skipped
- [x] `python scripts/check_doc_gate.py diff-gate --staged` — clean (no rule fired for the new files)

## Summary by CodeRabbit

* **New Features**
  * Added execution-governance checks for skill calls, including approval, deny, and pending-approval outcomes.
  * Added support for short-lived execution approvals so approved actions can retry without re-approval.
  * Expanded decision handling so approving an execution-gate decision can grant future execution.

* **Bug Fixes**
  * Strengthened policy precedence so agent-specific rules override global defaults.
  * Preserved existing authentication checks ahead of governance checks for protected actions.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -321,6 +321,10 @@ async def client(app, tmp_data_dir):
     if decision_store._db is not None:
         await decision_store.close()
     await decision_store.init()
+    execution_policies = app.state.execution_policies
+    if execution_policies._db is not None:
+        await execution_policies.close()
+    await execution_policies.init()
     coding_session_store = app.state.coding_session_store
     if coding_session_store._db is not None:
         await coding_session_store.close()
--- a/tests/governance/test_policy_store.py
+++ b/tests/governance/test_policy_store.py
@@ -1,0 +1,112 @@
+import pytest
+import pytest_asyncio
+
+from tinyagentos.governance.policy_store import ExecutionPolicyStore
+from tinyagentos.governance.action_classes import SENSITIVE_ACTION_CLASSES, classify
+
+
+@pytest_asyncio.fixture
+async def store(tmp_path):
+    s = ExecutionPolicyStore(tmp_path / "execution_policies.db")
+    await s.init()
+    yield s
+    await s.close()
+
+
+class TestActionClasses:
+    def test_code_exec_classified(self):
+        assert classify("code_exec") == "code-exec"
+
+    def test_http_request_classified(self):
+        assert classify("http_request") == "external-network"
+
+    def test_unknown_skill_unclassified(self):
+        assert classify("file_read") is None
+        assert classify("nonexistent_tool") is None
+
+
+@pytest.mark.asyncio
+class TestConservativeSeed:
+    async def test_seeded_global_rows_require_approval(self, store):
+        policies = await store.list_policies()
+        seeded_classes = {p["action_class"] for p in policies if p["scope"] == "global"}
+        assert seeded_classes == set(SENSITIVE_ACTION_CLASSES)
+        assert all(p["effect"] == "require_approval" for p in policies)
+
+    async def test_sensitive_classes_default_to_require_approval(self, store):
+        for action_class in SENSITIVE_ACTION_CLASSES:
+            assert await store.effective_effect("agent-a", action_class) == "require_approval"
+
+    async def test_unclassified_action_class_defaults_to_allow(self, store):
+        assert await store.effective_effect("agent-a", "some-harmless-class") == "allow"
+
+    async def test_reinit_does_not_duplicate_seed(self, tmp_path):
+        db_path = tmp_path / "policies.db"
+        s1 = ExecutionPolicyStore(db_path)
+        await s1.init()
+        await s1.close()
+        s2 = ExecutionPolicyStore(db_path)
+        await s2.init()
+        policies = await s2.list_policies()
+        assert len(policies) == len(SENSITIVE_ACTION_CLASSES)
+        await s2.close()
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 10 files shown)


jaylfc:
Follow-up commit `a05bca78` applies two product-default decisions from the owner:

**1. Narrowed the conservative default.** `web_search` is removed from `TOOL_ACTION_CLASSES` — it's a read-only lookup with no host/write blast radius, so `classify("web_search")` now returns `None` (always allowed). The conservative seed (code-exec, external-network, deploy, file-write-external, spend) is unchanged. Net effect: today only `code_exec` and `http_request` are actually gated.

**2. Admin human sessions bypass governance.** A new guard at the very top of `_check_execution_policy` returns `None` (no gating) when the caller is an admin human session (`via == "session" and is_admin`). A deployed agent (`via == "local_token"`) is still fully governed. So enforcement only ever applies to agent callers — the human is the approval authority and isn't asked to self-approve.

Tests updated to match: all gating assertions now run through a local-token (agent) client, since the admin-session fixture bypasses governance; added an admin-bypass-vs-local-token-gated pair; `test_admin_session_allowed` no longer needs a policy workaround (`test_valid_local_token_allowed` keeps it, as local-token callers remain governed).

Verification:
- `pytest tests/governance/ tests/test_skill_exec_governance.py tests/test_skill_exec_authz.py tests/test_routes_decisions.py -q` → 61 passed
- `pytest tests/ -k "skill_exec or decision or governance or auth" -q --ignore=tests/e2e` → 567 passed, 1 skipped

PR #1671, opened 2026-07-06
feat(devices): taOS Apple client — Slice 1 server foundation (device registry + scoped tokens + APNs)

jaylfc:
First slice of the native iOS/watchOS client (design + plan in docs/superpowers). This is the **server-side foundation** the app builds on: pure Python, runs in the existing pytest CI, no Mac needed. Implemented subagent-driven, one task per commit.

## What
- **DeviceStore** (`tinyagentos/device_store.py`) - per-user device registry keyed on `device_id`, each with a `platform`, `push_token`, and a per-device `scoped_token` (`taosdev_` + `secrets.token_urlsafe`). Revocable; `get_by_token` refuses revoked devices; `list_for_user` hides the token.
- **Device routes** (`tinyagentos/routes/devices.py`) - `POST /api/devices/register` (returns the scoped_token once, for Keychain), `GET /api/devices`, `PATCH .../push-token`, `DELETE` (revoke). Scoped to the session user; `user_id` never trusted from the body. **Devices are strictly personal - no admin bypass** (each holds a scoped token + its owner's future sensor grants).
- **require_device auth** (`tinyagentos/device_auth.py`) - a dependency that authenticates a request as a device by its bearer scoped token (401 on missing/unknown/revoked, touches `last_seen`). This is what the app authenticates as, not the master credential. Later slices (sensor-pull) depend on it.
- **APNs sender** (`tinyagentos/push/apns.py`) - `ApnsSender` protocol + `NullApnsSender` default (so the system runs with no Apple creds) + `HttpApnsSender` (ES256 provider JWT via the existing `cryptography` dep, HTTP/2 to Apple). Pure JWT/payload builders are unit-tested; the network send is tested against `httpx.MockTransport`. Selected via `apns_sender_from_env()`.

## Tests
17 new tests, all green (`test_device_store`, `routes/test_devices`, `test_device_auth`, `push/test_apns`). App constructs cleanly.

## Not in this slice (later slices/plans)
iOS shell + watchOS app, the capability/sensor-pull protocol, and the Pro relay. Off-LAN connectivity is confirmed to ride the existing Headscale + taOSgo cluster-join (coordinated with @taOS-website-dev), so no new relay server surface is needed.

## Notes
- The lifespan-bypassing test `client` fixture now inits `device_store` (it hand-inits lifespan-owned stores; this was the first route-tested one missing).
- Fixed a latent wiring bug: `app.state.apns_sender` is assigned where `app` actually exists (mirroring `device_store`), not before its construction.

## Summary by CodeRabbit

* **New Features**
  * Added device self-service management: register, list, update push tokens, and revoke devices (with per-user device cap).
  * Added APNs push support with automatic sender selection from environment (silent/background support included).

* **Bug Fixes**
  * Strengthened device authorization so only the owning, non-revoked device can be modified.
  * Prevented secret `scoped_token` from appearing in list/update responses.

* **Tests**
  * Added/expanded async tests for device store, device auth helpers, device routes security, and APNs payload/sender behavior (including HTTP request assertions).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -372,6 +372,10 @@ async def client(app, tmp_data_dir):
     if client_log_store._db is not None:
         await client_log_store.close()
     await client_log_store.init()
+    device_store = app.state.device_store
+    if device_store._db is not None:
+        await device_store.close()
+    await device_store.init()
     # BrowserApp v2 stores
     from tinyagentos.routes.desktop_browser.store import BrowserStore, BrowserCookieStore
     _browser_store = BrowserStore(tmp_data_dir / "browser.sqlite3")
--- a/tests/push/test_apns.py
+++ b/tests/push/test_apns.py
@@ -1,0 +1,78 @@
+import json
+import httpx
+import pytest
+
+from tinyagentos.push.apns import (
+    NullApnsSender,
+    HttpApnsSender,
+    build_apns_payload,
+    build_apns_jwt,
+    apns_sender_from_env,
+)
+
+
+def test_build_payload_alert_and_silent():
+    alert = build_apns_payload(title="Hi", body="there", data={"k": "v"})
+    assert alert["aps"]["alert"] == {"title": "Hi", "body": "there"}
+    assert alert["k"] == "v"
+
+    silent = build_apns_payload(title="", body="", content_available=True)
+    assert silent["aps"]["content-available"] == 1
+
+
+@pytest.mark.asyncio
+async def test_null_sender_returns_false():
+    assert await NullApnsSender().send("tok", {"aps": {}}) is False
+
+
+def test_apns_sender_from_env_defaults_to_null(monkeypatch):
+    for var in ("TAOS_APNS_KEY_ID", "TAOS_APNS_TEAM_ID", "TAOS_APNS_BUNDLE_ID", "TAOS_APNS_KEY_PATH"):
+        monkeypatch.delenv(var, raising=False)
+    assert isinstance(apns_sender_from_env(), NullApnsSender)
+
+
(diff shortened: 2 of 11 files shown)


jaylfc:
Kilo re-review triaged. The five prior real findings are folded. On the new pass:

**CRITICAL (device_auth.py:36, `last_seen` TypeError) — verified FALSE POSITIVE.** `last_seen` is declared `INTEGER`, and SQLite's INTEGER type affinity converts the `strftime('%s','now')` string to an int on storage, so it reads back as a Python `int`. Proven empirically: a freshly registered device returns `last_seen=1783348826` (type `int`), and `time.time() - last_seen` yields a float. The existing `test_device_auth.py` tests already execute that exact subtraction and pass — a real TypeError would 500 them. No change needed. (This is Kilo's second false-positive CRITICAL on this PR, after the APNs `exp` one.)

**Residuals, accepted with reasoning:**
- Device-cap TOCTOU: the cap is a soft DoS/abuse guardrail, not a hard invariant. A concurrent-registration race can at worst let a user land 1-2 devices over 50, which confers no benefit — acceptable soft bound.
- `.p8` warn-not-refuse and the `{1,true,yes}` truthy set: defensible choices for scaffolding (the sender defaults to NullApnsSender and isn't wired to real sends until the pull slice). Noted for the APNs hardening pass in a later slice.
- COUNT(*) vs list, `{error}` vs `{detail}`: micro-optimizations / matches the existing `decisions.py` convention.

Merging: all CI green (tests 3.12/3.13, spa-build, lint, doc-gate, cla), CodeRabbit + Gitar pass, and the one blocker is a verified false positive.

Issue #893, opened 2026-06-14
Shared-hardware (GPU) coordination protocol between agents over A2A


jaylfc:
## Need (Jay 2026-06-14)
@taOS and @taOSmd share hardware (Fedora RTX 3060 on linstation, 12GB GPU). They must coordinate over A2A: post when claiming/using, post when free again, CHECK before use. Found when @taOS FLUX image-gen OOMd because @taOSmd had ~9.4GB of Ollama models loaded on the same GPU despite an earlier free signal.

## Interim protocol (agents follow now via the A2A integration channel)
GPU LEASE for a shared node:
1. CHECK before loading: scan the channel for an open `[GPU CLAIM]` on that node AND run nvidia-smi. Load only if free + unclaimed.
2. CLAIM: `[GPU CLAIM] node=<host> holder=@you vram=~Xgb reason=... eta=...` before loading.
3. RELEASE: `[GPU RELEASE] node=<host> holder=@you` immediately when done.
4. REQUEST when blocked: `[GPU REQUEST] node=<host> need=~Xgb`; holder releases or negotiates a split/window.
5. Courtesy: do not hold the GPU idle; use keep-alive auto-unload.

## Future (productize)
Fold into the cluster scheduler/ClusterManager as a real resource lease: per-device VRAM accounting, a lease registry, admission control (a backend wont load if it cannot fit alongside current leases), auto-eviction by keep-alive. The A2A text protocol is the stop-gap. Relates to #890/#892 (worker contract) and dynamic resource scheduling.

## Acceptance
Both agents check + claim + release over A2A and never silently co-load past the GPU VRAM; then a scheduler-enforced version automates it.

hognek:
Claiming for @hognek / skald-engineer-taos. GPU lease/coordination protocol. Depends on #2112 (the bus must be authenticated first). Will branch from origin/dev.

hognek:
Draft → ready: PR #2988 implements the GPU lease/coordination protocol over the A2A bus.

https://github.com/jaylfc/taOS/pull/2988

- `tinyagentos/gpu_lease.py` — the CHECK/CLAIM/RELEASE/REQUEST text protocol: parse/render + fold a channel's history into the claims still open.
- `tinyagentos/routes/a2a_gpu_lease.py` — `/api/a2a/gpu/{check,claim,release,request,renew}`, agent-token scoped (`a2a_receive` to check, `a2a_send` to act).
[...]

PR #1675, opened 2026-07-06
feat(cluster): GPU lease claim/release API for shared coordination (#893)

hognek:
Prototype: GPU lease claim/release API on taOS controller. See docs/design/a2a-gpu-lease-proposal.md for full RFC.

----
## Summary by Gitar

- **Cluster API:**
  - Added `/api/cluster/leases/claim`, `/release`, `/renew`, and `/list` to enable GPU capacity reservations.
  - Added `GpuLease` tracking to `ClusterManager` with automatic expiry via `_monitor_loop`.
- **Worker heartbeats:**
  - Extended heartbeat protocol to include real-time `free_vram_mb` and `used_vram_mb`.
  - Added `gpu_vram_snapshot` probe for workers to report VRAM usage via `nvidia-smi`.
- **Model management:**
  - Implemented `model_archive` to support promoting archived models when hardware requirements are met.
  - Added `GET /api/cluster/promote-archived` for manual compatibility-based model promotion.
- **Testing:**
  - Added `tests/test_leases.py` for lease lifecycle and `tests/test_model_archive.py` for promotion logic.

### DIFF
--- a/docs/design/a2a-gpu-lease-proposal.md
+++ b/docs/design/a2a-gpu-lease-proposal.md
@@ -1,0 +1,326 @@
+# RFC: Shared-GPU Lease Coordination on the TAOS Controller (#893)
+
+**Author:** Hogne (Skald platform, taOS fork)  
+**Date:** 2026-07-06  
+**Status:** Draft — seeking jaylfc buy-in before prototyping  
+**Target:** taOS controller endpoints + Skald TaosDispatcher integration
+
+---
+
+## Problem
+
+Skald's `TaosDispatcher._trigger_load()` calls `POST /api/system/load-model` on the
+Skald server — a heavyweight operation that kills running processes, waits for VRAM
+to drain, and starts a new backend process. The dispatcher fires this *before checking
+whether the target worker has enough free VRAM*.
+
+taOS now surfaces real-time VRAM per worker (`free_vram_mb` / `used_vram_mb` on
+`/api/cluster/workers`, shipped in #894), but that data is a **snapshot**, not a
+**reservation**. Two concurrent requests can both see "8 GiB free", both fire
+load-model, and one loses the race — the loser's load fails or (worse) OOM-kills
+the winner.
+
+The gap: **no coordination primitive on the controller** that lets an external
+caller (Skald, any A2A agent, future tool servers) atomically *reserve* GPU capacity
+before committing to a load.
+
+## Proposal: Lease API on `/api/cluster`
+
+Add two endpoints to the taOS controller:
+
+### `POST /api/cluster/leases/claim`
+
+**Request body:**
+
+```json
+{
+  "resource_id": "worker-name:gpu-cuda-0",
+  "ttl_seconds": 30,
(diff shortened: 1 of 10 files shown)


jaylfc:
This is a great RFC, thank you. The problem is real (a VRAM snapshot is not a reservation, so concurrent load-model calls race to OOM) and a lease primitive is the right shape. I am in on the direction. A few things to land it:

**1. Retarget to `dev`.** `master` is the stable release branch; all work goes through `dev` first (the bot flagged this too).

**2. The claim must actually be atomic.** Right now `claim_lease` (manager.py:279) documents an asyncio lock but does not hold one, so the existing-lease check and the VRAM check can interleave between two callers, which is exactly the race the lease is meant to remove. The whole claim (check existing lease -> check free VRAM -> record the lease) has to run under a single lock, or the primitive does not do its job.

**3. Distinguish "no VRAM probe" from "zero free VRAM".** The heartbeat sends literal `0` for `free_vram_mb`/`used_vram_mb` when `nvidia-smi` is unavailable (agent.py:421-422). On taOS that breaks every non-NVIDIA worker: the RK3588 boards, Apple Silicon, and CPU-only nodes have no `nvidia-smi`, so they would report 0 free and refuse every claim. Send `null` / omit the field when there is no probe, and treat "unknown VRAM" as "do not gate on VRAM" rather than "no capacity". This one matters even on its own.

**4. A few more from the review worth folding:** wrap the `nvidia-smi` probe in a timeout (worker_capacity.py:102 - a hung probe stalls the heartbeat loop); the `resource_id` worker match uses `startswith(name + ":")` (manager.py:474) which collides `gpu-node` with `gpu-node-2`, so match the worker name exactly; and the MLX `gpu_accel` branch is missing (model_archive.py:78) so MLX-only models silently match any worker.

**5. Keep it a taOS primitive, not a Skald-specific one.** The lease is useful to taOS's own cluster scheduler and any A2A agent, not just the Skald dispatcher. The `caller` field already generalizes it; just make sure nothing downstream assumes a Skald caller.

Land those (especially 2 and 3), resolve the master conflicts, and I will review again. Solid contribution, and exactly the kind of cluster-coordination piece taOS needs.

jaylfc:
On reflection I will take this the rest of the way rather than bounce it back. It is a solid prototype and, as maintainer, the fixes are mine to make. I will retarget to `dev`, make the claim atomic under the lock, fix the VRAM-0-vs-unavailable bug so non-NVIDIA workers (RK3588, Apple Silicon, CPU-only) are not broken, fold the rest of the review, and resolve the conflicts. Your commits and authorship stay on the PR. Thanks for the RFC and the working prototype: this is exactly the kind of upstream contribution taOS wants.

jaylfc:
Continued and finished in #1680 - your three commits and authorship are carried over, plus the maintainer fixes (atomic claim, VRAM unknown-vs-zero so non-NVIDIA workers stay leasable, dropped the model-archive half since dev already has it). Closing this in favour of #1680. Thanks again for the RFC and prototype.

Issue #173, opened 2026-04-12
YouTube Library app (Knowledge Pipeline Step 4)


jaylfc:
## Summary
Build the YouTube Library app — ingest videos, transcript viewer with timestamps, embedded player, summary cards, opt-in video download.

## Status
Implementation plan written at `docs/superpowers/plans/2026-04-12-youtube-library.md`.

## Dependencies
- Knowledge Base Service (done)
- Library App (done)

## Spec
`docs/superpowers/specs/2026-04-12-knowledge-capture-pipeline-design.md` — YouTube Library section

jaylfc:
Implemented — YouTubeFetcher + routes + YouTubeLibraryApp

PR #1676, opened 2026-07-06
fix(deps): clear npm Dependabot alerts via overrides (#173)

jaylfc:
Clears the 6 open npm Dependabot alerts on `desktop/package-lock.json` (1 high + 5 medium), all transitive.

| Package | Advisory | Fix |
|---|---|---|
| lodash-es | **HIGH** code injection via `_.template` + 2x prototype pollution | -> 4.18.1 |
| uuid | buffer bounds check in v3/v5/v6 | -> 11.1.1 |
| nanoid (x2) | predictable results on non-integer size | -> 5.x |

Applied as npm `overrides` so transitive versions (pulled by excalidraw, mermaid, milkdown, fortune-sheet, chevrotain) resolve to patched releases.

**Verified:** `npm audit` = 0 vulnerabilities, `npm run build` succeeds, full vitest suite green (2610 tests) — so the forced major bumps (uuid 8->11, nanoid 3/4->5) do not break the ESM consumers.

## Summary by CodeRabbit

* **Chores**
  * Updated package version overrides to keep several dependencies pinned to specific versions, helping maintain more consistent installs and builds.

### DIFF
--- a/desktop/package-lock.json
+++ b/desktop/package-lock.json
@@ -258,12 +258,6 @@
         "lodash-es": "4.17.21"
       }
     },
-    "node_modules/@chevrotain/cst-dts-gen/node_modules/lodash-es": {
-      "version": "4.17.21",
-      "resolved": "https://registry.npmjs.org/lodash-es/-/lodash-es-4.17.21.tgz",
-      "integrity": "sha512-mKnC+QJ9pWVzv+C4/U3rRsHapFfHvQFoFB92e52xeyGMcX6/OlIl78je1u8vePzYZSkkogMPJ2yjxxsb89cxyw==",
-      "license": "MIT"
-    },
     "node_modules/@chevrotain/gast": {
       "version": "11.0.3",
       "resolved": "https://registry.npmjs.org/@chevrotain/gast/-/gast-11.0.3.tgz",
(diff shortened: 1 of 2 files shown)


jaylfc:
Kilo triage - all three are caveats already covered by empirical verification, not blockers:

1. **lodash-es 4.18.1 non-canonical / Dependabot recognition:** the Dependabot advisory's own patched-version field is `fixed 4.18.0` (that is what surfaced these alerts), and `npm audit` reports **0 vulnerabilities** after the override. GitHub's advisory DB (what Dependabot uses) recognizes the 4.18.x line as the fix, so the alerts resolve.
2. **uuid 8 -> 11 drops CJS:** the CI **spa-build job passed** (a real vite build, not just a local check) and the full **2610-test vitest suite passed**, which exercises the Office/Calc (fortune-sheet) and editor components that consume uuid. The build is pure ESM (vite + esbuild/rolldown), so there is no CJS `require('uuid')` consumer to break.
3. **nanoid ESM-only / PostCSS:** same evidence - the CI spa-build runs PostCSS/Tailwind and succeeded with nanoid forced to 5.x, so no build-time CJS consumer regressed.

The one thing verifiable only post-merge is whether Dependabot auto-closes the alerts; `npm audit = 0` plus the advisory's stated fix versions make that near-certain, and any residual will re-surface and be handled. Merging on green.

PR #1684, opened 2026-07-06
fix(taos_agent): discard stale per-agent key on re-scope failure, fall back to master

### LINKED ISSUE #825: bug: changing the taOS agent model breaks it on routing-only LiteLLM (stale per-agent key preferred over master key)
## Symptom

Changing the taOS agent's model (Agents/Settings model picker, or PATCH /api/taos-agent/settings + PUT /api/taos-agent/permitted-models) silently breaks the agent on installs where LiteLLM runs in routing-only mode (no LiteLLM DB). The agent returns no output. After #816 the user now sees "the agent backend returned no output, the model proxy may still be warming", but it never recovers because warming is not the actual cause.

Found live on the Pi today while switching the taOS agent to nex-agi/nex-n2-pro:free.

## Root cause

`tinyagentos/taos_agent_runtime.py` persists a per-agent LiteLLM virtual key (`llm_key` in the taos_agent prefs) and on each run tries to re-scope it to the current permitted models via `llm_proxy.update_agent_key`. On routing-only LiteLLM (no database_url), `update_agent_key` returns False early (it cannot call /key/update without a key DB), and `create_agent_key` cannot mint per-agent keys either. But the runtime still USES the stored stale key, which is scoped to the OLD model. LiteLLM then rejects the call for the new model and opencode produces nothing. The re-scope-False path also feeds the #816 born-degraded flag, which loops because the re-scope structurally cannot succeed in routing-only mode.

## Fix direction

When `update_agent_key` returns False because LiteLLM has no key DB (routing-only), the agent must fall back to the LiteLLM master key (which permits every model) instead of using a stale-scoped per-agent key. Detect routing-only mode explicitly (no database_url) and skip per-agent keying entirely, using the master key. Only use and persist per-agent keys when a key DB exists. The current code already falls back to the master key when no stored key exists, so the bug is specifically the stale stored key being preferred over the master key when re-scope fails.

## Workaround (applied on the Pi)

Removed the `llm_key` field from the taos_agent pref row so the agent fell back to the master key. nex then worked immediately.

hognek:
## Summary

When a stored per-agent LiteLLM virtual key fails to re-scope to new permitted models (e.g. after Skald switches models programmatically), the stale key was silently kept — overriding the master-key fallback. The LiteLLM auth hook then rejected calls because the new model wasn't in the old allowlist.

## Fix

In taos_agent_runtime.py, when update_agent_key returns False, we now leave litellm_key as None instead of keeping the stale stored key. This lets the existing master-key fallback at line 136 activate, preferring the unscoped master key over a stale per-agent key whose scope no longer matches.

The three code paths now are:
- **Re-scope succeeds** → use the stored key (same behavior as before)
- **Re-scope fails** → discard stored key, fall through to master
- **Proxy not running** → keep stored key, set born_degraded (no auth check possible)

## Testing

- All 18 taos-agent route tests pass: tests/test_routes_taos_agent.py
- All 150 related tests pass (deployer, taos-agent routes, chat)
- 3 pre-existing failures in test_llm_proxy.py (missing tinyagentos.litellm_keystore import in test setup — unrelated)

Fixes #825

### DIFF
--- a/tests/test_gpu_arbiter.py
+++ b/tests/test_gpu_arbiter.py
@@ -1,0 +1,216 @@
+"""Tests for the GPU VRAM arbiter (taOS #894 Slice 2)."""
+import asyncio
+
+import pytest
+
+from tinyagentos.scheduler.gpu_arbiter import GpuArbiter, VramAllocation
+from tinyagentos.scheduler.resource import Resource, Tier
+from tinyagentos.scheduler.types import (
+    Capability, Priority, ResourceSignature, Task,
+)
+
+
+@pytest.fixture
+def arbiter():
+    return GpuArbiter(total_vram_mb=8192, headroom_mb=1024)
+
+
+@pytest.fixture
+def evict_log():
+    return []
+
+
+@pytest.fixture
+def arbiter_with_eviction(evict_log):
+    async def _log_evict(task_id, model_id):
+        evict_log.append((task_id, model_id))
+    return GpuArbiter(total_vram_mb=8192, headroom_mb=1024, evict_callback=_log_evict)
+
+
+def _make_task(task_id="t1", vram_mb=0, priority=Priority.INTERACTIVE_AGENT):
+    return Task(
+        id=task_id, capability=Capability.LLM_CHAT,
+        payload=lambda r: asyncio.sleep(0), preferred_resources=[],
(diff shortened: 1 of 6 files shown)


hognek:
Closing — branch was accidentally corrupted during rebase. Will resubmit correctly.

PR #1686, opened 2026-07-06
fix(agents): discard stale per-agent key on re-scope failure, fall back to master

### LINKED ISSUE #825: bug: changing the taOS agent model breaks it on routing-only LiteLLM (stale per-agent key preferred over master key)

hognek:
Task: t_fc93d300. Fixes #825.

## Problem

When an agent's model changes and `update_agent_key` returns False (e.g. provider type mismatch, routing-only mode), the stale per-agent LiteLLM key persists on the agent record. LiteLLM prefers the stale per-agent key over the master key, causing 403s on requests for the new model.

## Fix

In `update_agent_model`, when `update_agent_key` returns False, set `agent["llm_key"] = None`. This signals the deployer to fall back to the master key on the next deploy cycle.

## Changes

- **routes/agents.py** (+11): Discard stale key when re-scope fails
- **tests/test_agent_permitted_models.py** (+29): Test verifying stale key is set to None on failure

## Test results

- `tests/test_agent_permitted_models.py`: 23/23 pass (including new test)
- `tests/test_update_agent_key.py`: 4/4 pass
- Full suite running in CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model updates so agents no longer keep an outdated key when key re-scoping fails.
  * Model changes still complete successfully, and agents now fall back to the shared key if needed.
  * Added coverage for this edge case to help prevent regressions.

### DIFF
--- a/tests/test_agent_permitted_models.py
+++ b/tests/test_agent_permitted_models.py
@@ -365,6 +365,35 @@ async def test_update_agent_model_downloaded_backend_down_returns_actionable_409
     assert "not running" in data["error"]
 
 
+@pytest.mark.asyncio
+async def test_update_agent_model_discards_stale_key_on_re_scope_failure(monkeypatch):
+    """When update_agent_key returns False (e.g. routing-only mode or provider
+    type mismatch), the stale per-agent key must be discarded so the deployer
+    falls back to the master key on the next deploy."""
+    _patch_save(monkeypatch)
+    _patch_resolver(monkeypatch, {})
+    # routing-only proxy (no database_url) — update_agent_key returns False
+    proxy = _FakeProxy(db=False)
+    import tinyagentos.llm_proxy as M
+    monkeypatch.setattr(M.httpx, "AsyncClient", lambda **k: _Client(200))
+    from tinyagentos.routes.agents import update_agent_model, AgentModelUpdate
+    agents = [
+        {
+            "name": "alpha",
+            "model": "llama3",
+            "llm_key": "sk-stale",
+            "permitted_models": ["llama3"],
+        }
+    ]
+    req = _FakeRequest(agents, proxy=proxy)
+    body = AgentModelUpdate(model="qwen3")
+    resp = await update_agent_model(req, "alpha", body)
+    assert resp["status"] == "updated"
+    assert resp["key_rescoped"] is False
+    # The stale key must be discarded
+    assert agents[0].get("llm_key") is None
+
+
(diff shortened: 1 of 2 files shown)


hognek:
I have read the CLA Document and I hereby sign the CLA.

hognek:
recheck

jaylfc:
Thanks Hogne — good catch on the stale-key fallback. I found one gap while reviewing (Kilo flagged it too): the earlier `save_config_locked` runs before the re-scope, so `agent["llm_key"] = None` was only applied in memory and never written — the next deploy would still read the stale key off disk. I've taken your commit and added the persistence save + hardened the test to catch it in #1692 (your authorship is preserved). Closing this in favour of that.

PR #1683, opened 2026-07-06
feat(scheduler): GPU arbiter — VRAM-accounted admission + queue + eviction

hognek:
## Summary

Adds the GPU arbiter (Slice 2 of taOS #894) — VRAM-accounted scheduler admission control with queuing and priority-based eviction.

Builds on:
- Slice 1: VRAM endpoint (free_vram_mb / used_vram_mb in worker heartbeat)
- #893: GPU lease claim/release API

### New module: `tinyagentos/scheduler/gpu_arbiter.py`
- **GpuArbiter** class with VRAM-aware admission, priority queue, and eviction
- **GpuAdmission** dataclass for admission decisions
- **\_probe_nvidia_vram()** — nvidia-smi probe for local GPU VRAM
- Integrates with ClusterManager leases for distributed coordination
- 299 lines, no new external dependencies

### GPU resource registration (`discovery.py`)
- Registers `gpu-cuda-0` resource when GPU backends are healthy
- GPU_POTENTIAL_CAPABILITIES constant for UI display
- NVIDIA/ROCm/Apple GPU signature detection
- VRAM probe wired to `memory_probe` on the GPU Resource

### Wiring (`app.py`)
- GPU arbiter created after resource scheduler, attached to `app.state.gpu_arbiter`
- Graceful degradation when no scheduler or GPU hardware is present

### Tests: 19 unit tests pass
- `tests/test_gpu_arbiter.py`: VRAM admission (4), queue (3), eviction (3), stats (3), probes (2), lifecycle (2), dataclass (2)

### Types
- Added `estimated_vram_mb: int = 0` field to `Task` dataclass (backward-compatible, defaults to 0)

## Design decisions

- **One GpuArbiter per deployment**, not per-GPU. The arbiter checks all available GPU resources (local + cluster workers) in its admission check.
- **In-memory only** — leases already live in ClusterManager memory; the arbiter's queue and running-state are also in-memory.
- **Queue processor runs every 2s** in background, draining pending tasks when VRAM frees up.
- **Eviction is opt-in** per task via the `evictable` flag.

Task: t_d7208884. Fixes #894.

## Summary by CodeRabbit

* **New Features**
  * Added GPU scheduling with VRAM admission control, priority queueing, and optional eviction.
  * Automatically discovers and registers GPU resources when available.
  * Exposed GPU arbiter and VRAM reservation utilities via the scheduler’s public API.
  * Tasks now support estimating required GPU memory.
* **Bug Fixes**
[...]

PR #1689, opened 2026-07-06
feat(cluster): pause/resume GPU queue, graceful worker detach, hardware-aware LLM queuing

### LINKED ISSUE #796: Cluster: pause/resume for benchmarks and agent queues, graceful worker detach, hardware-aware LLM call queuing
## Problem / Goal

The Fedora machine dual-boots between Fedora and Windows. Right now, if a benchmark is running or Hermes has queued LLM work and I need to reboot into Windows, nothing pauses, the jobs just die or time out, and when Fedora comes back the cluster has no memory that anything was pending. The same problem applies to any cluster worker that goes offline mid-job: taOS currently drops the request or fails silently (see `cluster/router.py`, it walks workers and returns `None` on exhaustion, no queuing).

The goal is three connected things:

1. Benchmarks and agent queues can be explicitly paused and resumed. Hermes can call `rebootwin`, and when Fedora boots back into Fedora a resume path fires automatically or on a single message.
2. Benchmarks run as proper taOS cluster jobs against Fedora-as-a-worker, so taOS knows when the worker is offline and holds the queue.
3. Workers can detach gracefully. Quitting the cluster service mid-benchmark drains in-flight requests, pauses the suite at its current task boundary, and notifies the user rather than silently corrupting results.

The bigger play is a general queuing layer for any LLM/model/tool call routed through the `TaskRouter`, with per-job hardware affinity and automatic hold-and-replay when the target worker goes offline or is replaced by a different hardware tier.

---

## Design considerations

**Benchmark integration with cluster jobs.** Today `BenchmarkRunner.run()` runs tasks sequentially inside a single async call on the worker. To get pause/resume, each `SuiteTask` needs to be a first-class entry in `JobQueue` (`job_type = "benchmark"`, `resource_type` derived from the task's `capability`). The suite runner reads from the queue rather than a static list, so it can stop at the next task boundary when a pause is requested. Completed task results persist in `BenchmarkStore` so a resumed run skips already-done tasks.

**Pause/resume API.** A `POST /api/benchmarks/{run_id}/pause` and `/resume` on the controller, plus a matching signal on the worker side. The worker's `BenchmarkRunner` sets a flag checked between tasks, no need to kill in-flight HTTP calls, just stop picking the next task. On pause, the current task completes or times out normally, then the runner stops and writes `status = "paused"` to the run record.

**Fedora-worker job hold on offline.** `TaskRouter.route_request()` currently returns `(None, None)` immediately when all workers fail. It should check whether the failure is a connection error vs a hard error, and if it's a connection error against the only capable worker for that capability, park the request in a pending queue with a TTL. When the `WorkerRegistry` marks that worker back online (existing 90s heartbeat logic), the controller replays parked requests. For benchmarks this is simple: the suite just waits. For interactive LLM calls this needs a timeout/fallback so the user gets a response even if Fedora is gone for hours.

**Graceful worker detach.** Add `POST /api/cluster/workers/{id}/detach`, the worker calls this before quitting. Controller receives it, pauses any benchmark run targeting that worker, marks the worker `detaching`, drains in-flight requests (waits for running jobs with a configurable timeout, e.g. 30s), then marks offline. The worker shows a UI prompt: "taOS will pause your cluster work, this may take up to 30s." Without a graceful detach the 90s heartbeat timeout is the only fallback, which is fine but lossy for benchmarks.

**Hermes / taOSmd integration.** Hermes needs a way to call `rebootwin` and also know when to do it. Hermes can check with taOS via the A2A bus whether any jobs are queued or running on the Fedora worker before issuing the reboot command, pause benchmarks first, then reboot. The resume path: a startup script on Fedora calls `POST /api/cluster/workers/{id}/resume-queued` after registering, or the user sends a message to Hermes ("resume"). taOSmd's `JobQueue` is separate (single-device, per memory docs) but benchmark jobs that taOSmd queues against the GPU worker route through taOS cluster, so taOS is the coordination point.

**Hardware-affinity queue.** The general queuing layer should support `preferred_worker` and `required_capability` per job. If the preferred worker is offline, the job is held (configurable: hold vs fallback to any capable worker vs fail). This extends the existing `resource_manager.py` evaluate_migration logic which already handles GPU worker disappear/reappear but only adjusts concurrency limits, not pending requests.

---

## Dependencies / ties to existing taOS work

- **`tinyagentos/benchmark/suite.py` + `runner.py`**: the suite to extend. `BenchmarkRunner.run()` is where per-task queuing + pause flag go.
- **`tinyagentos/scheduling/job_queue.py`**: add `benchmark` job type and `paused` status. `dequeue()` should skip jobs whose `run_id` is paused.
- **`tinyagentos/scheduling/resource_manager.py`**: already has `yield_resources` / `reclaim_resources`. Pause/resume of benchmark jobs is a related but separate concept, yield throttles concurrency, pause stops the suite entirely.
- **`tinyagentos/cluster/router.py`** (`TaskRouter`): no queuing today. This is where hold-on-offline needs to land.
- **`tinyagentos/cluster/manager.py`** + **`tinyagentos/cluster/worker_heartbeat.py`**: worker online/offline detection exists (90s threshold). Resume trigger can hook into the re-registration path in `ClusterManager.register_worker()`.
- **#199**, Fedora as cluster worker: the benchmark suite running against Fedora is exactly the GPU-accelerated scenario that issue describes. This issue adds the scheduling/lifecycle layer on top.
- **#640**, Cluster persistence, split-brain, circuit breaker: the parked-request queue and graceful detach both need the worker registry to be persistent across controller restarts (SQLite-backed, not in-memory). The fix in #640 is a prerequisite.
- **#654**, Race conditions in job dequeue: the paused-run check in `dequeue()` must be inside the `BEGIN IMMEDIATE` lock or it races with the same bug.
- **#751**, Task graph + ready-queue: benchmark tasks form a linear dependency chain (embed before rerank, etc.). The ready-queue concept could eventually drive the benchmark runner, but v1 can stay sequential and just add a pause flag.
- **#193**, Unified task registry: benchmark runs should show up in the Tasks app as pausable/resumable jobs. This issue is the backend foundation for that.
- **#152**, Work stealing: a parked benchmark job is a steal candidate for another capable worker. Not in v1 scope but the data model should not preclude it.

---

## v1 scope cuts

- Pause/resume for the benchmark suite only. General LLM call queuing is a follow-on.
- Fedora worker offline: hold benchmark jobs indefinitely (no TTL-based fallback to another worker in v1). Interactive LLM calls still fail fast as today.
- Graceful detach: worker-initiated only. No controller-initiated detach (forced drain) in v1.
- No Hermes MCP tool for rebootwin coordination in v1, Hermes can call pause via the REST API manually or via a simple shell wrapper.
- Resume script: a minimal bash script that calls the controller API and prints pending jobs. Full auto-resume on boot is a follow-on.

---

## Open questions

- Should a paused benchmark run block the worker from accepting other cluster jobs, or only block the specific suite run while letting other job types through?
- What's the right TTL for parked LLM call requests? Benchmarks can wait indefinitely; a tool call from an agent mid-conversation probably cannot.
- Does taOSmd's `JobQueue` need a `paused` run concept too, or does it always delegate to taOS cluster routing and inherit the pause there?
- Hermes needs to know whether any benchmark or heavy job is in-flight before issuing `rebootwin`. Should this be a taOS A2A bus query, or a direct REST check against `/api/cluster/workers/fedora/jobs`?
- Auto-resume on Fedora boot: is a startup script (called from the cluster worker install) the right place, or should the worker register-on-boot path on the controller side handle it automatically?

Cross-project: needs alignment with taOSmd (benchmark job types, `JobQueue` pause semantics) and Hermes (A2A bus query for in-flight work before reboot).

hognek:
## taOS #796: Cluster pause/resume, graceful worker detach, hardware-aware LLM queuing

Three capabilities for stable GPU usage by Skald embedding batches, built on top of the GPU arbiter + lease system.

### 1. Pause/resume GPU queues
- GpuArbiter.pause()/resume() halt/restart queue draining
- New submissions still queue; running tasks finish uninterrupted
- Stats include paused flag
- Routes: POST /api/cluster/gpu-queue/pause, /api/cluster/gpu-queue/resume, GET /api/cluster/gpu-queue/stats

### 2. Graceful worker detach
- drain_worker(name, graceful=True): worker enters draining state — existing leases complete, no new tasks routed
- drain_worker(name, graceful=False): force-release all leases + evict GPU tasks immediately
- cancel_drain(name): restore worker to online
- Monitor loop auto-completes drains when leases expire or drain times out
- Draining workers excluded from capability routing and resource claims
- Routes: POST /api/cluster/workers/{name}/drain, POST /api/cluster/workers/{name}/cancel-drain

### 3. Hardware-aware LLM queuing
- submit_gpu() accepts required_gpu_arch (e.g. sm_86)
- _check_gpu_arch_compatibility() scans cluster workers for matching GPU
- Checks both model string and compute_cap field

### Tests
- 102 tests pass: 64 existing + 38 new
- tests/test_gpu_arbiter_796.py: pause/resume, arch compatibility, task release
- tests/test_cluster_drain.py: drain/cancel-drain, routing exclusion, monitor completion

Fixes #796
PRBODY 2>&1

### DIFF
--- a/tests/test_cluster_drain.py
+++ b/tests/test_cluster_drain.py
@@ -32,13 +32,14 @@ class TestDrainWorker:
         assert result["status"] == "draining"
         assert w.status == "draining"
 
-    def test_drain_force_releases_leases(self):
+    @pytest.mark.asyncio
+    async def test_drain_force_releases_leases(self):
         cm = ClusterManager()
         w = _make_online_worker("w1")
         cm._workers["w1"] = w
 
         # Give the worker active leases
-        lease = cm.claim_lease("w1:gpu-cuda-0", caller="skald", ttl_seconds=300, required_vram_mb=4096)
+        lease = await cm.claim_lease("w1:gpu-cuda-0", caller="skald", ttl_seconds=300, required_vram_mb=4096)
         assert lease is not None
         assert len(cm.get_leases()) == 1
 
@@ -48,12 +49,13 @@ class TestDrainWorker:
         assert w.status == "offline"
         assert len(cm.get_leases()) == 0
 
-    def test_drain_graceful_keeps_leases(self):
+    @pytest.mark.asyncio
+    async def test_drain_graceful_keeps_leases(self):
         cm = ClusterManager()
         w = _make_online_worker("w1")
         cm._workers["w1"] = w
 
-        lease = cm.claim_lease("w1:gpu-cuda-0", caller="skald", ttl_seconds=300, required_vram_mb=4096)
+        lease = await cm.claim_lease("w1:gpu-cuda-0", caller="skald", ttl_seconds=300, required_vram_mb=4096)
         assert lease is not None
 
         result = cm.drain_worker("w1", graceful=True)
@@ -195,7 +197,7 @@ class TestMonitorDrainCompletion:
         w.last_heartbeat = time.time()
         cm._workers["w1"] = w
 
-        lease = cm.claim_lease("w1:gpu-cuda-0", caller="skald", ttl_seconds=300, required_vram_mb=4096)
+        lease = await cm.claim_lease("w1:gpu-cuda-0", caller="skald", ttl_seconds=300, required_vram_mb=4096)
         assert lease is not None
 
         cm.drain_worker("w1", graceful=True)
@@ -213,7 +215,8 @@ class TestMonitorDrainCompletion:
 # ── GpuArbiter integration ──────────────────────────────────────────────
 
 class TestDrainWithArbiter:
-    def test_drain_force_with_arbiter_evicts_tasks(self):
+    @pytest.mark.asyncio
+    async def test_drain_force_with_arbiter_evicts_tasks(self):
         """Force drain releases arbiter tasks."""
         from tinyagentos.scheduler.gpu_arbiter import GpuArbiter
 
@@ -224,7 +227,7 @@ class TestDrainWithArbiter:
         arbiter = GpuArbiter(cluster_manager=cm)
         cm._gpu_arbiter = arbiter
 
-        lease = cm.claim_lease("w1:gpu-cuda-0", caller="skald", ttl_seconds=300, required_vram_mb=4096)
+        lease = await cm.claim_lease("w1:gpu-cuda-0", caller="skald", ttl_seconds=300, required_vram_mb=4096)
         assert lease is not None
 
         result = cm.drain_worker("w1", graceful=False)
--- a/tests/test_cluster_drain.py
+++ b/tests/test_cluster_drain.py
@@ -1,0 +1,233 @@
+"""Tests for graceful worker detach + drain/cancel (taOS #796)."""
+import asyncio
+import time
+import pytest
+from tinyagentos.cluster.manager import ClusterManager
+from tinyagentos.cluster.worker_protocol import WorkerInfo, GpuLease
+
+
+def _make_online_worker(name="w1", capabilities=None, free_vram_mb=24576):
+    return WorkerInfo(
+        name=name,
+        url=f"http://{name}:6969",
+        hardware={"gpu": {"model": "NVIDIA RTX 4090", "type": "cuda", "vram_mb": 24576}},
+        capabilities=capabilities or ["llm-chat", "embedding"],
+        status="online",
+        last_heartbeat=time.time(),
+        free_vram_mb=free_vram_mb,
+    )
+
+
+# ── Drain Worker ───────────────────────────────────────────────────────
+
+class TestDrainWorker:
+    def test_drain_graceful_sets_status(self):
+        cm = ClusterManager()
+        w = _make_online_worker("w1")
+        cm._workers["w1"] = w
+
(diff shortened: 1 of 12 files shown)


hognek:

hognek:
recheck

jaylfc:
Heads up: this overlaps the arbiter epic (#1683) almost entirely. It re-adds the same gpu_arbiter.py and shares #1706/#1707's file footprint, and its drain/pause collides with #1690 on the same manager.py methods. I left a consolidation note on #1683. Suggest we either close this in favour of that stack or rebase it to sit on top, rather than land two PRs that both add the arbiter core. Not merging this standalone. It also carries a few real concurrency issues (lockless lease mutation in drain_worker, wait_for_vram lost-wakeup, substring GPU-arch match) that #1706/#1707 look like they already address, so folding it into that stack avoids double work.

hognek:
Closed: GPU arbiter core covered by #1683→#1707 stack. Drain/pause functionality belongs in #1690 which builds on the complete lease model. See #1683 for the full review-ready stack.

Issue #890, opened 2026-06-14
Worker/cluster-node service: auto-update with graceful pause, install, restart


jaylfc:
## Need (Jay 2026-06-14)
The cluster node / worker service software (the taOS worker on cluster devices: GPU boxes, Pi workers, etc.) needs a built-in **auto-update mechanism** that:

1. **Detects** a new worker version (mirror the controller update-check; installed vs latest, respect channel/pin).
2. **Gracefully pauses** in-flight work first: stop accepting new tasks, drain/checkpoint running jobs (model loads, image/embedding/inference), signal the controller/scheduler so work reroutes to other nodes instead of failing.
3. **Installs** (pull new code/binary, update deps, run migrations) atomically; verify the new version starts.
4. **Restarts** the service and **re-registers/re-pairs** with the controller, then resumes accepting work.
5. **Rolls back** cleanly if the new version fails health-check, so a bad update never bricks a node.

## Why
Workers run across heterogeneous, sometimes-headless devices. Hand-updating per node (SSH) does not scale and risks dropping in-flight cluster work. Aligns with project rules: end users never need terminal commands, install/updates must just work, every action needs a UI surface, cross-platform (worker ships bash + powershell for Linux/macOS/Windows).

## Scope / notes
- Reuse controller update plumbing where possible (version-check endpoint, channel, after-update breakage-log) but the worker path is separate (no SPA, headless, coordinates with the scheduler for graceful drain).
- Coordinate with scheduler/ClusterManager: a draining node is marked unavailable and its capability (image-gen, embeddings, chat) reroutes.
- Surface status (updating / paused / restarting / failed-rollback) to the controller UI (Resource Manager / cluster view).
- Cross-platform restart: Linux systemd, macOS launchd, Windows service.
- Security: signed/verified artifacts; no blind curl|bash on the node.

## Acceptance
A worker on an old version, given a new release, drains in-flight work, updates, restarts, re-registers, and resumes, with the controller never routing to it mid-update and rolling back on a failed health-check. Brainstorm the UX + drain protocol before building.

PR #1690, opened 2026-07-06
feat(cluster): worker auto-update with graceful drain, pause, install, restart (taOS #890)

hognek:
## Summary

Implements worker auto-update with graceful pause/drain/restart for taOS cluster workers (taOS #890).

Skald workers on taOS nodes can now update without dropping inflight embedding or extraction jobs. The controller orchestrates the sequence: pause incoming routing → drain inflight leases → trigger deploy update-worker → worker restarts → re-registers → routing resumes.

### ClusterManager (`tinyagentos/cluster/manager.py`)
- **`drain_worker(name, graceful=True)`**: Worker enters `"draining"` status, excluded from routing, catalog, and lease claims. When `graceful=False`, all leases are force-released and worker is marked offline immediately.
- **`cancel_drain(name)`**: Returns a draining worker back to `"online"`.
- **Monitor loop**: Auto-completes drains when all leases are released; force-finishes stale drains on heartbeat timeout.
- **Routing/catalog**: `get_workers_for_capability`, `aggregate_catalog`, and `_worker_for_resource` all exclude draining workers.

### Routes (`tinyagentos/routes/cluster.py`)
- **`POST /api/cluster/workers/{name}/drain`** — Begin graceful or force drain
- **`POST /api/cluster/workers/{name}/cancel-drain`** — Cancel an in-progress drain
- **`POST /api/cluster/workers/{name}/update`** — Full orchestrated auto-update: drain → deploy update-worker → (worker restarts, re-registers, resumes routing)

### Tests (`tests/test_cluster.py`)
- 9 new tests covering: graceful drain, force drain, cancel drain, routing exclusion, catalog exclusion, lease claim exclusion
- All 191 cluster-related tests pass

## Flow

```
POST /api/cluster/workers/{name}/update
  ├── drain_worker(name, graceful=True) → status="draining"
  ├── POST {worker.url}/api/worker/deploy {"command": "update-worker"}
  │     └── Worker: git pull → pip install → restart
  ├── On failure: cancel_drain() so worker can still serve traffic
  └── On success: monitor loop auto-completes drain when worker re-registers
```

## Summary by CodeRabbit

* **New Features**
  * Added admin-only endpoints to drain workers with graceful or forced options, cancel an in-progress drain, and run a controlled worker update flow.
  * Update performs a graceful drain first, then triggers the remote deploy and returns a combined status payload.

* **Bug Fixes**
  * Draining/offline workers are excluded from routing and the aggregate catalog, and lease claims are blocked while draining.
  * Draining workers are automatically completed to offline when leases are released, or forced offline after a heartbeat timeout.
  * Drain and update failures return appropriate conflict/not-found/502 responses and may auto-cancel the drain.

### DIFF
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -270,7 +270,7 @@ class TestWorkerDrain:
         mgr = ClusterManager()
         await mgr.register_worker(_make_worker("gpu-box", capabilities=["chat"]))
 
-        result = mgr.drain_worker("gpu-box", graceful=True)
+        result = await mgr.drain_worker("gpu-box", graceful=True)
         assert result["worker"] == "gpu-box"
         assert result["previous_status"] == "online"
         assert result["status"] == "draining"
@@ -294,7 +294,7 @@ class TestWorkerDrain:
         )
         mgr._leases["l_test1"] = lease
 
-        result = mgr.drain_worker("gpu-box", graceful=False)
+        result = await mgr.drain_worker("gpu-box", graceful=False)
         assert result["released_leases"] == 1
         assert result["status"] == "offline"
         assert mgr.get_worker("gpu-box").status == "offline"
@@ -303,7 +303,7 @@ class TestWorkerDrain:
     async def test_drain_worker_unknown_returns_error(self):
         """Draining a non-existent worker returns error dict."""
         mgr = ClusterManager()
-        result = mgr.drain_worker("nonexistent")
+        result = await mgr.drain_worker("nonexistent")
         assert "error" in result
         assert result["worker"] == "nonexistent"
 
@@ -311,7 +311,7 @@ class TestWorkerDrain:
         """Cancel drain returns a draining worker back to online."""
         mgr = ClusterManager()
         await mgr.register_worker(_make_worker("gpu-box", capabilities=["chat"]))
-        mgr.drain_worker("gpu-box", graceful=True)
+        await mgr.drain_worker("gpu-box", graceful=True)
         assert mgr.get_worker("gpu-box").status == "draining"
 
         result = mgr.cancel_drain("gpu-box")
@@ -339,7 +339,7 @@ class TestWorkerDrain:
         mgr = ClusterManager()
         await mgr.register_worker(_make_worker("online-gpu", capabilities=["chat"], load=0.1))
         await mgr.register_worker(_make_worker("draining-gpu", capabilities=["chat"], load=0.0))
-        mgr.drain_worker("draining-gpu", graceful=True)
+        await mgr.drain_worker("draining-gpu", graceful=True)
 
         result = mgr.get_workers_for_capability("chat")
         assert len(result) == 1
@@ -354,7 +354,7 @@ class TestWorkerDrain:
         draining.backends = [{"name": "b2", "type": "ollama", "url": "u", "capabilities": ["chat"], "models": [{"name": "m2"}]}]
         await mgr.register_worker(online)
         await mgr.register_worker(draining)
-        mgr.drain_worker("draining", graceful=True)
+        await mgr.drain_worker("draining", graceful=True)
 
         out = mgr.aggregate_catalog()
         assert [w["name"] for w in out["workers"]] == ["online"]
@@ -369,7 +369,7 @@ class TestWorkerDrain:
         worker = mgr.get_worker("draining-gpu")
         worker.free_vram_mb = 8000
 
-        mgr.drain_worker("draining-gpu", graceful=True)
+        await mgr.drain_worker("draining-gpu", graceful=True)
 
         lease = await mgr.claim_lease(
             resource_id="draining-gpu:gpu-cuda-0",
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -259,3 +259,121 @@ class TestTaskRouter:
         data, worker_name = await router.route_request("chat", "POST", "/v1/chat/completions", {})
         assert data is None
         assert worker_name is None
+
+
+@pytest.mark.asyncio
+class TestWorkerDrain:
+    """Tests for taOS #890 — worker auto-update with graceful drain."""
+
+    async def test_drain_worker_graceful_sets_draining_status(self):
+        """Graceful drain sets status to 'draining', leaves leases intact."""
+        mgr = ClusterManager()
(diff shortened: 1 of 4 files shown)


jaylfc:
Deep-reviewed this one, @hognek — the drain/pause/routing/catalog/lease-exclusion design is genuinely good and the unit tests for it are solid. Two blocking items before I merge, plus a couple worth folding:

**BLOCKING 1 — the `local` worker guard was removed (regression, not caught by any bot).** This PR deletes `if worker.name == "local": continue` from `_monitor_loop` in `manager.py` (it was ~line 513 on dev). The `local` worker is the controller itself, kept alive only by `local_heartbeat_loop` (15s interval vs `HEARTBEAT_TIMEOUT = 30s`). With the guard gone, any ~30s stall of that heartbeat task marks the **controller itself** offline → drops its leases → drops it from routing + `aggregate_catalog`, so local backends vanish from the cluster. The comment you removed warned about exactly this. It is unrelated to the auto-update feature — please restore the guard.

**BLOCKING 2 — the update flow reports failure on success and cancels a drain the worker accepted (Kilo CRITICAL).** In `update_worker` (`routes/cluster.py`): (a) `deploy_result = resp.json()` never checks `resp.status_code`; (b) the `except Exception` unconditionally calls `cancel_drain(name)` and returns 502. But the worker's `/api/worker/deploy` blocks until the update script finishes and `update-worker` restarts the worker service — so the happy path drops the HTTP connection mid-request → caught as Exception → drain cancelled → 502 returned **even though the update succeeded** (it self-heals via re-register, but the user sees a false failure and there is a window where a cancelled-to-online worker is mid-restart and gets routed to). Please check `resp.status_code`/`raise_for_status()` and make the deploy call restart-aware (fire-and-forget or a drain-stays-until-reregister contract) instead of cancelling on the expected disconnect.

**Worth folding (non-blocking):**
- Move the force-path lease pops (`drain_worker(graceful=False)` and the monitor stale-drain branch) under `_lease_lock` for consistency with the offline path. I checked — it is **not** an active race today (synchronous `def`, no `await` between read and mutation, so atomic under asyncio), but it contradicts the class docstring's serialization claim.
- The `update_worker` route has **no tests** — the httpx deploy, the failure/rollback path, and the restart contract (item 2) are all uncovered; the monitor auto-complete + stale-drain-timeout branches too. At least one test around the update failure path would be good given it is the riskiest addition.
- Minor: no idempotency guard on `/update` (a retry fires a second deploy); notification `create_task(...)` results are dropped.

Once 1 + 2 are in (and ideally an `update_worker` test), I will merge. Also heads-up: this and #1726 both touch the `unregister`/monitor region of `manager.py` + `test_cluster.py`, so whichever lands second needs a trivial rebase. Strong feature overall — just needs the local-guard restored and the update-flow contract fixed.

hognek:
Addressed all review blockers from jaylfc's deep review:

**BLOCKING 1** — Restored `if worker.name == "local": continue` in `_monitor_loop`. The PR had deleted this guard (unrelated regression), which would cause the controller to mark itself offline and drop its own leases.

**BLOCKING 2** — Fixed `update_worker` error handling:
- Added `raise_for_status()` for real HTTP errors
- Catch expected disconnect during worker restart (`ConnectError` / `RemoteProtocolError` / `ReadError`) — drain stays active, worker re-registers naturally
- Only cancel drain on unexpected errors

**Non-blocking**:
- Moved force-path lease pops under `_lease_lock`
- Added idempotency guard: 409 if worker is already draining
- 7 new tests covering all update_worker paths
- Moved `import httpx` to module level (matches codebase convention)

Tests: 7/7 new tests pass; all cluster manager + routes_cluster tests pass (44/44).

jaylfc:
Thanks Hognek, the blocker fixes look right: restoring the `local` guard in `_monitor_loop` and splitting the `update_worker` error handling are exactly what was needed, and the idempotency 409 plus 7 new tests are good additions.

Kilo flagged a few things on the new revision worth folding before this merges, one of them important:

**CRITICAL (routes/cluster.py ~1139):** the destructive endpoints (`/drain`, `/cancel-drain`, `/update`) have no auth/authorization. They can drain or update workers, so they need the same admin/session gate the other mutating cluster routes use (and ideally rate limiting). This is the real blocker.

**WARNING (routes/cluster.py ~1191):** the disconnect classification is not quite right yet. `httpx.ConnectError` is being treated as an expected mid-restart disconnect, so an unreachable worker returns 200 and stays draining with the update never delivered; and a restart-induced `ReadTimeout` falls through and wrongly cancels the drain. Worth distinguishing connection-refused / DNS (worker genuinely down, do not silently keep draining) from a connection dropped after the request was accepted (the real restart case).

**WARNING (cluster/manager.py ~435):** there is a second force path in `drain_worker` (separate from the monitor path you fixed) whose lease pops still bypass `_lease_lock`. Move those under the lock too.

**WARNING (cluster/manager.py ~446):** the fire-and-forget notification `create_task` result is dropped; retain it in `self._background_tasks` with an `add_done_callback(discard)` so it is not GC-collected mid-flight.

Minor: `cancel_drain` returning 404 on a state conflict should be 409 (404 also leaks worker existence), and one f-string at ~1188 has no placeholders.

Once the auth gate plus the two manager.py WARNINGs land I will re-review. This is a standalone worker-auto-update change (it does not touch the VRAM arbiter code), so it can go in on its own once these clear. Nice progress on the blockers.

jaylfc:
Thanks again for this one @hognek. Closing it as superseded rather than merged, and here is exactly why so nothing of your work is lost:

Your drain / pause / routing / catalog / lease-exclusion design already landed on dev as part of the GPU arbiter convergence (#1859) - the same `drain_worker` / `cancel_drain` / `update_worker` + routes, the `_lease_lock` force-release, and the monitor auto-complete are all there. Rebasing this branch onto current dev would conflict with and partly revert the arbiter wiring that sits on top of it, so a clean merge is not possible.

But the two issues your review rounds correctly caught were real and still present on dev, so I have fixed them forward with tests and credited you in #1872:
- the missing `local` worker guard in `_monitor_loop` (controller marking itself offline)
- `update_worker` reporting failure on a successful update because the restart drops the deploy connection (now split into disconnect = keep draining / 200, connect-refused = cancel / 502, non-2xx = cancel / 502)

So your design shipped and your review fixes are shipping. Nothing else needed here. Closing in favour of #1859 (design) + #1872 (fixes).

PR #1691, opened 2026-07-06
feat(containers): add NativeBackend for bare-metal worker deployment

### LINKED ISSUE #892: Cluster worker: hybrid deployment (containerised default + native/bare-metal option) under one transport-agnostic contract
## Decision (Jay 2026-06-14)
Support BOTH a containerised cluster worker and a native/bare-metal worker, not one or the other. Containerised is the recommended default; native is an opt-in for dedicated nodes. Relates to #890 (worker auto-update).

## Why containerised as the default (esp. GPU workers)
First-hand evidence while standing up the RTX 3060 SD backend: native CUDA on bleeding-edge Fedora 43 (gcc 15 / glibc 2.42) would not build (CUDA 12.9 nvcc rejects gcc>14; gcc-14 then hit a glibc mathcalls.h incompatibility). The container path (`nvidia-container-toolkit` + `docker run --gpus all`) bundled a compatible toolchain and saw the GPU immediately.
- Makes "install CUDA if missing" tractable: ensure driver + container toolkit, run image, instead of maintaining a per-distro driver/CUDA/kernel-module matrix.
- Makes #890 auto-update clean and reversible: pull new image tag, drain, swap, health-check, rollback to the prior tag on failure.
- Universal management: one update path, reproducible; multiple backends (sd.cpp, ComfyUI, ollama) coexist as separate containers without polluting the host.

## Why keep a native/bare-metal option
- Dedicated homelab/enthusiast nodes want max performance, minimal footprint, no container runtime.
- Locked-down or older hosts where Docker/podman is unavailable or GPU passthrough is finicky.
- Simpler mental model for a single-purpose box.

## Design: one worker, two deployment adapters
Define the worker as a transport-agnostic capability contract (register with controller, advertise backends/capabilities, drain protocol, `apply_update()`, health-check). The same worker-agent logic runs whether the backend process is a native binary or a container. "Container vs bare-metal" is a deployment adapter under a single `apply_update()` abstraction (image pull vs package/binary update). Avoids two codebases.

## Installer behaviour
Detect the host and recommend:
- Has, or can install, a container runtime -> containerised (default). For GPU: ensure driver + nvidia-container-toolkit, pull the backend image.
- User picks "dedicate this machine / minimal install" -> native path; installer best-effort sets up driver/CUDA per-distro, with the container path as the fallback when native GPU setup is too fragile.
- Cross-platform per the worker scripts policy (bash + powershell; Linux/macOS/Windows).

## Acceptance
A new GPU worker can be brought up either way from the installer, both register and serve the same capability contract, and #890's update/drain/rollback works identically across both adapters. Brainstorm the contract + adapter boundary before building; do not interrupt the current storybook/image-gen line.

hognek:
## Summary

Adds a **bare-metal deployment option** for TinyAgentOS workers — no containers, no Incus, no Docker. The worker agent runs directly on the host, avoiding container overhead on constrained nodes (e.g., CPU-only Qwen3-Embedding-8B on a low-power node).

Fixes #892.

## What's new

- **`tinyagentos/containers/native.py`** — `NativeBackend` implements the full `ContainerBackend` ABC using systemd service units and direct subprocess calls.
  - `create_container()` writes a systemd unit, `exec_in_container()` runs commands directly on the host, `push_file()` copies files natively, etc.
  - Gracefully falls back when systemd is unavailable (e.g., CI, container-in-container) with sensible no-op defaults.
- **`containers/backend.py`** — registered `"native"` in `configure_container_runtime()`. Native is **config-only** — never auto-detected (every host is bare-metal-capable, so auto-detect would shadow LXC/Docker).
- **`containers/__init__.py`** — exported `NativeBackend`.
- **`routes/settings.py`** — added `"native"` to the allowed runtime list and backend-switching logic.

## Tests

- **`test_container_native.py`** — 24 new tests covering:
  - E2E fallback path (no systemd): create, list, exec, push, start/stop/destroy, proxy, snapshots, quota, env
  - Systemd-mocked path: unit file creation with memory/cpu/env, destroy, rename, set_env (add + replace), logs
  - Helper functions: `_service_unit_path`, `_has_systemd`
- **`test_container_runtime_config.py`** — 2 new tests: config override to native, non-auto-detection guarantee
- All existing container tests (83), worker tests (101), and detection tests (49) pass — **zero regressions**.

## Usage

Set in `config.yaml`:
```yaml
container_runtime: native
```
Or via the settings API:
```
PUT /api/settings/container-runtime  {"runtime": "native"}
```

### DIFF
--- a/tests/test_container_native.py
+++ b/tests/test_container_native.py
@@ -1,0 +1,321 @@
+"""Tests for the bare-metal NativeBackend."""
+
+from __future__ import annotations
+
+import os
+import tempfile
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from tinyagentos.containers.native import (
+    NativeBackend,
+    _has_systemd,
+    _service_unit_path,
+    _SYSTEMD_DIR,
+)
+
+
+class TestNativeBackendE2E:
+    """End-to-end tests against the real filesystem (no mocks).
+
+    These tests write unit files into a temporary directory and clean up
+    afterwards.  They do NOT require systemd — they test the backend's
+    logic directly when systemd is unavailable (the fallback path).
+    """
+
+    @pytest.fixture
+    def backend(self, tmp_path, monkeypatch):
+        """Return a NativeBackend with a temp _SYSTEMD_DIR."""
+        import tinyagentos.containers.native as _mod
+        monkeypatch.setattr(_mod, "_SYSTEMD_DIR", tmp_path)
+        # Force systemd-unavailable path so tests don't need root
+        monkeypatch.setattr(_mod, "_has_systemd", lambda: False)
+        return NativeBackend()
+
+    # --- create_container ---
+
+    @pytest.mark.asyncio
+    async def test_create_container_no_systemd(self, backend):
+        result = await backend.create_container("taos-agent-foo")
+        assert result["success"] is True
+        assert result["name"] == "taos-agent-foo"
+        assert "systemd not available" in result.get("note", "")
+
+    # --- list_containers (fallback path) ---
+
+    @pytest.mark.asyncio
+    async def test_list_containers_fallback_empty(self, backend, tmp_path):
+        result = await backend.list_containers("taos-agent-")
+        assert isinstance(result, list)
+        assert len(result) == 0
+
+    @pytest.mark.asyncio
+    async def test_list_containers_fallback_finds_units(self, backend, tmp_path, monkeypatch):
+        import tinyagentos.containers.native as _mod
+        # Create two service files
+        (tmp_path / "taos-agent-bar.service").write_text("[Unit]\n")
+        (tmp_path / "taos-agent-baz.service").write_text("[Unit]\n")
+        (tmp_path / "other.service").write_text("[Unit]\n")  # should not match
+        monkeypatch.setattr(_mod, "_has_systemd", lambda: False)
+
+        result = await backend.list_containers("taos-agent-")
+        names = {c.name for c in result}
+        assert names == {"taos-agent-bar", "taos-agent-baz"}
+
+    # --- exec_in_container ---
+
+    @pytest.mark.asyncio
+    async def test_exec_in_container_runs_command(self, backend):
+        code, output = await backend.exec_in_container(
+            "ignored", ["echo", "hello", "world"], timeout=10,
+        )
+        assert code == 0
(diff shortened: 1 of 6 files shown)


hognek:

jaylfc:
Did a deep pass on this. As dormant opt-in code it is safe (no auto-detect, existing backends untouched), but a couple of things before it should be selectable:

1. Host-mutation surface: once native is chosen, the agent deployer runs apt/pip installs and pushes files directly onto the controller host, and push_file writes to arbitrary absolute paths with no traversal guard. Please add a path-traversal guard on push_file/remote_path and an upfront root/writable-unit-dir check, plus a class-level SECURITY note so callers do not carry the LXC/Docker 'name is the boundary' assumption (name is decorative on exec/push/spawn_pty here).
2. Supervision is a no-op: ExecStart=/bin/true with Restart=no means create/start return success while nothing actually runs the agent. Either wire the unit to supervise a real payload, or return success:False while it is still a stub, so 'deployed and running' is not reported falsely.
3. Two real correctness bugs Kilo flagged: list_containers reads the wrong systemd field (SUB instead of ACTIVE) so it misreports state, and the async PTY close does a blocking proc.wait(timeout=5) that stalls the event loop.

Also worth confirming scope: issue #892 describes a cluster worker capability contract (register/advertise/drain/apply_update/health-check for model-serving backends), whereas this implements a ContainerBackend ABC used by the agent deployer. Related but a different abstraction, so I would hold off marking #892 fixed. Happy to merge once the isolation posture and supervision are sorted.

hognek:
Makes sense — will add the path-traversal guard + writable-dir precheck + the SECURITY note that name is decorative here, and either wire real supervision or return success:False while it is a stub rather than report deployed-and-running. Will also fix the SUB-vs-ACTIVE state read and the blocking proc.wait in the async PTY close. And agreed on scope: leaving #892 open — this is the ContainerBackend ABC, not the cluster-worker capability contract.

hognek:
Superseded by #1713, which contains this same NativeBackend implementation PLUS your requested review fixes [path-traversal guard on push_file/remote_path + writable-dir precheck + SECURITY note that name is decorative; real supervision or success:False for the stub; SUB->ACTIVE state read; non-blocking PTY close]. Please review #1713 instead — closing this to avoid two PRs for the same work. Scope note still stands: this is the ContainerBackend ABC, not the #892 cluster-worker capability contract, so #892 stays open.

PR #1695, opened 2026-07-06
fix: surface divergence message when branch diverged from tracked remote

### LINKED ISSUE #841: Update check silently shows 'no updates' when the local branch has diverged from origin
## Symptom

A Pi on `dev` showed "no updates available" despite `dev` being 30 commits ahead. Root cause: the local branch had **diverged** from `origin/dev` (5 local commits not on origin, after an upstream force-push removed them), so it was 5 ahead + 30 behind.

## Why

`auto_update.remote_is_strictly_ahead()` only flags an update when local HEAD is a **strict ancestor** of `origin/<branch>`. That guard is correct for the "running ahead on dev while master's tip is older" case, but it also makes a **diverged** branch report nothing, with no explanation.

## Fix

When HEAD is neither equal to, an ancestor of, nor a descendant of `origin/<branch>` (i.e. genuinely diverged), surface a distinct state to the UI, e.g. `diverged: true` with a message like "Your install has diverged from origin/<branch> (likely after an upstream history rewrite). Reset to recover." The Updates panel can then offer a one-click "reset to origin/<branch>" (it already has restart/branch-switch machinery), instead of silently showing "up to date".

Recovery today is manual: `git fetch origin && git reset --hard origin/<branch>` in the install dir.

hognek:
Task: t_b5c4c9b4. Fixes #841. Tests: 27/27 pass.

----
## Summary by Gitar

- **Divergence Handling:**
  - Added logic to detect and surface divergence messages when a branch deviates from its tracked remote.
  - Updated the UI feedback layer to display these messages in `status_reporter.py` and `branch_manager.py`.

### DIFF
--- a/tests/test_docs_only_update.py
+++ b/tests/test_docs_only_update.py
@@ -143,6 +143,11 @@ async def test_update_check_hides_docs_only_diff(client, monkeypatch):
         fake_docs_only,
         raising=False,
     )
+    monkeypatch.setattr(
+        "tinyagentos.auto_update.branch_is_diverged",
+        AsyncMock(return_value=False),
+        raising=False,
+    )
 
     r = await client.get("/api/settings/update-check")
     assert r.status_code == 200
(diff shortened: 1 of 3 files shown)


jaylfc:
@coderabbitai review

PR #1699, opened 2026-07-06
fix(models): write sidecar marker for rkllama registry corroboration

### LINKED ISSUE #1548: taOS Desktop cannot install models after a fresh installation (downloads complete instantly, models are not available)

hognek:
Fixes #1548: after fresh install, rkllama model downloads complete instantly but models are not available for selection.

## Problem

When rkllama installs a model, the weight goes to rkllama's own store (not models_dir). If catalog.refresh() fails silently, has_live_evidence stays False, has_disk_evidence is False, and registry_corroborated fails — the download shows complete but the variant never flips to "downloaded".

## Fix

Instead of a phantom .bin file in models_dir (which appeared as a 0 MB entry in the UI and whose delete removed only the marker), write a **sidecar metadata marker** at `models_dir/.taos-installed/<app_id>/<variant_id>.json` and add `_scan_sidecar_evidence()` that provides disk evidence for registry corroboration without polluting the downloaded-files list.

### Changes

- **`tinyagentos/routes/models.py`**:
  - Added `_scan_sidecar_evidence()` — scans `.taos-installed/` for app_ids with marker files
  - Added `sidecar_installed` kwarg to `_model_to_dict` for sidecar-backed disk evidence
  - Replaced phantom `.bin` marker with sidecar `.json` marker at `.taos-installed/<app_id>/<variant_id>.json`
  - Used variant's real `format` field instead of hardcoded `.bin`
  - Added path-traversal validation: rejects `/`, `\\`, and `..` in `app_id`/`variant_id`
  - Surface marker-write failures in install result (not just logged)
  - Used `json.dumps({...})` instead of `%`-template
  - `list_models()` now scans sidecars and passes to `_model_to_dict`

- **`tests/routes/test_models_download.py`** (new, 14 tests):
  - Sidecar detection (3 tests): `_scan_sidecar_evidence` returns app_ids, empty set, ignores empty dirs
  - Disk evidence (3 tests): sidecar-backed registry corroboration, no false positive without registry, sidecar not in downloaded files
  - `_attach_model_ids` (2 tests): bin and rkllm format matching
  - Path traversal (6 tests): rejects `/`, `\\`, `..` in both fields; accepts clean ids
  - Hoisted `FakeManifest`/`FakeHW` to module fixtures (Kilo feedback)

## Tests

55 tests pass (14 new + 30 existing model routes + 11 installation state).

### DIFF
--- a/tests/routes/test_models_download.py
+++ b/tests/routes/test_models_download.py
@@ -1,0 +1,215 @@
+"""Tests for the model download pipeline, including the rkllama sidecar-marker
+fix for #1548: when catalog.refresh() fails silently after a successful
+rkllama install, the variant never flips to "downloaded" because both
+has_disk_evidence and has_live_evidence are False.
+
+The fix writes a sidecar metadata marker at
+``models_dir/.taos-installed/<app_id>/<variant_id>.json`` so
+``_scan_sidecar_evidence()`` provides disk evidence for registry
+corroboration without appearing as a phantom 0 MB downloaded file.
+"""
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import pytest
+
+from tinyagentos.routes.models import (
+    _attach_model_ids,
+    _model_to_dict,
+    _scan_sidecar_evidence,
+    get_downloaded_models,
+)
+
+
+# ---------------------------------------------------------------------------
+# Module-level fixtures (Kilo: hoisted from inline classes)
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def rkllama_variant() -> dict:
+    """Standard rkllama variant dict with ``format: rkllm``."""
+    return {"id": "w8a8", "format": "rkllm", "size_mb": 3500}
+
+
+@pytest.fixture
+def qwen_variant() -> dict:
+    """Legacy variant with ``format: bin`` for backward-compat testing."""
+    return {"id": "w8a8", "format": "bin", "size_mb": 3500}
+
+
+class FakeManifest:
+    """Lightweight stand-in for an AppManifest used by _model_to_dict."""
+
+    def __init__(self, id: str = "qwen2.5-3b-rkllm", variants: list[dict] | None = None):
+        self.id = id
+        self.name = "Qwen 2.5 3B"
+        self.version = "1.0"
+        self.description = "test"
+        self.capabilities = ["chat"]
+        self.hardware_tiers = {}
+        self.variants = variants or [{"id": "w8a8", "format": "rkllm", "size_mb": 3500}]
+
+
+class FakeHW:
+    """Lightweight stand-in for a HardwareProfile used by _model_to_dict."""
+    ram_mb = 8192
+
+    class CPU:
+        soc = "rk3588"
+
+    cpu = CPU()
+
+
+# ---------------------------------------------------------------------------
+# Sidecar marker detection
+# ---------------------------------------------------------------------------
+
+
+class TestSidecarEvidence:
+    """A sidecar marker at .taos-installed/<app_id>/<variant_id>.json must
+    provide disk evidence for registry corroboration without appearing
+    as a phantom downloaded file (#1548)."""
+
+    def test_scan_returns_app_ids_with_markers(self, tmp_path):
+        """_scan_sidecar_evidence returns app_ids that have marker files."""
+        sidecar_dir = tmp_path / ".taos-installed" / "some-model"
+        sidecar_dir.mkdir(parents=True)
+        (sidecar_dir / "q4.json").write_text(
+            json.dumps({"source": "rkllama", "app_id": "some-model", "variant_id": "q4"})
+        )
+        ids = _scan_sidecar_evidence(tmp_path)
+        assert "some-model" in ids
+
+    def test_scan_returns_empty_for_no_markers(self, tmp_path):
+        """No markers → empty set."""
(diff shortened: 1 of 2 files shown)


hognek:

hognek:
recheck

jaylfc:
Thanks Hogne, the belt-and-suspenders intent here is good (a marker so a silent catalog-refresh failure doesn't leave the model invisible). A few things to sort before this can go in:

**1. Path traversal (Kilo CRITICAL, must fix).** `body.app_id` and `body.variant_id` are unsanitized request input interpolated straight into the marker path, and with `mkdir(parents=True)` a value like `../../` escapes `models_dir`. Validate them against the catalog (the app_id/variant must exist) or reject anything with a path separator before building the path.

**2. The `.bin` suffix won't match rkllm variants.** rkllama variants usually declare `format: rkllm`, so a `{app_id}-{variant_id}.bin` marker won't line up with the `{id}-{variant}.{format}` naming that `_attach_model_ids` / `_model_to_dict` match on. Use the variant's real `format` (the test passes only because its fixture manifest uses `format: bin`).

**3. Don't swallow the write failure silently.** The bare `except Exception` around the marker write means a read-only or full disk leaves the exact #1548 symptom in place with no signal. At least surface it in the install result.

**Bigger picture:** #1704 just merged, which makes taOS scan wherever the rkllama service actually writes (read from its systemd unit), so a genuinely-installed model is now discovered by its real file. That covers the primary path, so this marker is really a narrower fallback for the 'install succeeded but catalog.refresh() failed and the real file isn't scannable' case. Worth rebasing on current dev and re-framing it as that fallback. One option worth considering: a sidecar metadata marker that `_model_to_dict` reads as evidence, rather than a fake `.bin` in `models_dir` that itself shows up as a 0 MB downloaded file (and where a UI delete removes the marker but not the real model).

Rebase + the three fixes above and I'll take another look.

hognek:
Addressed all owner review feedback (2026-07-07):

**FIX 1 (path traversal)**: Added sanitisation loop that rejects `/`, `\\`, and `..` in both `app_id` and `variant_id` before building the marker path. Invalid values set `result.success = False` + error message and return early.

**FIX 2 (format field)**: No longer hardcoded `.bin` — uses `variant.get("format", "rkllm")` from the variant dict. The sidecar file is always `.json` (not a model file) so the format field is stored as metadata inside it.

**FIX 3 (surface failures)**: The bare `except Exception` now captures `exc` and appends to `result["warnings"]` so callers can see when the marker write failed.

**DESIGN (sidecar)**: Replaced the phantom `.bin` in `models_dir` with a metadata marker at `models_dir/.taos-installed/<app_id>/<variant_id>.json`. A new `_scan_sidecar_evidence()` helper returns the set of app_ids with marker files, and `list_models()` passes `sidecar_installed=(m.id in sidecar_ids)` to `_model_to_dict`. The sidecar is NOT in `_MODEL_FILE_SUFFIXES` so it never appears as a phantom downloaded file, and the DELETE handler won't touch it.

**Kilo feedback**: `json.dumps({...})` instead of `%`-template. `FakeManifest`/`FakeHW` hoisted to module fixtures.

**Tests**: 14 new tests (6 sidecar detection + 2 attach + 6 path-traversal), 55 total pass. Ready for re-review.

jaylfc:
Thanks for this, and for the careful path-traversal handling in the marker write.

Holding off on merging this one: #1704 landed on dev first and resolves #1548 by a more general route. Instead of writing our own sidecar marker at install time, #1704 adds rkllama's real service models directory as a scan root, so the actual weight files are discovered as disk evidence. That also covers the harder case from #1548 (models downloaded by the rkllama service before the taOS update), which a marker written at taOS-install-time cannot retroactively cover.

So this sidecar path would mostly duplicate #1704 for fresh taOS-driven installs, while the pre-existing-install case is the one users actually hit. There is also a stale-state risk: a marker persists even if the underlying rkllama model is deleted out of band, showing a model as "downloaded" when it is not; #1704's real-dir scan is self-correcting.

One thing that would change my mind: if you can show a concrete case where #1704's discovery misses a real rkllama models dir (e.g. its manifest-id/filename match fails against the actual on-disk names), that is a real gap worth a targeted fix to the match logic or the installer write location. In that case a couple of items from the Kilo review would need folding first: guard `_scan_sidecar_evidence` against FS errors (an unhandled OSError there 500s all of /api/models), keep `success=False` when the marker write fails, and point the traversal test at the real inline check rather than a duplicated local helper.

Closing for now as covered by #1704 - reopen if the discovery-miss case is real. Appreciate the work.

PR #1703, opened 2026-07-06
feat(chat): per-agent framework slash commands via @agent / prefix in groups

### LINKED ISSUE #840: Chat: per-agent framework slash commands (Telegram-style menu) in DMs and via @agent /
## What

Surface an agent's framework-exposed slash commands as an autocomplete menu in chat, the way Telegram shows a bot's command list.

- **In a DM with an agent**: typing `/` shows that agent's available slash commands as a menu (command + description), filterable as you type.
- **In a group/topic channel**: typing `@<agentname> /` shows that agent's slash commands (the menu is scoped to the mentioned agent, since a group has many participants).
- Selecting a command inserts it; the agent's framework handles the command.

## Why

Each agent framework exposes its own slash commands; today the user has to know them. A Telegram-style menu makes per-agent capabilities discoverable in chat.

## How it fits the codebase

- A `SlashMenu` component already exists (`desktop/src/apps/chat/SlashMenu.tsx`) and `MessagesApp.tsx` already drives a `/command` autocomplete from `/api/frameworks/slash-commands`. This feature extends that:
  - The current slash menu is taOS-level (e.g. /help). This adds the **per-agent, framework-exposed** command set.
  - DM case: the channel's single agent member determines whose commands to show on a leading `/`.
  - Group case: detect the `@<agentname> /` pattern, resolve the mentioned agent, then show that agent's commands.
- Backend: need an endpoint (or extend the existing slash-commands route) that returns a given agent's framework slash commands (name, description, usage). Frameworks expose these differently, so this needs a per-framework adapter to enumerate them.
- Relates to the in-progress `@mention autocomplete` work (job 19) and the existing SlashMenu pattern.

## Open questions for design

- How each supported framework exposes its slash command list (adapter surface).
- Caching/refresh of an agent's command list.
- Whether `@agent /` should also work in DMs (redundant there) or only groups.

hognek:
Task: t_a6cd1609. Fixes #840.

### Backend
- Add `slash_commands` to all 10 frameworks that were missing them (generic, agent_zero, ironclaw, microclaw, moltics, nanoclaw, nullclaw, picoclaw, shibaclaw, zeroclaw) — every framework now exposes at least `/help`.
- New endpoint `GET /api/agents/{slug}/slash-commands` returns a single agent's command list, keyed by slug. Returns 404 for unknown agents.

### Frontend
- `SlashMenu`: accepts optional `scopedAgent` prop. When set, only commands for that agent are shown, with the agent header visible (makes it clear whose commands are shown in an `@agent /` scenario).
- `MessagesApp`: in group channels (3+ members), the slash menu now only opens for the `@agentname /` pattern — bare `/` no longer shows all agents' commands, avoiding confusion about which agent a command targets. In DMs, leading `/` still opens the single agent's menu as before.

### Tests
- 51 framework tests pass (frameworks.py + framework_manifest + slash_commands + routes_framework)
- 6 SlashMenu tests pass (including new `scopedAgent` test)

## Summary by CodeRabbit

* **New Features**
  * Added agent-specific slash commands, including a new endpoint for fetching commands for a single agent.
  * Slash command menus now respect chat context, showing the right agent’s commands in group chats and allowing simple slash commands in direct messages.

* **Bug Fixes**
  * Improved slash-command behavior so command menus open only when the input format matches the current channel type.
  * Command lists now display more clearly when scoped to a specific agent.

### DIFF
--- a/desktop/src/apps/MessagesApp.tsx
+++ b/desktop/src/apps/MessagesApp.tsx
@@ -1021,8 +1021,16 @@ export function MessagesApp({
 
   /* ---- typing emitter + slash menu derived state ---- */
   const emitTyping = useTypingEmitter(selectedChannel, "user");
-  const showSlash = input.startsWith("/");
-  const slashQuery = showSlash ? input.slice(1).split(/\s/, 1)[0] || "" : "";
+  // In a DM (2 members: user + 1 agent), a leading "/" opens the agent's
+  // slash menu.  In a group channel (3+ members), the user must prefix
+  // with "@agentname /" so we know which agent's commands to show.
+  const isDm = (currentChannel?.members?.length ?? 0) <= 2;
+  const showSlash = isDm ? input.startsWith("/") : /@\S+\s+\//.test(input);
+  // The agent scoped by "@agentname /" (group only; undefined in DM).
+  const slashAgent = !isDm && showSlash ? input.match(/@(\S+)\s+\//)?.[1] || undefined : undefined;
+  const slashQuery = showSlash
+    ? (isDm ? input.slice(1).split(/\s/, 1)[0] : input.slice(input.indexOf("/") + 1).split(/\s/, 1)[0]) || ""
+    : "";
 
   /* ---- mutex: settings vs thread panel ---- */
   const handleOpenSettings = () => {
@@ -2586,6 +2594,7 @@ export function MessagesApp({
                   commands={slashCommands}
                   queryAfterSlash={slashQuery}
                   members={currentChannel?.members || []}
+                  scopedAgent={slashAgent}
                   onPick={(slug, cmd) => {
                     setInput(`@${slug} /${cmd} `);
                   }}
--- a/desktop/src/apps/chat/SlashMenu.tsx
+++ b/desktop/src/apps/chat/SlashMenu.tsx
@@ -11,18 +11,24 @@ export function SlashMenu({
   commands,
   queryAfterSlash,
   members,
+  scopedAgent,
   onPick,
   onClose,
 }: {
   commands: SlashCommandsBySlug;
   queryAfterSlash: string;
   members: string[]; // agents in current channel (ordered)
+  /** When set, only commands for this agent are shown (for @agent / in group channels). */
+  scopedAgent?: string;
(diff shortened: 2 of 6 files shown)


hognek:
Updated: amended to add 3 backend tests for the new `GET /api/agents/{slug}/slash-commands` endpoint. Test count: 54 backend + 6 frontend SlashMenu tests pass locally. CLA re-verified on amended commit. CI matrix still needs maintainer approval (action_required on run 28831787254).

jaylfc:
Nice feature, and per-agent scoping is the right call for group channels. Three checks are red and all three are genuine (not flaky, not CI-gating):

1. `spa-build`: `MessagesApp.tsx(1027): TS2448 'currentChannel' used before its declaration`. The new slash-derived block (isDm / showSlash / slashAgent / slashQuery around 1024-1030) reads `currentChannel`, but that const is declared later at ~1422 (temporal dead zone). Move the derived block below the `currentChannel` declaration (it is only consumed by the SlashMenu render ~2588), or hoist the declaration above 1024.
2 and 3. `test (3.12)` and `test (3.13)`: both fail on the same assertion, `test_framework_without_slash_commands` (tests/test_routes_framework.py:300). It asserts `generic` returns no slash commands, but this PR adds `/help` to `generic`, so the endpoint now returns `[{name: help ...}]`. Update that test to the new baseline (point it at a framework that truly has no slash_commands, or invert it to assert the `[help]` behavior). That single change clears both Python jobs.

Optional but worth folding while you are in there (from the Kilo review): anchor the `@token /` group regex so an older or quoted mention cannot hijack the menu; guard `isDm` against 0/1-member channels; and the new endpoint docstring says "unknown agent -> empty list" but the code returns 404, so fix the doc. Push those and the matrix should go green.

hognek:
CI fix PR is up: #1720 — resolves TS2448 TDZ, test_framework_without_slash_commands, and endpoint docstring.

jaylfc:
Superseded by #1720, which carries this feature plus the three CI fixes (TS2448 TDZ, the generic /help test baseline, endpoint docstring) and the folded Kilo items. #1720 is merged, so closing this. Thanks.

PR #1711, opened 2026-07-07
fix(cluster): authenticate + harden GPU lease endpoints (#1680 retro)

jaylfc:
## What

Retrospective hardening of the GPU lease coordination API (landed in #1680), from a deep-dive audit of the merged code.

The four lease endpoints (`claim` / `release` / `renew` / `GET leases`) were reachable **unauthenticated**, unlike every sibling cluster route. Any process on the LAN could `GET /api/cluster/leases` to enumerate every lease id and then release them, freeing all reservations and re-enabling the concurrent-GPU-load (OOM / Xid 62) the lease system exists to prevent.

The lease API has **no callers on dev yet** (the GPU arbiter that will consume it is not merged), so this closes the hole before it goes live rather than fixing a live exploit.

## Changes

- **Auth gate** on all four endpoints: a trusted principal is required, accepting an admin session, a valid local token (same-host trust, for controller-side dispatchers/agents), or a signed worker HMAC.
- **Wider lease ids** (`token_hex(4)` -> `token_hex(16)`) so ids cannot collide and free the wrong holder's reservation.
- **Input validation**: reject non-positive `ttl_seconds` (a zero ttl minted a lease that was already expired and granted no exclusivity) and negative `required_vram_mb` (bypassed the capacity guard) at the request boundary.
- **TOCTOU fix**: resolve the worker and re-check its status inside `_lease_lock`, so a claim cannot be admitted against a worker the monitor loop took offline while the claim waited on the lock.

## Tests

- Lease calls now authenticate via the local token.
- New: unauthenticated requests are rejected on every endpoint; zero/negative ttl and negative vram are rejected with 422.
- 31 lease + 78 cluster/pairing/worker-auth tests pass locally.

Follow-up (not in scope here): per-holder ownership binding on release/renew, deferred until the arbiter wires real callers and defines holder semantics.

## Summary by CodeRabbit

* **New Features**
  * Lease operations now require proper authorization, allowing access only for approved sessions or paired worker requests.
  * Lease requests now enforce a maximum time-to-live of 24 hours.

* **Bug Fixes**
  * Improved lease handling to reduce race conditions when a worker goes offline during lease assignment.
  * Lease identifiers are now harder to guess.
  * Updated lease API behavior and tests to cover authenticated access, validation, and error cases consistently.

### DIFF
--- a/tests/test_leases.py
+++ b/tests/test_leases.py
@@ -42,6 +42,15 @@ async def _heartbeat(client, key, name, **fields):
     return resp
 
 
+def _auth(app) -> dict:
+    """Authorization header carrying the app's local token.
+
+    The lease endpoints require a trusted principal (admin session, local
+    token, or worker HMAC); the local token is the simplest in tests and
+    stands in for a same-host controller-side caller (dispatcher/agent)."""
+    return {"Authorization": f"Bearer {app.state.auth.get_local_token()}"}
+
+
 @pytest.mark.asyncio
 async def test_claim_lease_success(client, app):
     """Claiming a lease on an online worker with enough VRAM returns 200."""
@@ -54,7 +63,7 @@ async def test_claim_lease_success(client, app):
     # Send a heartbeat with free VRAM
     await _heartbeat(client, key, "gpu-node", free_vram_mb=6000, used_vram_mb=2000)
 
-    resp = await client.post("/api/cluster/leases/claim", json={
+    resp = await client.post("/api/cluster/leases/claim", headers=_auth(app), json={
         "resource_id": "gpu-node:gpu-cuda-0",
         "caller": "skald-dispatcher",
         "ttl_seconds": 30,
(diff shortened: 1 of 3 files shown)


jaylfc:
@coderabbitai review

jaylfc:
@coderabbitai full review

jaylfc:
Tracking the two Kilo WARNINGs as follow-ups rather than blocking this PR:

1. Per-holder ownership binding on release/renew is deferred until the arbiter wires real callers and can define holder semantics (noted in the PR body). Until then any authenticated principal can act on any lease, which is a weaker threat than the unauthenticated-LAN access this PR closes.
2. The gate returns a generic 403 on any auth failure. That is a deliberate do-not-reveal-which-method choice, but it does drop the specific HMAC diagnostic a legit worker would want when debugging pairing.

Both are non-blocking; the load-bearing fix (no unauthenticated access to the lease endpoints) is in place.

Issue #174, opened 2026-04-12
X Monitor app (Knowledge Pipeline Step 5)


jaylfc:
## Summary
Build the X Monitor app — cookie auth, save threads, monitor engagement, author tracking.

## Status
Implementation plan written at `docs/superpowers/plans/2026-04-12-x-monitor.md`.

## Dependencies
- Knowledge Base Service (done)
- Agent Browsers (needed for cookie auth)

## Spec
`docs/superpowers/specs/2026-04-12-knowledge-capture-pipeline-design.md` — X Monitor section

jaylfc:
Implemented — XFetcher + watch store + routes + XMonitorApp

PR #1712, opened 2026-07-07
fix(governance): delegation + org-model hardening (#1661/#1662 retro, #174)

jaylfc:
Retrospective audit fixes for the merged agent org model (#1661) and heartbeat (#1662), all verified against current `dev`. Each fix ships with a test. Relates to #174.

## Fixes

**1. CRITICAL - deny-event audit rows were unfindable.** On a policy `deny`, `delegation.py` recorded the board-audit event with a synthetic `task_id=f"agent:{from_agent}"` and no `project_id` (defaults to `""`). Since `recent_for_project` filters on `project_id`, deny events never appeared in any project feed. Now the deny is keyed on the real `task_id` when a task is present (so it also shows in the task's history) and always carries the real `project_id`; with no task it records a `policy:delegate` event rather than a fabricated agent-as-task id.

**2. WARNING - raw task id leaked to the human approval inbox.** When delegating an existing task with no `task_title`, the blocking Decision question read "wants to delegate `<raw task_id>`". The task's title (already fetched in `delegate_task`) is now threaded into `_check_delegation_policy` and used, falling back to `task_title` then `task_id`.

**3. MEDIUM - assignee identity mismatch (from #1662).** `complete_delegation` wrote `assignee_id=to_agent`, but that is the agent NAME (slug), while beads, project members, and the heartbeat sweep (`list_ready_tasks_for_assignee`) all key on the config HEX id (`config.py` mints `uuid4().hex[:12]`). Delegated tasks were therefore invisible to the heartbeat and id-keyed member lookups. The name is now resolved to the hex id for `assignee_id` (name kept for display/notify), with a graceful fallback when the agent has no distinct id (`id == name`).

**4. WARNING - set_reporting TOCTOU cycle race.** The cycle guard walked the reporting chain via multiple `await self.get(...)` reads and then issued a separate UPDATE with no serialization. On the shared aiosqlite connection the awaits yield, so two concurrent edits (A->B and B->A) could both pass the check then both write, persisting a cycle. The check-then-write is now held under an `asyncio.Lock`.

**5. WARNING - PUT /org empty-body no-op + no "" normalization.** A body of `{}` (all fields None) returned 200 while changing nothing, and `role`/`title` were stored verbatim including whitespace, whereas the codebase convention (applied to `reports_to`) is that `""` clears a nullable field. An all-empty body is now rejected with 400, and empty/whitespace-only `role`/`title` clear the field (NULL).

## Tests

- `tests/test_routes_delegation.py`: fixtures given distinct `id` vs `name` (which masked #3); new tests for the findable deny audit event, the title-not-raw-id approval question, and the `id == name` fallback.
- `tests/test_agent_registry_store.py`: empty-string/whitespace clear for `set_role_title`; concurrent A->B / B->A edits cannot persist a cycle.
- `tests/test_routes_agent_org.py`: empty body 400, clear role via `""`, whitespace-only title cleared.

All affected and adjacent suites pass (`test_routes_delegation`, `test_routes_agent_org`, `test_agent_registry_store`, `test_agent_registry`, `test_board_audit`, `test_routes_decisions`, `test_agent_heartbeat`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Updating agent org fields now rejects `{}` (HTTP 400) and properly clears `role`/`title` when empty or whitespace is provided, returning `null` in responses.
  * Concurrent reporting-line updates are now protected against persisting circular reporting relationships.
  * Delegation handling now uses stable agent identifiers for task assignment and produces more discoverable denial/audit events; approval prompts use the existing task’s title when available.
* **Tests**
  * Added coverage for role/title clearing and concurrent reporting-cycle prevention, plus updated route/delegation scenarios and audit/approval behaviors.

### DIFF
--- a/tests/test_agent_registry_store.py
+++ b/tests/test_agent_registry_store.py
@@ -986,6 +986,21 @@ class TestSetRoleTitle:
         result = await store.set_role_title("no-such-20260101-000000", title="X")
         assert result is None
 
+    @pytest.mark.asyncio
+    async def test_empty_string_clears_field(self, store):
+        row = await store.register(framework="openclaw", display_name="A", role="worker")
+        cleared = await store.set_role_title(row["canonical_id"], role="")
+        assert cleared["role"] is None
+
+    @pytest.mark.asyncio
+    async def test_whitespace_only_is_treated_as_set_not_clear(self, store):
+        # set_role_title clears only on a truly empty string; the route strips
+        # whitespace-only input to "" before calling, so raw "   " here stays
+        # verbatim (the store does not itself strip).
+        row = await store.register(framework="openclaw", display_name="A", role="worker")
+        updated = await store.set_role_title(row["canonical_id"], role="   ")
+        assert updated["role"] == "   "
+
 
 # ---------------------------------------------------------------------------
 # set_reporting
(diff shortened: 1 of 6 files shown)


jaylfc:
Folded the _reporting_lock WARNING: the lock is now created in __init__ so it always exists before any caller, independent of init ordering.

Tracking the remaining Kilo WARNING as a follow-up: _resolve_agent_id falls back to the stale slug if a config agent is renamed or removed between a delegation gate's creation and its approval, which would leave that one delegated task invisible to id-keyed lookups. Low impact (needs an agent to vanish mid-approval) and resolving it cleanly means capturing the id at gate-creation time, which is a larger change than this PR's scope.

jaylfc:
@coderabbitai review

PR #1713, opened 2026-07-07
fix(containers): NativeBackend security + supervision + kilo bug fixes

hognek:
Task: t_f310e934. Fixes #1691.

Five fixes on the NativeBackend before it becomes selectable:

**SECURITY (3 items):**
- `push_file` now passes `remote_path` through `_safe_host_path()`, rejecting `..` traversal and writes outside `_ALLOWED_WRITE_ROOTS` (opt/taos/, /etc/systemd/system/, /var/lib/taos/, /tmp/). Symlink escapes are caught via `realpath`.
- `create_container` checks `_systemd_dir_is_writable()` upfront (lazily, at create time) and fails early if the unit directory is read-only.
- Class-level `.. SECURITY::` docstring noting that `name` is DECORATIVE — no container boundary on exec/push/spawn_pty.

**Supervision:**
- Unit stub changed from `ExecStart=/bin/true` + `Restart=no` (silent NO-OP) to `ExecStart=/bin/false` + `Restart=no`. A bare `start_container` now fails rather than returning success with nothing running.
- `create_container` returns a "stub unit created" note telling the caller to update ExecStart before starting.
- `set_env(key="ExecStart", value=...)` now replaces the ExecStart line so the deployer can wire the real agent payload after install.

**Kilo bugs:**
- `list_containers`: reads ACTIVE column (`parts[2]`) instead of SUB (`parts[3]`), correctly mapping "inactive"/"failed" to Stopped.
- `_NativePtyHandle.close()` and `_IncusPtyHandle.close()`: guard `proc.wait(timeout=5)` with `TimeoutExpired` → kill + re-wait; added docstring noting async callers should use `asyncio.to_thread`.

Tests: 57/57 pass (38 native + 19 runtime-config/PTY).
SCOPE: does NOT mark #892 fixed — #892 is a cluster-worker capability contract.

## Summary by CodeRabbit

* **New Features**
  * Added a new **native** container runtime option for running workloads directly on the host.
  * The runtime settings UI/API now recognizes and can switch to the native backend.

* **Bug Fixes**
  * Improved container path handling to better prevent unsafe file access.
  * Updated runtime detection messaging to include the native option.
  * Refined container lifecycle behavior and log handling for more consistent results.

### DIFF
--- a/tests/test_container_native.py
+++ b/tests/test_container_native.py
@@ -11,10 +11,62 @@ import pytest
 
 from tinyagentos.containers.native import (
     NativeBackend,
+    _ALLOWED_WRITE_ROOTS,
     _has_systemd,
+    _safe_host_path,
     _service_unit_path,
     _SYSTEMD_DIR,
 )
+
+
+class TestSafeHostPath:
+    """Tests for the path-traversal guard."""
+
+    def test_rejects_relative_path(self):
+        assert _safe_host_path("foo/bar") is None
+        assert _safe_host_path("../etc/passwd") is None
+
+    def test_rejects_dotdot_traversal(self):
+        assert _safe_host_path("/opt/taos/../../etc/passwd") is None
+        assert _safe_host_path("/tmp/../etc/shadow") is None
+
+    def test_rejects_outside_allowed_roots(self):
+        assert _safe_host_path("/etc/passwd") is None
+        assert _safe_host_path("/home/user/.bashrc") is None
+        assert _safe_host_path("/var/log/syslog") is None
+
+    def test_accepts_allowed_roots(self):
+        assert _safe_host_path("/opt/taos/config.yaml") == "/opt/taos/config.yaml"
+        assert _safe_host_path("/tmp/test.txt") is not None
+        assert _safe_host_path("/etc/systemd/system/foo.service") is not None
+        assert _safe_host_path("/var/lib/taos/agents/foo/data") is not None
+
+    def test_accepts_resolved_symlink_within_root(self, tmp_path, monkeypatch):
+        # Create a symlink inside an allowed root that points within the same root
+        allowed = tmp_path / "opt" / "taos"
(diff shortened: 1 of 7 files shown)


hognek:
This supersedes #1691 — same NativeBackend impl plus all the review fixes you flagged there [security path-traversal guard + writable-dir precheck + decorative-name SECURITY note, real-supervision-or-success:False, SUB->ACTIVE state read, non-blocking async PTY close]. Closed #1691 to keep it to one PR.

jaylfc:
Reviewed this in depth, nice work. Confirmed all five fixes are correctly applied: the _safe_host_path guard fails closed against traversal, outside-root, and symlink-prefix attempts; the writable-dir precheck and SECURITY notes are in; the stub no longer reports a false running state; and list_containers now reads the ACTIVE column with the PTY close offloaded. I approved the CI so pytest actually validates on Linux (it had not run yet, since fork PRs need approval). Two things before I merge:

1. Escape env values in create_container/set_env: a newline in a value currently injects arbitrary systemd directives into the unit file. No privilege boundary is crossed (an env-setter already has exec_in_container), but it is a real defense-in-depth gap on a security-hardening PR, so worth closing here. Reject or strip newlines.
2. Wrap the post-kill wait(timeout=2) in both native.py and lxc.py in try/except TimeoutExpired so a stuck kill does not propagate out of close().

The TOCTOU parent-symlink race and the Type=simple stub caveat are fine as follow-ups given the backend is dormant/opt-in. Once CI is green and those two are in, I will merge.

jaylfc:
Merged, thanks for the fast turnaround on the env escaping and the wait guards. One trivial follow-up whenever you are next in native.py: the create_container docstring still says the failure shape returns a note field, but the error branches return error, so a caller reading the docstring would look up the wrong key. Doc-only, no runtime impact, so I did not hold the merge on it.

Issue #1710, opened 2026-07-07
taOS Agent returns "the agent backend returned no output" despite backend, model, and OpenCode being available


mandresve:
## Environment

- **Version:** v1.0.0-beta.34
- **Platform:** Orange Pi 5 Pro (RK3588)
- **OS:** Debian 12 (Bookworm)
- **Installation:** Clean installation using the latest one-line installer.

## Description

After resolving the model discovery issues (incorrect RKLLM provider endpoint), the taOS Agent is now able to detect the local model correctly.

The agent configuration shows:

- OpenCode installed.
- Local RKLLM backend running.
- Model successfully selected.
- Model added to the permitted models list.

However, the taOS Agent itself is still unable to generate any response.

Sending even a simple prompt such as `Hi` immediately returns:

```text
Error: the agent backend returned no output;
if taOS just restarted the model proxy may still be warming,
try again shortly.
```

This behavior is persistent and is not related to the backend startup time. The backend has been running for several minutes before testing.

## Steps to Reproduce

1. Configure the RKLLM provider with the correct endpoint.
2. Verify the provider reports **Available**.
3. Download a local model.
4. Assign the model to the taOS system agent.
5. Add the model to the permitted models list.
6. Open the taOS Agent chat.
7. Send any prompt (e.g. `Test`).

## Expected Behavior

- The taOS Agent should forward the request to the configured local model.
- The model should generate a response.

## Actual Behavior

- The message is sent successfully.
- The backend immediately returns:

```text
Error: the agent backend returned no output;
if taOS just restarted the model proxy may still be warming,
try again shortly.
```

- No response is ever generated.

At this point, the remaining failure seems to occur when the taOS Agent attempts to invoke the backend for inference.

<img width="949" height="655" alt="Image" src="https://github.com/user-attachments/assets/d1951b87-7520-4647-b01e-5a768f6ab67e" />

jaylfc:
Thanks for this one, and for the sharp diagnosis on the provider port and model discovery in your other reports. Those two (the stale 8080 port in #1697 and the model-discovery gap in #1548) are both fixed in v1.0.0-beta.34, and they matter here: if the RKLLM provider was unhealthy or the model was not discoverable, the agent would get an empty response, which is exactly this symptom. So the first ask is a retest: update to v1.0.0-beta.34, confirm the RKLLM provider shows healthy and your model appears under Downloaded Models, then try the agent again.

[...]

Issue #1697, opened 2026-07-06
Local RKLLM provider uses incorrect default endpoint


mandresve:
## Environment

- **Version:** v1.0.0-beta.33
- **Platform:** Orange Pi 5 Pro (RK3588)
- **OS:** Debian 12 (Bookworm)

## Description

While investigating the Providers issues, I found that the local RKLLM provider appears to be configured with an incorrect default endpoint.

The provider is configured to use:

```text
http://localhost:8080
```

However, the rkllama backend is actually running on:

```text
http://localhost:7833
```

After manually changing the endpoint to port **7833**, clicking **Test** succeeds immediately, indicating that the backend is healthy and reachable.

## Steps to Reproduce

1. Open **Providers**.
2. Select the local RKLLM provider.
3. Observe that the configured endpoint is `http://localhost:8080`.
4. Change the endpoint to `http://localhost:7833`.
5. Click **Test**.
6. Observe that the connection succeeds.
7. Click **Save**.

## Expected Behavior

- The local RKLLM provider should point to the correct default endpoint.

## Actual Behavior

- The provider is configured to use `http://localhost:8080`.
- Changing it to `http://localhost:7833` allows the connection test to succeed.

## Additional Information

This appears to be a separate issue from the model discovery bugs.

The successful connection on port **7833** suggests that the rkllama backend itself is functioning correctly. The problem seems to be that the provider is either initialized with an incorrect default endpoint.

This may also explain why the Providers page reports the local RKLLM backend as unavailable on my installation, despite the backend service being healthy.

<img width="951" height="640" alt="Image" src="https://github.com/user-attachments/assets/86b8de2c-2d1c-4573-92f0-609676b36516" />

<img width="946" height="645" alt="Image" src="https://github.com/user-attachments/assets/43085a9e-ea12-42d2-8f59-3ab7943a7bbb" />

mandresve:
Solved

PR #1715, opened 2026-07-07
fix(config): rkllama backend name local-npu -> local-rkllama so installed models reach LiteLLM (#1710)

jaylfc:
## What

Fixes the root cause of #1710 ("taOS Agent returns 'the agent backend returned no output'"), reported by mandresve.

## Root cause

LiteLLM registers an installed local chat model only when a backend named `local-<service-id>` exists whose service-id matches the `requires.backends[].id` the model's manifest declares. Every RKLLM model manifest declares `id: rkllama`, but the shipped seed named the rkllama backend **`local-npu`**. `_local_backend_models_from_registry` derives the service-id from the `local-` name suffix, so `local-npu` yields service-id `npu`, which never matches `rkllama`. Result: an installed RKLLM chat model gets **no LiteLLM alias**. The agent (via OpenCode) then calls `litellm/<model-id>`, LiteLLM has no such model, and the completion returns zero parts, surfacing as "the agent backend returned no output."

The model is installed and in the registry, and the manifest is correct. It is purely a backend-name mismatch. mandresve independently traced the failing hop to LiteLLM not exposing his chat model, which matches exactly.

## Fix (registry-first, design intact)

- Rename the seeded backend `local-npu` -> `local-rkllama` in `data/config.yaml.example` (fresh installs), so the service-id is `rkllama`.
- Add `_migrate_legacy_rkllama_backend_name` in `config.py` `load_config` to heal existing installs in place, mirroring the #1697 port migration. It also collapses the boot-time dedup collision with `auto_register`'s own `local-rkllama`.
- No discovery-based change: chat models continue to flow through the installed-model registry, which is the only supported source.

## Verification

- Confirmed on a live Pi install that its config carries the `local-npu` name (the exact mismatch).
- Tests: added a migration test (`local-npu` rkllama renamed, custom names + non-rkllama `local-npu` untouched); updated the `test_llm_proxy` fixture. 163 config/llm_proxy/litellm_config tests pass locally.

## Follow-up (out of scope)

The rkllama service manifest's `hardware_tiers` floor (16GB) can strand sub-16GB NPU boards from the auto-registered `local-rkllama`; lowering it to match the models it serves is a separate change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Legacy RKLLM backend configurations are now automatically normalized to the current backend name when loading saved settings.
  * Generated configuration output now shows the updated backend name, keeping backend lists consistent.
  * Existing custom backend names are preserved, and unrelated backends are not changed.

### DIFF
--- a/data/config.yaml.example
+++ b/data/config.yaml.example
@@ -10,7 +10,7 @@ server:
   browser_proxy_port: 6970
 
 backends:
-  - name: local-npu
+  - name: local-rkllama
     type: rkllama
     url: http://localhost:8080
     priority: 3
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -379,3 +379,31 @@ def test_load_config_migrates_legacy_rkllama_port(tmp_path):
     assert by_name["custom-rk"] == "http://10.0.0.5:8080"
     # non-rkllama backend on 8080 -> untouched
     assert by_name["ollama"] == "http://localhost:8080"
+
+
+def test_load_config_migrates_legacy_rkllama_backend_name(tmp_path):
+    """An install seeded before #1710 names the rkllama backend local-npu,
+    whose service-id (npu) never matches the RKLLM manifests'
+    requires.backends id (rkllama), so installed chat models get no LiteLLM
+    alias. load_config renames it to local-rkllama so the service-id matches."""
+    import yaml
+    from tinyagentos.config import load_config
+
+    cfg_path = tmp_path / "config.yaml"
+    cfg_path.write_text(yaml.safe_dump({
+        "backends": [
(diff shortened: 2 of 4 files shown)


jaylfc:
Merged. Two minor bot notes for the record, both non-blocking: the rename is in-memory each boot and re-logs on legacy installs, which is intentional and matches the existing _migrate_legacy_rkllama_port behavior (LiteLLM gets the correct name every boot; the config is only persisted when another path saves it). And a follow-up test for a local-npu rkllama coexisting with an already-present local-rkllama at a different url would be worth adding; the common same-url case is collapsed by the auto_register dedup.

Issue #185, opened 2026-04-12
Fix store catalog type taxonomy — live API categories broken


jaylfc:
## Summary
The Store app frontend expects fine-grained types (`voice`, `streaming-app`, `image-gen`, `dev-tool`, `home`, `monitoring`, `infrastructure`) but `catalog.yaml` only uses coarse types (`model`, `service`, `agent-framework`). This causes most categories to be empty when using the live API.

## Current state
- MOCK_APPS fallback has been fixed with placeholders for empty categories
- Kokoro duplicate ID fixed (renamed to `kokoro-tts-server`)
- Root cause remains: catalog.yaml entries need their `type` fields updated to match what the frontend expects, OR the frontend category→type mapping needs to accept coarse types and use a secondary field for sub-categorization

## Fix options
1. Update catalog.yaml entries with fine-grained types (whisper→voice, code-server→dev-tool, home-assistant→home, etc.)
2. Add a `category` field to manifests separate from `type`, let the frontend filter on that
3. Make the frontend category mapping accept multiple type levels

Option 1 is simplest and most correct.

jaylfc:
Implemented — catalog.yaml types updated, MOCK_APPS placeholders added

PR #1707, opened 2026-07-07
fix(scheduler): add _running_lock coverage to eviction paths in GpuArbiter

hognek:
## Summary

PR #1683 review feedback #5. Closes #894.

`_running` was mutated under `_running_lock` in `_run_gpu_task` but read/popped without the lock in `evict_lowest_priority`, `_evict_task`, `release_tasks_for_worker`, `stats`, and `running_tasks`. Safe under single-thread-no-await but fragile.

## Changes

- `gpu_arbiter.py`: Make `evict_lowest_priority`, `_evict_task`, `release_tasks_for_worker`, `stats`, `running_tasks` async; guard `_running` dict access with `async with self._running_lock:`. Lease release and task cancellation happen outside the lock to avoid deadlock.
- `cluster/manager.py`: `drain_worker` → async, await `release_tasks_for_worker`
- `routes/cluster.py`: await `stats()` and `running_tasks()`; await `drain_worker`
- Tests: update all callers — add `await` and `@pytest.mark.asyncio` where needed

## Tests

69/69 targeted tests pass: `test_gpu_arbiter_894.py`, `test_gpu_arbiter_796.py`, `test_gpu_arbiter.py`, `test_cluster_drain.py`

PR #1706, opened 2026-07-07
fix(scheduler): non-blocking drain + eviction-to-make-room in GPU arbiter

hognek:
## What

Fixes the queue processor serialization blocking drain (PR #1683 review feedback #4).

### Changes

1. **Non-blocking drain:** `_drain_queue` no longer awaits `_run_gpu_task`. The admitted task is spawned as a background `asyncio.Task` with a done-callback that propagates the result/exception to the submitter's `_arbiter_future`. The drain loop stays responsive on subsequent ticks.

2. **Eviction-to-make-room:** When admission fails for a queued task, `_drain_queue` now calls `evict_lowest_priority(min_priority=int(task.priority))` and re-checks admission. Lower-priority running tasks are evicted to make room; higher-priority runners are preserved.

3. **Documentation:** Added a docstring explaining the intentional serialization — only one task is admitted per drain cycle to avoid flooding the GPU with concurrent loads.

### Tests

- 6 new tests in `test_gpu_arbiter_894.py`:
  - `TestDrainQueueNonBlocking`: drain returns immediately, done_callback propagates results and exceptions
  - `TestEvictionToMakeRoom`: eviction triggered when VRAM full, no eviction of higher-priority runners, eviction disabled guard
- All 117 related tests pass (60 arbiter + 57 cluster/lease)

### Not changed

- Queue processing remains intentionally serial — one task per drain cycle
- `_evict_task` / `evict_lowest_priority` / `_run_gpu_task` are unchanged
- Pause/resume (#796) still gates the drain via `_process_queue`

## Summary by CodeRabbit

* **New Features**
  * Added GPU task admission control with queueing, priority-based scheduling, and eviction support when GPU memory is tight.
  * GPU resources are now detected and registered more reliably during startup, improving visibility of available hardware.
  * Added VRAM reservation tracking so GPU usage is accounted for more accurately across running and pending tasks.

* **Bug Fixes**
  * Improved startup resilience when resource setup is unavailable.
  * Fixed GPU admission behavior to avoid overcommitting memory during concurrent task submissions.

PR #1718, opened 2026-07-07
fix(cluster): async drain_worker/cancel_drain + _lease_lock on force-release + wire arbiter task cancellation

hognek:
## Summary

Fixes #1690 (lock fix) and cross-cutting arbiter wiring for the taOS arbiter epic (#1683/#1705/#1706/#1707/#1690).

### Changes

**#1690 lock fix** — `drain_worker` force-release and `_monitor_loop` stale-drain path now hold `_lease_lock` during `self._leases` mutation, serialized with claim/release/sweep. `drain_worker` and `cancel_drain` are now `async def` (consistent with `claim_lease`/`release_lease` — all callers were already in async handlers).

**Cross-cutting wiring** — Added `GpuArbiter.cancel_running_for_leases()` and wired it into:
- `drain_worker(graceful=False)` — after force-releasing leases
- `_monitor_loop` stale-drain branch — when a draining worker stops heartbeating

When leases are force-released, any running GPU arbiter tasks for those leases are cancelled via `_evict_task`. Eviction is no longer a prod no-op when the arbiter is running.

**Also:** Added `TYPE_CHECKING` import for `GpuArbiter` type annotation on `ClusterManager._gpu_arbiter`.

### Files changed
- `tinyagentos/cluster/manager.py` — async drain_worker/cancel_drain, lock fix, arbiter cancellation wiring
- `tinyagentos/routes/cluster.py` — await on async drain_worker/cancel_drain calls
- `tinyagentos/scheduler/gpu_arbiter.py` — new `cancel_running_for_leases()` method
- `tests/test_cluster.py` — await on all 10 drain_worker/cancel_drain test calls

### Tests
47/47 pass: `tests/test_cluster.py` (26), `tests/test_gpu_arbiter_894.py` (14), `tests/test_gpu_arbiter_toctou.py` (7)


hognek:
Addressed all 4 Kilo review findings in b6e17e6:

- **CRITICAL** (manager.py:666): Moved `worker.status = "offline"` inside `_lease_lock` in stale-drain path — prevents duplicate worker.leave on next monitor tick.
- **WARNING** (manager.py:447): Same fix applied to `drain_worker(graceful=False)` path.
- **WARNING** (manager.py:676) + **SUGGESTION** (gpu_arbiter.py:427): `cancel_running_for_leases` now returns `(cancelled, already_completed)` — operators can distinguish force-kills from natural completions. Both callers log both counts.

47/47 tests pass.

jaylfc:
Did a full SET review of the arbiter epic across every branch (#1683/1690/1705/1706/1707/1718/1725/1726/1727). #1718 is the one to converge on: it is the rebased superset composing #1683 + #1705 + #1706 + #1707 + #1690 plus pause/resume and the blocker-2/4 fixes, and kilo is clean on it (0 critical, 0 warning). My four original composed-state blockers are all confirmed fixed in #1718:
- evict_lowest_priority is now awaited (gpu_arbiter.py:445 async def; call site :515 `await`).
- post-eviction re-check uses `await _reserve_and_check` (:517), not _check_admission - TOCTOU closed.
- lease claim is inside _run_gpu_task's try/finally (:374/:400), and the cluster admission subtracts _reserved_vram_mb (:347) - leak closed.
- /drain, /cancel-drain, /update have _require_admin (:1124/1137/1158); _leases mutations are under _lease_lock.

Thank you - the mechanical blockers landed. But the epic is not mergeable as a set yet. Blocking:

1. THE EPIC GATE IS STILL OPEN (the #185 gate). GpuArbiter.submit_gpu has ZERO production callers in any branch - no route, scheduler, or model-load path admits work through the arbiter, so _running is always empty in prod and eviction + cancel_running_for_leases are no-ops. The arbiter ships inert. And #1725 adds REAL wired VRAM admission via a parallel VramReservationManager (app.state.vram_reservation, used in routes/models.py) that bypasses the arbiter entirely. As composed the epic ships TWO unintegrated VRAM systems - an inert arbiter and a live model-load reservation that does not know the arbiter exists. We need ONE authority: route a real admission through GpuArbiter.submit_gpu and reconcile #1725 into it, or consciously drop the arbiter and keep only #1725 - but we cannot ship "the arbiter epic" with the arbiter dormant.

2. LATENT EVICTION TOCTOU in #1718 (fix before submit_gpu is ever wired). _evict_task frees the reservation (:467) BEFORE cancelling the evicted task (:471), and cancel() only schedules cooperative cancellation. So evict-to-make-room (_drain_queue:515-517) can free A's 4GB, pass B's _reserve_and_check against the now-free 4GB, and start loading B while A's model is still physically resident -> concurrent load -> VRAM over-subscription = the Xid-62 crash the arbiter exists to prevent. Masked today only because the arbiter is dormant, so it activates the moment submit_gpu is wired. Fix: free the reservation only after cancellation completes and physical VRAM is reclaimed (await the cancelled task, or probe before re-admitting).

3. #1725 unresolved kilo CRITICAL: routes/models.py:377 reserves min_ram_mb (host RAM) as VRAM -> over-reserves -> false 503 denials; use a VRAM estimate. Plus warnings: :430 reservation leaks if start_installer_task raises before the bg task runs; vram_reservation.py:90 stats()/available_vram() probe outside the lock.

4. #1726 unresolved kilo WARNING: manager.py:274 self._workers.pop(name) runs AFTER the _lease_lock block exits, so any future await between release and pop reopens the exact leak it fixes. Move the pop inside `async with self._lease_lock`.

5. #1727 - please close it. It is a competing single-commit rewrite of gpu_arbiter.py + drain_worker that conflicts with #1718 (both create the same file) and is strictly worse: drops the app.py wiring (set_gpu_arbiter never called), omits the #1690 worker-update routes, and carries CRITICAL :421 (worker.free_vram_mb is None for non-NVIDIA/unreported workers -> TypeError on every admission), :585 (no general except -> one bad drain kills the queue processor), and :190 (stop() never resolves pending _arbiter_futures -> hung submit_gpu callers).

6. Minor on #1718: routes/cluster.py:1185 resp.json() lacks raise_for_status, so a non-2xx worker response with a JSON error body reports status:"updating" (silent success).

MERGE PLAN (I will land it as a coherent package, not piecemeal):
- Close the superseded: #1683, #1690, #1705, #1706, #1707 (folded into #1718) and #1727 (competing/worse).
- Land #1718 after the eviction-ordering fix (2) + the resp.json nit (6).
- Land #1726 after moving the pop inside the lock (4).
- Land #1725 after the min_ram_mb CRITICAL (3).
- Gate-closer: wire a real admission through submit_gpu and reconcile with #1725 so there is one VRAM authority (1) - this is what actually closes #185.

Once 1-6 are green together I will merge #1718 + #1726 + #1725 as the package. Solid work clearing the mechanical blockers; the rest is integration + the eviction-ordering safety.

hognek:
Agreed on converging everything on #1718 — the composed-state analysis is exactly right, thanks for the thorough SET pass.

**On item 1 (the #185 gate): let's do it properly.** Wire a real admission path through `GpuArbiter.submit_gpu` and make the arbiter the single VRAM authority, rather than shipping it dormant beside #1725's parallel reservation. Concretely: the model-load path in `routes/models.py` admits through `submit_gpu`, and #1725's `VramReservationManager` becomes the arbiter's internal accounting rather than a second gate. That closes #185 for real and gives us one authority instead of two.

One shape question before I build the reconciliation, so I match your intended API the first time: should the **arbiter be the front door** with `VramReservationManager` demoted to its reservation bookkeeping, or keep `VramReservationManager` as the public reserve API with `submit_gpu` calling through it? Either works — I'll follow your preference for the long-term surface.

I'll land **item 2's eviction-ordering fix in the same change**, since wiring `submit_gpu` is exactly what arms that TOCTOU: free the reservation only after the evicted task's cancellation completes and its VRAM is physically reclaimed, not at `_evict_task:467` ahead of `cancel()` at `:471`. And I'll fold the `routes/cluster.py:1185` `raise_for_status` nit (item 6) into #1718.

Items **3 (#1725** min_ram_mb→VRAM estimate + the `:430` leak + `vram_reservation.py:90` probe-under-lock**)** and **4 (#1726** pop inside `_lease_lock`**)** I'll handle on their own PRs. **Closing #1727** now as superseded.

Target sequence, landing as the one package you described: #1718 (eviction-ordering + resp.json + submit_gpu wiring) → #1726 → #1725 reconciled into the arbiter. Say the word if you'd rather I split the `submit_gpu` wiring into a follow-up PR on top of the mechanical #1718 fixes.

jaylfc:
Answering the shape question so you can build the reconciliation once: make the arbiter the front door. `GpuArbiter.submit_gpu` is the single public admission API, and `VramReservationManager` becomes the arbiter's internal reservation bookkeeping, not a second public gate.

The whole point of #185 is one VRAM authority. If `VramReservationManager.reserve` stays a public entry point, callers can still take VRAM without passing through admission (queue plus eviction plus accounting), which is exactly the parallel-path bypass the SET review flagged on #1725. So the model-load path in `routes/models.py` admits through `submit_gpu`, and everything that consumes VRAM is subject to the same gate.

On sequencing: your target is good and I will take it as the package. Keeping item 2's eviction-ordering fix in the same change as the `submit_gpu` wiring is correct for the reason you gave, the wiring is what makes the eviction path reachable, so they have to be consistent in one diff. No need to split it out.

The one hard constraint from #185: nothing merges piecemeal. #1718 (mechanical fixes plus eviction-ordering plus resp.json plus submit_gpu wiring) plus #1726 (pop inside `_lease_lock`) plus #1725 (min_ram_mb to real VRAM estimate, the `:430` leak, the probe-under-lock, reconciled into the arbiter's internal accounting) land together as one consistent package after I re-review the composed state end to end. Ping me when the three are in their final shape and I will do the SET re-review before anything merges. Thanks for closing #1727 and for driving this cleanly.

jaylfc:
Hi @hognek — I did a full cross-PR pass over the arbiter epic (#1705/#1706/#1707/#1718/#1759 + #1690/#1726) to untangle it before merging, since they all rewrite `scheduler/gpu_arbiter.py` + `app.py`. Summary and a concrete plan:

**Supersession (verified by commit ancestry, not titles):** #1705 ⊂ #1706 ⊂ #1707 ⊂ **#1718** is a strict git-ancestor chain — the first three are older snapshots of this same branch, so I am closing them as superseded (pointing here). **#1759** branches off separately into a competing design (a second class also named `VramReservationManager` inside the arbiter) and Kilo flags it CRITICAL: it constructs `GpuArbiter(...)` with no `vram_probe` (admission is a fail-open no-op) and `cluster_manager._gpu_arbiter` is never read (eviction is a prod no-op). I am closing #1759 too — this PR (#1718) already does both correctly.

**The keepers:** #1690 (worker auto-update) merges clean and is the right home for that feature; #1726 (release leases on unregister) is a real fix (one trivial test conflict). Both land on their own.

**The one architectural blocker on this PR:** dev already merged the epic's first reservation layer — `tinyagentos/vram_reservation.py::VramReservationManager` (#1725, hotfixed by #1767), instantiated in `app.py` as `app.state.vram_reservation`. #1718 introduces a *parallel* reservation system (`_reserved_vram_mb` + `_pending_reservations`) that does not share state with it, so we would ship two VRAM authorities that double-count. Could you rework this PR to **build on the merged `VramReservationManager`** instead, and split the worker-auto-update commits back out to #1690 so this is arbiter-only?

**Correctness items to fold in the rework (all in `scheduler/gpu_arbiter.py` unless noted):**
1. **HIGH — fail-OPEN on probe failure** (`_check_admission`): `_probe_nvidia_vram` collapses every error (nvidia-smi missing/timeout/parse) to `(0,0)`, which admission reads as "no GPU → admit". A GPU box whose probe transiently fails then admits unlimited concurrent loads — the exact Xid-62 crash the arbiter prevents. Must fail *closed* when a GPU is expected (the merged manager already distinguishes "no probe" from "known-empty").
2. **HIGH — blocking `nvidia-smi` on the event loop, under `_reservation_lock`**: two `subprocess.run(timeout=5)` calls (~10s) block the whole asyncio loop while holding the lock. The merged manager fixed this with `await asyncio.to_thread(...)`; please reuse that path.
3. **MEDIUM — eviction preempts EQUAL priority** (`evict_lowest_priority`): guard is `pri < min_priority`; with "lower int wins" that evicts a *running* same-priority task for a queued one (thrash). Should be `pri <= min_priority: continue` (only evict strictly-lower).
4. **MEDIUM — VRAM released before physical unload** (`_evict_task`): accounting is freed, then the task is `cancel()`ed (cooperative), but the model stays resident until it actually unloads — a new admission sees phantom capacity. Tie the release to real unload (again what the merged manager is built to track).

Full write-up is long; happy to drop the rest (D-6/7 queued-caller-on-stop, D-8 double-count, D-9 `worker.status` write outside `_lease_lock` in the drain paths) inline if useful. No blocking rush — once this is arbiter-only + on the merged reservation manager + 1–4 are in, I will merge. Really solid work overall; the eviction/drain design is right, it just needs to converge on the one reservation authority.

jaylfc:
Hi @hognek, consolidating what is left to close the arbiter set as one package, since #1726 rebased cleanly and #1690 now gates the destructive routes with _require_admin (that closes the earlier auth CRITICAL, thanks). Remaining before I do the SET re-review and merge #1718 + #1726 + #1725 together:

1. Gate-closer (this is what actually closes #185). submit_gpu still needs a real production caller: route the model-load admission in routes/models.py through GpuArbiter.submit_gpu, and make VramReservationManager the arbiter internal bookkeeping rather than a second public entry point, so there is ONE VRAM authority. Right now the arbiter ships inert and #1725 is a parallel path.

2. Eviction ordering (#1718). Free the reservation only after the evicted task physically unloads, not before cancel(). As composed, evict-to-make-room can pass the next admission against not-yet-reclaimed VRAM and over-subscribe once submit_gpu is wired.

3. #1725 items: fail CLOSED on probe failure (a transient nvidia-smi failure must not read as no GPU then admit), run nvidia-smi via asyncio.to_thread off the event loop, and do not preempt EQUAL priority in evict_lowest_priority (use pri <= min: continue).

4. #1726: move self._workers.pop(name) inside the async with self._lease_lock block so a future await between release and pop cannot reopen the leak.

5. Please close the superseded ones so the set is unambiguous: #1705, #1706, #1707 (strict git-ancestors of #1718) and #1759 (competing design).

Optional, not blocking: rate limiting on the /drain, /cancel-drain, /update routes (already admin-gated).

Ping me when #1718 + #1726 + #1725 are in final shape and I will re-review the composed state end to end. Nothing merges piecemeal. Solid work, the drain and eviction design is right, it just needs to converge on the one reservation authority.

jaylfc:
Hi @hognek, status sync to close this out. Two of the three sibling PRs are now merged to dev:

- #1725 (VRAM guard / atomic check-and-reserve) is in.
- #1726 (release GPU leases on worker unregister) is in as of today, after a code-level re-review: the lease release now mirrors the _monitor_loop offline path exactly and the _workers.pop is inside _lease_lock, so it is atomic. Thanks.

That means the "land together" package is now just this PR landing on top. #1690 is composed into this branch (commit ancestry confirms it), so it rides in with #1718 rather than merging separately.

Remaining before I do the SET re-review and merge, unchanged from what we agreed on the 8th/11th plus two items a fresh review surfaced:

1. Gate-closer (#185), the real blocker. submit_gpu still needs a production caller: route the model-load admission in routes/models.py through GpuArbiter.submit_gpu, and demote VramReservationManager to the arbiter's internal bookkeeping so there is ONE VRAM authority. As it stands the arbiter ships inert beside #1725's now-merged path.
2. Eviction ordering. Free the reservation only after the evicted task physically unloads, not before cancel(). As composed, evict-to-make-room can re-admit against not-yet-reclaimed VRAM and start a second load while the first is resident (the Xid 62 the arbiter exists to prevent). Keep this in the same change as the submit_gpu wiring.
3. Rebase onto current dev. Only app.py conflicts now, and that conflict IS the #185 reconciliation: dev wired app.state.vram_reservation in the same lifespan block where this PR starts/stops the arbiter, so it needs the one-authority resolution rather than a both-hunks merge.

Two smaller items from the fresh pass:

4. heartbeat() re-onlines a draining worker. heartbeat() sets worker.status = "online" unconditionally with no draining guard, so a worker that is mid graceful-drain flips back to online on its next heartbeat and re-enters routing before its leases finish, which defeats the drain. A one-line guard fixes it: only set "online" when status is not "draining".
5. Nits: fail closed on the VRAM probe (do not admit when nvidia-smi fails), run the probe via asyncio.to_thread so the ~10s subprocess does not block the loop under _reservation_lock, use <= in the equal-priority eviction guard so equal priority does not preempt, and raise_for_status on the worker response in routes/cluster.py /update before reading .json().

If you still have bandwidth this is yours and I will do the SET re-review the moment it is pushed. If you would rather I take the submit_gpu wiring + rebase off your plate, say the word and I will pick it up and credit the branch to you. Either way, thanks for the careful work on this one.

jaylfc:
Picked this up to get it across the line, @hognek. Opened #1859 which continues this branch (your commits are preserved) rebased onto current dev, plus the #185 gate-closer we agreed: the arbiter now reserves against the shared VramReservationManager (one authority, no double-count) with reserve_vram/release_vram as its front door, and the eviction-ordering (Xid-62) + equal-priority-eviction + the heartbeat-draining-guard (#890) fixes are in with regression tests. #1725 and #1726 already merged, so #1859 is the last piece. Closing this in favour of #1859 to keep one convergence target; thanks for the careful arbiter work, it made the takeover straightforward.

PR #1724, opened 2026-07-07
fix(searxng): enable JSON output format via settings.yml bind mount, not env vars

### LINKED ISSUE #969: SearXNG catalog install: enable JSON output format by default (agent use)

hognek:
Fixes #969.

SearXNG only reads SEARXNG_SETTINGS_PATH and SEARXNG_DISABLE_ETC_SETTINGS — there is no generic SEARXNG_* nested override (the __ pattern is Home Assistant, not SearXNG). The previous env-var approach did nothing.

## What this does

1. **Adds `config_files` support to `DockerInstaller`** — a declarative way for manifests to seed config files into the app directory before compose generation. Supports `{secret_key}` templating for per-install random secrets.

2. **Seeds a real `settings.yml`** with:
   - `use_default_settings: true`
   - `search.formats: [html, json]` — JSON format enabled for agent queries
   - `server.secret_key` — per-install generated (64-hex-char random)

3. **Bind-mounts `./settings.yml:/etc/searxng/settings.yml:ro`** — overrides the default settings.yml inside the named config volume.

4. **Removes the non-functional `SEARXNG_SEARCH__FORMATS__*` env vars** from the old approach.

## Upgrade handling

Existing installs get the fix automatically: re-running the install flow writes the settings.yml and adds the bind mount to compose. The named volume's default settings.yml is overridden by the bind mount.

## Tests

- 53/53 pass: `pytest tests/test_installers.py tests/routes/test_store_install_v2.py tests/test_routes_store.py`
- 2 new tests: `test_write_config_files_creates_files_with_substitution`, `test_write_config_files_noop_when_no_config_files`

## Summary by CodeRabbit

* **New Features**
  * Added support for app-specific configuration files during installation.
  * Configuration can now include a generated secret placeholder and default settings.
  * Docker-based installs now write config files before deployment so local file mounts work reliably.

* **Bug Fixes**
  * Improved handling of configuration file paths to prevent invalid or unsafe locations.
  * Kept the generated secret consistent across reinstalls for stable configuration behavior.

### DIFF
--- a/app-catalog/services/searxng/manifest.yaml
+++ b/app-catalog/services/searxng/manifest.yaml
@@ -16,9 +16,25 @@ install:
   image: searxng/searxng:latest
   volumes:
     - config:/etc/searxng
+    - ./settings.yml:/etc/searxng/settings.yml:ro
   # 8080 is the container-internal port; the host port is allocated from the
   # managed pool (30000-40000) by the installer — never bound to 8080 on the host.
   ports: [8080]
+  config_files:
+    - path: settings.yml
+      content: |
+        # SearXNG settings — JSON format is enabled so agents can query
+        # the engine programmatically (?format=json).
+        use_default_settings: true
+
+        search:
+          formats:
+            - html
+            - json
+
+        server:
+          secret_key: "{secret_key}"
+          bind_address: "0.0.0.0"
 
 hardware_tiers:
   arm-npu-16gb: full
--- a/app-catalog/services/searxng/manifest.yaml
+++ b/app-catalog/services/searxng/manifest.yaml
@@ -15,7 +15,6 @@ install:
   method: docker
   image: searxng/searxng:latest
   volumes:
-    - config:/etc/searxng
     - ./settings.yml:/etc/searxng/settings.yml:ro
   # 8080 is the container-internal port; the host port is allocated from the
   # managed pool (30000-40000) by the installer — never bound to 8080 on the host.
--- a/tests/test_installers.py
+++ b/tests/test_installers.py
@@ -98,6 +98,35 @@ class TestDockerInstaller:
         assert "volumes" not in compose  # no named volumes → no top-level block
         assert host_port is None
 
+    def test_write_config_files_creates_files_with_substitution(self, tmp_path):
+        installer = DockerInstaller(apps_dir=tmp_path)
+        installer._write_config_files("searxng", {
+            "config_files": [
+                {"path": "settings.yml", "content": 'secret_key: "{secret_key}"'},
+                {"path": "sub/deep.yml", "content": "key: static"},
+            ]
+        })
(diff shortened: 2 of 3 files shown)


hognek:
Addressed 5 Kilo review findings (5ab57de):

**CRITICAL — path traversal** (docker_installer.py:42): `config_files[*].path` was written verbatim to `app_dir/path` with no traversal protection. Now validates each path:
- Rejects absolute paths (`/etc/passwd`)
- Rejects `..` components (`../escape.yml`)
- Resolves + confirms the result stays within `app_dir` (catches symlink escapes too)

**WARNING — entry validation** (docker_installer.py:38): Missing `path`/`content` keys raised bare KeyError/TypeError deep in the loop. Now validates each entry is a dict with both keys, with clear ValueError messages referencing the index.

**WARNING — secret_key rotation** (docker_installer.py:36): `secrets.token_hex(32)` regenerated on every `install()` call, silently rotating SearXNG's secret_key on re-install. Now persists the key in `<app_dir>/.secret_key` — generates fresh only when absent, reuses on subsequent calls.

**SUGGESTION — leftover volume** (manifest.yaml:19): Removed the named volume `config:/etc/searxng` — the bind mount `./settings.yml:/etc/searxng/settings.yml:ro` is the only config surface needed. The named volume was created but never used independently.

**SUGGESTION — brittle parse** (tests/test_installers.py:118): Replaced `content[len('secret_key: "'):-1]` with `content.split('"')[1]`.

6 new tests covering all validation paths + persistence. 29 installer+catalog tests pass.

Issue #894, opened 2026-06-14
Scheduler Phase 2: VRAM-accounted admission + queue + eviction (the GPU arbiter both agents use)


jaylfc:
## Why now (Jay 2026-06-14)
@taOS FLUX image-gen OOMd the shared 3060 because @taOSmd had ~9.4GB of Ollama models loaded. The A2A GPU-lease protocol (#893) is the stop-gap; the proper fix is the taOS resource queue both consumers submit to instead of negotiating by hand.

## Current state (scaffolding exists)
- `scheduling/resource_manager.py`: discovers hardware, **queries Ollama for loaded models**, tracks VRAM/RAM pressure, adjusts job-queue concurrency. Awareness only, no enforcement.
- `scheduler/scheduler.py` + `scheduler/resource.py`: Phase 1, deliberately minimal -- synchronous routing, per-resource concurrency semaphore, first-resource-that-passes-admission runs. Priority queues, aging, preemption, max_wait_ms are Phase 2 and intentionally absent (per the module docstrings).
- `scheduler/core_aware_scheduler.py` (#172): load_with_core_awareness(), not yet wired into the dispatch path.

## The gap to build (Phase 2)
1. **Per-device VRAM budget** in resource.py: track total/used/free per accelerator (free = real-time, already accounts for external Ollama via the resource_manager read, so @taOSmd co-tenancy is visible).
2. **Admission control**: a GPU job (load model / generate / inference) is admitted only if est_vram <= free_vram; otherwise it queues.
3. **Wait-queue with aging**: queue blocked jobs, FIFO + aging to avoid starvation; max_wait_ms enforcement.
4. **Eviction**: before rejecting, evict idle keep-alive models (LifecycleManager) to make room.
5. **Single arbiter for co-tenants**: route both taOS backends AND (ideally) taOSmd's Ollama loads through admission. First step: taOS-side admission respects real-time free VRAM and queues/evicts; full mutual arbitration has taOSmd defer GPU loads to the same manager.
6. Wire TaskRouter.generate_image + core-aware loading (#172) through the admission path.

## Acceptance
Two consumers (image-gen + chat models) never co-load past a GPU's VRAM; over-budget work queues and runs when it fits (after eviction), with no OOM. The describe_image_capabilities tool reports live VRAM + queue depth. Subsumes the #893 A2A stop-gap. Brainstorm -> spec -> Phase 2 impl; do not block the current storybook demo on it.

jaylfc:
[...]

PR #1725, opened 2026-07-07
feat(models): atomic VRAM check-and-reserve before model load (TOCTOU fix, #1706)

hognek:
## Summary

Implements atomic VRAM reservation to prevent TOCTOU races during concurrent model loads.

**Problem:** Two concurrent model pulls (ollama/rkllama) both check free VRAM via nvidia-smi, both see enough, both start loading — causing OOM or driver crashes.

**Solution:** `VramReservationManager` — an `asyncio.Lock`-protected reservation system that atomically checks free VRAM and reserves the required amount. Concurrent callers see the in-flight reservation and back off with HTTP 503.

### New files
- `tinyagentos/vram_reservation.py` — atomic reserve/release manager with real-time nvidia-smi probe
- `tests/test_vram_reservation.py` — 15 tests covering reserve/release, concurrent safety, zero-VRAM, stats, and real-probe smoke test

### Modified files
- `tinyagentos/app.py` — wires `VramReservationManager` as `app.state.vram_reservation` in the FastAPI lifespan
- `tinyagentos/routes/models.py` — integrates reservation into:
  - `POST /api/models/pull` — reserve before ollama pull, release in finally
  - `POST /api/models/download` (rkllama path) — reserve before installer, release in background task finally
  - `GET /api/models/vram-reservations` — new monitoring endpoint

## Design

The manager follows the same pattern as the GPU arbiter's TOCTOU fix (#894 review feedback #2): an `asyncio.Lock` serializes the check-and-reserve sequence so two concurrent `reserve()` calls never both see the same free VRAM. In-flight reservations are subtracted from the probe result so the effective free VRAM shrinks for subsequent callers.

Zero-VRAM reservations (`vram_mb <= 0`) always succeed as lightweight markers (no lock, no probe).

## Testing

```
pytest tests/test_vram_reservation.py tests/test_routes_models.py tests/test_config.py tests/test_cluster.py tests/test_install_progress.py tests/test_catalog_sync.py -v
# 125 passed in 39s
```

No regressions in existing model, config, cluster, installer, or catalog tests.

Issue: #1706

## Summary by CodeRabbit

* **New Features**
  * Added VRAM-aware model loading with atomic reservation checks to help prevent overcommitting GPU memory.
  * Added a new endpoint to view current VRAM reservation status and availability.
  * App startup now initializes VRAM tracking for use during model operations.

* **Bug Fixes**
  * Model downloads and pulls now fail cleanly when there isn’t enough VRAM, then release reservations after completion.

* **Tests**
  * Added coverage for reservation success/failure, release behavior, concurrency safety, and monitoring stats.

### DIFF
--- a/tests/test_vram_reservation.py
+++ b/tests/test_vram_reservation.py
@@ -1,0 +1,238 @@
+"""Tests for atomic VRAM reservation (taOS #1706 TOCTOU fix).
+
+Verifies that:
+- Concurrent reserve calls cannot over-commit VRAM.
+- reserve() returns None when there's not enough VRAM.
+- release() correctly returns VRAM to the pool.
+- release() is idempotent.
+- Zero-VRAM reservations always succeed.
+- stats() reflects the current state.
+"""
+
+import asyncio
+
+import pytest
+
+from tinyagentos.vram_reservation import VramReservationManager
+
+
+# ── test helpers ────────────────────────────────────────────────────
+
+
+def _manager_with_probe(free_mb: int, total_mb: int) -> VramReservationManager:
+    """Return a manager whose _probe_vram returns fixed values."""
+    mgr = VramReservationManager()
+
+    def _fake_probe() -> tuple[int, int]:
+        return free_mb, total_mb
+
+    mgr._probe_vram = staticmethod(_fake_probe)  # type: ignore[method-assign]
+    return mgr
+
+
+# ── basic reserve / release ─────────────────────────────────────────
+
+
+class TestReserveRelease:
+    """Happy-path reserve and release."""
+
+    @pytest.mark.asyncio
+    async def test_reserve_when_vram_available(self):
+        """reserve() succeeds when enough VRAM is free."""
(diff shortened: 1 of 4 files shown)


jaylfc:
Confirming the direction (full detail on #894): this is the VRAM guard we ship first, as the single authority on the model-load path. Kilo is clean and the atomic check-and-reserve is exactly what prevents the concurrent-load over-subscription.

Two things before it merges:
1. Please rebase on current `dev` so the required checks (test 3.12/3.13, doc-gate, lint, spa-build) run against the latest tree; it has been sitting since the 7th and GitHub reports the merge state as unknown.
2. The `min_ram_mb` VRAM estimate is fine for v1 (it over-reserves, so it fails safe), but leave a `# TODO` to replace it with a real per-model VRAM estimate, since on CUDA GGUF the host-RAM floor will over-reserve and cause occasional false 503s.

Once it is rebased and green I will merge it. The `GpuArbiter` queue/eviction is staged separately as the follow-up on #894.

PR #1705, opened 2026-07-07
fix(scheduler): close TOCTOU race in GPU admission with in-flight VRAM reservation

hognek:
## Summary

Closes the TOCTOU (time-of-check to time-of-use) race in `_check_admission` identified in [PR #1683 review feedback #2](https://github.com/jaylfc/tinyagentos/pull/1683#issuecomment-4897343798).

## Problem

`_check_admission` reads live `nvidia-smi` free VRAM with no reservation mechanism. Two concurrent `submit_gpu` calls can both pass admission before either model loads — the exact concurrent-load crash (NVIDIA Xid 62) the arbiter exists to prevent.

## Fix

Adds in-flight VRAM reservation tracking:
- `_reserved_vram_mb`: counter of VRAM promised to tasks whose models are still loading
- `_pending_reservations`: dict mapping task_id → reserved VRAM for idempotent release
- `_reservation_lock`: `asyncio.Lock` to make check+reserve atomic

New method `_reserve_and_check(task_id, required_vram_mb)` atomically:
1. Acquires `_reservation_lock`
2. Calls `_check_admission` (which now subtracts `_reserved_vram_mb` from the probe)
3. If admitted, immediately adds to `_reserved_vram_mb`

Reservations are released in:
- `_run_gpu_task` finally block (normal completion, error, cancellation)
- `_evict_task` (preemption)

## Tests

- **9 existing** `test_gpu_arbiter_894.py` tests continue to pass
- **7 new** `test_gpu_arbiter_toctou.py` tests:
  - Concurrent `_reserve_and_check` only admits one of two calls exceeding capacity
  - `submit_gpu` reserves and releases correctly
  - Eviction releases reservation
  - `_check_admission` subtracts reserved VRAM from probe
  - Atomic reserve-and-check with lock
  - Idempotent release
  - Reservation fields in `stats()`

All 168 tests pass (25 arbiter + 143 cluster/worker).

## Summary by CodeRabbit

* **New Features**
  * Added GPU workload admission control with VRAM-aware scheduling, queueing, and eviction support.
  * Expanded scheduler discovery so GPU resources can be recognized and surfaced at startup.
  * GPU scheduler availability is now initialized automatically during app startup, with fallback behavior if setup fails.
* **Bug Fixes**
  * Prevented GPU overcommit by accounting for in-flight VRAM reservations during admission.
  * Improved eviction behavior so running GPU work is actually stopped and resources are released correctly.

PR #1726, opened 2026-07-07
fix(cluster): release GPU leases on worker unregister (fixes #1705)

hognek:
## Summary

Fixes taOS #1705: reservation leak — GPU leases were not released when a worker is explicitly unregistered.

### The bug

`ClusterManager.unregister_worker()` simply popped the worker from `_workers` but left all its active GPU leases in `_leases`. This meant the resource slots remained locked indefinitely — new claims on those resources would get 409 "already leased" even though the worker was gone.

The heartbeat-timeout path in `_monitor_loop` already handled the crash/disconnect case correctly (marks worker offline → releases leases). But the **explicit disconnect path** (`DELETE /api/cluster/workers/{name}`) leaked leases.

### The fix

- **`manager.py`**: `unregister_worker()` made async, acquires `_lease_lock`, releases all leases for the worker's resources (exact name match via `_parse_resource_id`), then pops the worker. Logs count of leases released.
- **`routes/cluster.py`**: Route handler now `await`s the async method.

### Tests added (6 new)

| Test | What it covers |
|------|---------------|
| `test_unregister_releases_all_leases` | Unregistering releases all active leases |
| `test_unregister_unknown_worker_noop` | Unknown worker → False, no lease side effects |
| `test_unregister_one_worker_spares_another` | Unregistering A doesn't touch B's leases |
| `test_reclaim_after_unregister` | New claim succeeds on re-registered worker |
| `test_unregister_releases_leases_atomically` | `_lease_lock` serializes unregister with concurrent claims |
| `test_unregister_worker_releases_leases` | HTTP integration: DELETE /workers/{name} releases leases |

### Test results

```
tests/test_leases.py ........... 34 passed
tests/test_cluster.py ......... 16 passed
Total: 50 passed
```

## Summary by CodeRabbit

* **New Features**
  * Removing a worker now automatically releases all GPU leases assigned to it.
  * Workers can be re-added and claim GPU leases again after removal.

* **Bug Fixes**
  * Prevented stale leases from remaining after worker removal.
  * Preserved leases belonging to other workers during removal.
  * Improved handling of repeated removal requests with clear success or not-found responses.

### DIFF
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -43,9 +43,9 @@ class TestClusterManager:
     async def test_unregister_worker(self):
         mgr = ClusterManager()
         await mgr.register_worker(_make_worker("gpu-box"))
-        assert mgr.unregister_worker("gpu-box") is True
+        assert await mgr.unregister_worker("gpu-box") is True
         assert mgr.get_workers() == []
-        assert mgr.unregister_worker("gpu-box") is False
+        assert await mgr.unregister_worker("gpu-box") is False
 
     async def test_heartbeat_updates_load_and_status(self):
         mgr = ClusterManager()
--- a/tests/test_leases.py
+++ b/tests/test_leases.py
@@ -338,6 +338,52 @@ async def test_claim_rejects_negative_vram(client, app):
     assert resp.status_code == 422
 
 
+# ── taOS #1705: DELETE /worker endpoint leases ─────────────────────────
+
+
+@pytest.mark.asyncio
+async def test_unregister_worker_releases_leases(client, app):
+    """DELETE /api/cluster/workers/{name} releases all GPU leases held
+    by that worker (taOS #1705)."""
+    # Register a worker with enough VRAM
+    key = await _register_worker(
+        client, app, "gpu-node", "http://10.0.0.1:9000",
+        capabilities=["llm-chat"],
+        hardware={"gpu": {"model": "GTX 1080", "vram_mb": 8192}},
+    )
+    await _heartbeat(client, key, "gpu-node", free_vram_mb=8000, used_vram_mb=0)
+
+    # Claim two leases
+    claim1 = await client.post("/api/cluster/leases/claim", headers=_auth(app), json={
+        "resource_id": "gpu-node:gpu-cuda-0",
+        "caller": "test",
+        "ttl_seconds": 60,
+    })
+    assert claim1.status_code == 200
+
+    claim2 = await client.post("/api/cluster/leases/claim", headers=_auth(app), json={
+        "resource_id": "gpu-node:gpu-cuda-1",
+        "caller": "test",
+        "ttl_seconds": 60,
+    })
+    assert claim2.status_code == 200
+
+    # Verify both leases are present
+    list_resp = await client.get("/api/cluster/leases", headers=_auth(app))
+    assert list_resp.json()["count"] == 2
+
+    # Unregister the worker — this must release its leases
+    resp = await client.delete("/api/cluster/workers/gpu-node")
+    assert resp.status_code == 200
+    assert resp.json()["status"] == "removed"
+
+    # Verify all leases are gone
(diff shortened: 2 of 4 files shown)


jaylfc:
Reviewed this in depth, @hognek — the **production code is correct and safe to ship**: `unregister_worker` releases leases under `_lease_lock` (matching the monitor offline path), it is idempotent (unknown worker → False, double-unregister → False, `pop(lid, None)` cannot KeyError), and the DELETE route correctly awaits the now-async call. The 4 ClusterManager unit tests genuinely exercise the path. Nice fix.

One thing blocks the merge, and it is a **merge artifact, not your logic**: since your merge-base, `dev` added auth enforcement on the lease endpoints (`_require_lease_access`, `routes/cluster.py:1029`) and an `_auth(app)` test helper, adding `headers=_auth(app)` to every lease call in `tests/test_leases.py`. That is the textual conflict, and it also means your new integration test `test_unregister_worker_releases_leases` will start returning 401/403 post-merge — it calls `POST /api/cluster/leases/claim` (x2) and `GET /api/cluster/leases` **without** auth headers. CI is green today only because the auth gate lives on dev, not in your branch.

To land it: rebase on `dev`, keep dev's `_auth` helper + header edits, and add `headers=_auth(app)` to the two `leases/claim` POSTs and the `GET /leases` in your new test. Then it is a clean merge and I will take it.

Optional nits (non-blocking, Kilo agrees): the `self._workers.pop(name, None)` runs just after the `async with self._lease_lock` block — safe today (no `await` between release and pop) but you could move it inside the lock to keep it future-proof; and `test_unregister_releases_leases_atomically` proves the call blocks on the lock but does not race a real concurrent claim. Both optional.

hognek:
Rebased onto dev (b8acce7c) with merge conflict resolved.

**Merge artifact fix:** Added `headers=_auth(app)` to all four lease API calls in `test_unregister_worker_releases_leases` (two `POST /leases/claim` + two `GET /leases`). This keeps the test passing post-rebase against dev's auth gate (PR #1711).

**Optional nit applied:** Moved `self._workers.pop(name, None)` inside the `async with self._lease_lock:` block for future-proof atomicity.

**Tests:** All 53 tests in `tests/test_leases.py` + `tests/test_cluster.py` pass locally.

jaylfc:
Thanks Hognek, #1726 itself looks good now: the rebase auth-header fix on the four lease calls and moving the `pop` inside `_lease_lock` are exactly right, and Kilo is down to two non-blocking suggestions (the membership-check readability at manager.py:271 and the test name/strength at test_leases.py:640). Fold those if you like, but they are not blockers.

The reason I am not merging this yet is the epic gate, not #1726: this lands as a package with #1718 (and the merged #1725 VRAM manager), and #1718 is still on its 2026-07-07 commit, so the core arbiter findings are unaddressed there: submit_gpu still has no production caller (the arbiter ships inert), #1718 carries a parallel VRAM reservation that needs reworking onto the merged VramReservationManager rather than double-counting, and the eviction ordering frees the reservation before the task is cancelled and the VRAM is reclaimed. Once #1718 reworks onto the merged manager, wires the submit_gpu gate-closer, and fixes the eviction ordering, I will review #1718 + #1726 + #1725 together and merge the set. #1726 is ready and waiting on that.

jaylfc (approved):
Reviewed at the code level. The fix mirrors the existing _monitor_loop offline-path lease release exactly (same exact-name match so gpu-node does not over-release gpu-node-2, same _lease_lock, idempotent pop). The earlier concern about _workers.pop sitting outside the lock is resolved in the head commit (release + removal are now atomic under _lease_lock). Conflict-free against current dev, and the new tests exercise the #1705 regression end-to-end (DELETE endpoint drives lease count 2 to 0). Remaining bot notes are suggestion/nitpick tier (fold the None pre-check into the lock, rename the atomicity test, optionally extract a shared _release_leases_for_worker_locked helper shared with _monitor_loop) and are fine as a follow-up. Thanks hognek.

PR #1731, opened 2026-07-08
fix: remove --preload argument from systemd ExecStart

### LINKED ISSUE #1730: Regression: install-rknpu.sh generates incompatible rkllama.service after rkllama fix (#1529 resurfaced)
## Environment

- Version: beta.35
- Platform: Orange Pi 5 Pro (RK3588, 16GB)
- OS: Debian 12 (Bookworm) ARM64
- Installation: Fresh clean installation

## Description

While investigating #1710, the previous issue #1529 appears to have resurfaced.

Beta.35 fixed the previous rkllama `server_utils.py` f-string SyntaxError by updating the rkllama version. However, the generated `rkllama.service` from `install-rknpu.sh` is still using an old CLI argument that is no longer supported.

The generated service includes:

```bash
--preload qwen3-embedding-0.6b,qwen3-reranker-0.6b,qmd-query-expansion
```

but the current `rkllama_server` does not recognize `--preload`.

## Error

```text
rkllama_server: error: unrecognized arguments: --preload qwen3-embedding-0.6b,qwen3-reranker-0.6b,qmd-query-expansion
systemd: rkllama.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
```

## Expected Behavior

`install-rknpu.sh` should generate a systemd service compatible with the installed rkllama version.

## Context

The dependency update that fixed #1710 changed the rkllama CLI interface, but the generated systemd unit was not updated accordingly.

The installer completes successfully, but rkllama cannot start.

## Impact

Blocks:

- RKLLM provider availability
- Model inference
- taOS Agent functionality

Suggest: Pulling this again https://github.com/jaylfc/rkllama/pull/1

abhi-0203:
## Summary

Removes the unsupported `--preload` argument from the `ExecStart` line in `install-rknpu.sh`. This fixes a regression where the generated `rkllama.service` fails to start because the current `rkllama_server` no longer recognizes the `--preload` flag.

## Changes

- Removed `--preload qwen3-embedding-0.6b,qwen3-reranker-0.6b,qmd-query-expansion` from the `ExecStart` command in `install_systemd_unit()`
- Updated comments to explain why `--preload` was removed
- Changed log messages from "preloaded models" to "required models" for accuracy
- Added lazy-load mode note to the summary output

## Rationale

The `scripts/systemd/rkllama.service` file already documents why `--preload` was intentionally removed:
1. Frees NPU memory budget when chat models aren't in use
2. Each loaded model can use all 3 RK3588 NPU cores

Models now load lazily on first inference request, which is the correct behavior for the current rkllama version.

## Testing

- Verified the generated systemd unit no longer contains `--preload`
- The fix matches the behavior already documented in `scripts/systemd/rkllama.service`

Closes #1730

## Summary by CodeRabbit

* **New Features**
  * The service now starts in lazy-load mode, loading models on first use instead of at startup.
* **Bug Fixes**
  * Updated verification messaging to more clearly report missing required models.
  * Improved the final status summary to explicitly indicate lazy-load mode.

### DIFF
--- a/scripts/install-rknpu.sh
+++ b/scripts/install-rknpu.sh
@@ -618,9 +618,11 @@ install_systemd_unit() {
     local exec_start
     # This ExecStart line is copy-of-truth from the live orange pi:
     #   rkllama_server --processor rk3588 --port 8080 \
-    #     --models /home/jay/rkllama/models \
-    #     --preload qwen3-embedding-0.6b,qwen3-reranker-0.6b,qmd-query-expansion
-    exec_start="$RKLLAMA_VENV/bin/python $RKLLAMA_VENV/bin/rkllama_server --processor $SOC --port $RKLLAMA_PORT --models $RKLLAMA_MODELS --preload qwen3-embedding-0.6b,qwen3-reranker-0.6b,qmd-query-expansion"
+    #     --models /home/jay/rkllama/models
+    # Note: --preload was removed because the current rkllama server
+    # no longer supports it. Models load lazily on first inference
+    # request (see scripts/systemd/rkllama.service for rationale).
+    exec_start="$RKLLAMA_VENV/bin/python $RKLLAMA_VENV/bin/rkllama_server --processor $SOC --port $RKLLAMA_PORT --models $RKLLAMA_MODELS"
 
     log "installing $unit"
     sudo tee "$unit" >/dev/null <<EOF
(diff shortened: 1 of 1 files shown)


jaylfc:
Thanks for this, and for the clear write-up. You correctly diagnosed the regression and the change is clean.

I went a slightly different route: rather than remove `--preload` (which drops preloading of the embedding / reranker / query-expansion models, so the first memory operation after boot pays the cold-load cost), I restored the `--preload` flag on the pinned rkllama server (jaylfc/rkllama#1) and bumped the installer pin to it in #1733, which keeps those models resident.

So I am closing this as superseded by that approach, but your diagnosis was spot on and got the regression triaged fast. Please do send more; this was a good first contribution.

PR #1759, opened 2026-07-09
feat(scheduler): VramReservationManager — atomic VRAM bookkeeping for GpuArbiter [#1683]

hognek:
Adds VramReservationManager as the arbiter's internal reservation layer with atomic check-and-reserve under asyncio.Lock. Wired into GpuArbiter so the arbiter is the single VRAM authority.

- VramReservationManager: atomic reserve()/release(), sync available() for admission checks, stats()
- GpuArbiter._check_admission reads against reserved VRAM
- _run_gpu_task atomically reserves local VRAM before execution, releases in finally or on eviction
- Cluster lease path unchanged

Tests: 21 new (11 unit + 9 integration + 1 export), 59/59 related tests pass.

Task: t_19db5cd3 | Prior: #1718 (drain/auth/eviction — MERGEABLE)

## Summary by CodeRabbit

* **New Features**
  * Added GPU-aware workload scheduling with VRAM tracking, queueing, and priority-based eviction.
  * Scheduler startup now includes GPU resource discovery and GPU arbiter initialization when available.
  * GPU-related components are now available from the scheduler’s public imports.

* **Bug Fixes**
  * Improved task cancellation and cleanup during GPU eviction.
  * Fixed VRAM reservation handling so running and queued GPU tasks release memory correctly and respect available capacity.


jaylfc:
Thanks for pushing on the VRAM bookkeeping. I did a full code-level read of this against the epic gate (#185) and the blockers from the #1718 set review. The mechanical class itself is reasonable, but as the gate-closer it does not land yet, and there is one bug that CI is hiding. Concrete findings, most important first:

**1. The epic gate is still open.** `submit_gpu` has zero production callers. `app.py` constructs the arbiter, calls `start()` (spins the drain loop), and stashes it on `app.state.gpu_arbiter`, but nothing admits work through it. `routes/models.py` is untouched, and `discovery.py` registers the CUDA resource on the scheduler, bypassing the arbiter. So `_running` stays empty in prod and eviction is a no-op, exactly the inert-ships problem #1718 had. This is what actually closes #185: route the real model-load admission (and cluster dispatch) through `submit_gpu`.

**2. Two VRAM authorities, not one.** This adds a `VramReservationManager` inside the arbiter while #1725 adds a separate same-named manager wired into `routes/models.py`. If both merge we have two independent VRAM accountants that do not know about each other. We need ONE: fold #1725's live model-load reservation into the arbiter path rather than run them in parallel.

**3. HIGH, CI is masking it: async lease APIs called without await.** In `_run_gpu_task` (and `_evict_task`) the cluster path does `lease = self._cluster_manager.claim_lease(...)` / `self._cluster_manager.release_lease(...)`, but `ClusterManager.claim_lease` and `release_lease` are `async def` (manager.py). Called without `await`, `claim_lease` returns a coroutine (never None), so `lease.lease_id` raises AttributeError, and the release coroutine is never awaited (lease leak + RuntimeWarning). Every real cluster GPU task would fail. It is green only because `FakeClusterManager` in the tests is synchronous. Please `await` these (or make them sync) and switch the test double to an async fake so CI actually exercises the real contract.

**4. CRITICAL (kilo, confirmed): no vram_probe wired.** The arbiter is built without a `vram_probe`, so it falls to `_default_vram_probe` and reads `(0, 0)`. Local VRAM admission/reservation is therefore inert. `discovery.py` already has `_probe_nvidia_vram`; pass it into the arbiter.

**5. MEDIUM: `free_vram_mb is None` crash in admission.** `_check_admission`'s cluster branch does `worker.free_vram_mb - worker_leases` with no None guard; `WorkerInfo.free_vram_mb` defaults to None for a worker that never reported VRAM, so that is a TypeError on admission. `claim_lease` already guards this; admission should too.

**6. MEDIUM: `_running` / `_running_tasks` mutated and iterated outside `_running_lock`.** `_run_gpu_task` mutates them under the lock, but `_evict_task` pops them lock-free and `evict_lowest_priority` iterates `_running.items()` lock-free. Concurrent completion during eviction risks split-brain between the two dicts or a "dictionary changed size during iteration" error. Take `_running_lock` in the eviction paths.

**7. Eviction-ordering TOCTOU (blocker 2) still open.** `_evict_task` releases the reservation via a fire-and-forget `asyncio.create_task(... .release(...))` and then cancels, so it frees VRAM before the evicted model is physically gone. Harmless only while the arbiter is dormant; it activates the moment #1 is wired. Free the reservation only after cancellation completes / VRAM is reclaimed.

Net: hold this with the rest of the arbiter set rather than merge piecemeal. The unblocking sequence is (3) await the lease APIs + async test double, (4) wire the real vram_probe, then the big one (1)+(2): one VRAM authority routed through `submit_gpu`. Once a real admission flows through the arbiter and the set is consistent I will land #1718 + #1725 + this as the package. Good progress on the accounting core.

jaylfc:
Staging note (direction on #894): we are shipping #1725's model-load VRAM guard first as the single authority, and staging the arbiter (this PR + #1718) as the follow-up that closes #894. So this is held, not rejected. The unblock sequence from my review above stands: await the async lease APIs (with a matching async test double), wire the real vram_probe, then the big one, route a real admission through submit_gpu and reconcile with #1725 so there is one VRAM authority. That last step is what actually closes the epic.

jaylfc:
Closing as superseded by the merged #1725 (`vram_reservation.VramReservationManager`) + #1718. This branch is a competing design that Kilo flagged CRITICAL: `GpuArbiter` is built without a `vram_probe` (admission is a fail-open no-op) and `cluster_manager._gpu_arbiter` is never read (eviction is a prod no-op). #1718 does both correctly, so the epic converges there. Full reasoning in the #1718 plan. Thanks @hognek — reopen if you want to keep any piece.

PR #1762, opened 2026-07-09
feat(cluster): advertise available models from Skald sidecar manifest

hognek:
Task: t_634725d4. Completes end-to-end wiring for Skald PR #361 — the controller's cluster view now shows both loaded and available models per worker.

## What

Adds the TAOS-side half of Skald load-on-demand routing:
1. **`tinyagentos/worker/skald_manifest.py`** — new parser that reads `/etc/taos/skald-models.json` (env-configurable via `TAOS_SKALD_MANIFEST`) to enumerate which models a worker CAN load per the GPU catalog
2. **`WorkerInfo.available_models`** — new `list[dict]` field on the worker protocol dataclass
3. **`ClusterManager.heartbeat()`** — derives available_models from backend entries, deduplicating model_ids across backends
4. **`WorkerAgent.detect_backends()`** — enriches each backend with available models from the sidecar manifest, tagging each as 'loaded' or 'available'

## Tests

- **74 tests pass, 0 regressions:**
  - 14 new manifest parser tests (parsing, env var, edge cases)
  - 5 new WorkerInfo field tests (defaults, serialization, roundtrip)
  - 5 new heartbeat derivation tests (derivation, dedup, additive, empty, preserve)
  - 18 existing kv_quant + VRAM protocol tests — zero regressions
  - 32 existing worker agent tests — zero regressions

## Changes (minimal, additive)

- `worker_protocol.py`: +1 line (available_models field)
- `manager.py`: +12 lines (derivation block in heartbeat)
- `agent.py`: +36 lines (enrichment after backend probe)
- `skald_manifest.py`: new file (49 lines)
- Tests: 234 new lines across 2 test files

No existing code modified — purely additive. Does not touch pairing, lease management, VRAM fields, or any other subsystem.

## Summary by CodeRabbit

* **New Features**
  * Workers can now report a list of available models from a local manifest, including whether each model is already loaded or simply available.
  * Cluster heartbeat updates now carry forward and merge available model information from backend status reports.
  * Added support for reading a local worker manifest file, with configurable path handling.

* **Bug Fixes**
  * Duplicate model entries are now removed when combining backend-reported model availability.
  * Missing manifest files no longer cause worker startup/reporting to fail.


jaylfc:
Thanks, picked this up. The diff is clean (no workflow/secret/dependency changes), Kilo is green, and it fits the existing Skald prior art in the repo. I have approved the fork CI so the required checks run. Since this deepens the core-to-Skald coupling (a new skald_manifest module plus a WorkerInfo protocol field), I am confirming scope with the maintainer before merge, purely a where-does-this-live call, not a code concern. Will follow up shortly. No rush on your end.

jaylfc:
Thanks again for this. Before I place it, I want to get the boundary right so taOS core stays fork-agnostic and Skald stays cleanly decoupled. A few clarifications:

1. **Direction of the dependency.** My understanding: Skald is your taOS fork / external platform, and its TaosDispatcher consumes the controller APIs (cluster workers, load-model, the proposed lease API). So the contract is taOS core surfaces data, Skald consumes it, and taOS core never depends on Skald. Can you confirm that is the intended direction here?

2. **Who owns the manifest.** Who writes `/etc/taos/skald-models.json`, Skald or the taOS worker, and what is its schema? If Skald writes it, then taOS core reading a Skald-named, Skald-owned file couples core to your fork, which is the one thing I would like to avoid.

3. **Naming and placement (the main ask).** The `available_models` field on `WorkerInfo` is a clean, generic worker capability and I am happy to take that into core as-is. My hesitation is a `skald_manifest.py` living in core `tinyagentos/worker/` and reading a `skald-` prefixed file. Could we generalise it so core is Skald-agnostic, for example the worker derives `available_models` from a neutral source it already owns (the GPU catalog, or a generic `worker-models.json`), and Skald becomes one consumer of the resulting field rather than the field being defined by Skald's file? That keeps the genuinely useful part (available models surfaced per worker) in core, and leaves anything Skald-specific in Skald.

4. **Inert by default.** The tests suggest `available_models` defaults to an empty list with no behaviour change when no manifest is present, so non-Skald deployments are unaffected. Can you confirm that holds end-to-end (heartbeat, worker agent, cluster view)?

I am keen to land the generic version quickly, so if you are happy to rename/re-source it to be fork-agnostic I will take it straight in. What is the cleanest split from your side?

jaylfc:
This is exactly the split I was after, thank you. worker_manifest reading a neutral worker-models.json via TAOS_WORKER_MANIFEST keeps core fork-agnostic, and available_models stays a generic worker capability with Skald as one documented consumer. Kilo and CodeRabbit are green and the diff is clean. I have approved the fork CI and set it to auto-merge to dev once the required checks pass. Nice fast turnaround.

PR #1825, opened 2026-07-14
feat(agents): canvas routes join the agent-token allowlist (#1800 slice 2)

jaylfc:
slice 2 of #1800 identity epic, DO NOT MERGE (held for full-set review)

----
## Summary by Gitar

- **Authentication middleware:**
  - Added `_AGENT_CANVAS_ROUTES` and `_is_agent_canvas_path` to include specific canvas endpoints in the agent-token allowlist.
  - Updated `AuthMiddleware` to authorize these routes via bearer tokens, explicitly excluding sensitive `PATCH` operations.
- **Testing:**
  - Added comprehensive unit tests in `test_auth_middleware.py` verifying both route matching logic and middleware access control.

## Summary by CodeRabbit

- **New Features**
  - Added support for agent-token access to approved project canvas operations, including viewing and removing canvas elements.
  - Supported canvas routes now work with eligible bearer-token authentication.

- **Bug Fixes**
  - Improved authorization safeguards to reject unsupported methods, malformed paths, nested routes, and unauthorized canvas permission changes.
  - Corrected snapshot filename matching so similarly named paths are not incorrectly accepted.

### DIFF
--- a/tests/test_auth_middleware.py
+++ b/tests/test_auth_middleware.py
@@ -304,6 +304,19 @@ class TestIsAgentCanvasPath:
     def test_nested_element_path_not_allowed(self):
         assert _is_agent_canvas_path("GET", "/api/projects/proj-1/canvas/elements/x/y") is False
 
+    def test_single_element_get_not_allowed(self):
+        # There is no GET /elements/{id} route in the allowlist; only the
+        # collection GET and the DELETE of a single element are permitted.
+        assert _is_agent_canvas_path("GET", "/api/projects/proj-1/canvas/elements/el-1") is False
+
+    def test_snapshot_dot_is_literal_not_wildcard(self):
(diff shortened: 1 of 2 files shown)


jaylfc:
Folded an adversarial-review finding: the snapshot.png and snapshot.tldr allowlist regexes used an unescaped dot, so the pattern matched a near-miss like snapshotXpng. Escaped both dots for a literal match and added regression tests (snapshotXpng, snapshot_png, snapshotXtldr, and the single-element GET boundary). Note on the reviewer's other flag (agent-token branch sits before the first-boot onboarding gate): that ordering is pre-existing for the task and A2A agent routes, and it is not a bypass because the branch only passes through while the route itself verifies the registry JWT and fails closed, and no valid agent token can exist on an unconfigured system. Not folding it here; it can be a separate defense-in-depth note if we want the gate moved ahead of the pass-through.

PR #1826, opened 2026-07-14
feat(agents): consent approval sets registry handle (#1800 slice 7)

jaylfc:
slice 7 of #1800 identity epic, DO NOT MERGE

When an agent auth-request is approved, set the registry handle from the
sanitized identity claim (via _slugify). If the handle collides with an
ACTIVE identity, return 409 and leave the request PENDING so the approver
can pick a different variant. A SUSPENDED holder does not block reuse.


jaylfc:
Superseded by #1838, which landed all seven slices as one consistent set on dev (with the CI-caught active-handle migration-order fix folded). Closing.

PR #1859, opened 2026-07-16
feat(scheduler): land GPU arbiter on one VRAM authority + eviction/heartbeat fixes (#894/#185, continues #1718)

jaylfc:
Maintainer takeover of the GPU arbiter epic (#894), continuing @hognek's work on #1718 (his commits are preserved here). Rebased onto current dev and closes the #185 gate.

## The #185 gate-closer: one VRAM authority

The arbiter shipped inert beside #1725's now-merged `VramReservationManager`, so there were two reservation ledgers that would double-count once `submit_gpu` got a caller. This makes the manager the single ledger:

- The arbiter no longer keeps its own `_reserved_vram_mb` accounting. It reserves against the shared `VramReservationManager` that the model-load path (`routes/models.py`) already uses. `app.py` builds one manager and passes it into the arbiter, so GPU scheduler tasks and model loads draw from one authority and never double-count.
- `GpuArbiter.reserve_vram()` / `release_vram()` is the arbiter front door for non-task callers.
- Folding the manager in also retires the arbiter's own probe. The manager probes nvidia-smi in a worker thread (no event-loop stall while holding the reservation lock) and distinguishes no-probe (fail-open, cluster `claim_lease` parity) from known-full, which is the behaviour #1725 already reviewed and shipped.

Model-loads-as-evictable-`submit_gpu`-tasks (full queue/eviction semantics for pulls, so a pull that can't fit waits or evicts instead of 503) is intentionally a **follow-up**, not this PR: eviction of a user's running model is a product/policy decision, and this PR keeps model-pull behaviour identical to today. The one-ledger foundation here is the prerequisite for that upgrade.

## Correctness fixes

- **Eviction ordering (Xid-62).** `_evict_task` now cancels AND awaits the evicted task before releasing its reservation + lease, so an evict-to-make-room re-admission cannot pass against not-yet-reclaimed VRAM and start a second concurrent load. Regression test added.
- **Equal-priority eviction.** `evict_lowest_priority` no longer preempts an equal-priority running task (strictly-lower priority only), avoiding eviction thrash between peers.
- **Draining worker heartbeat (#890).** `heartbeat()` no longer flips a `draining` worker back to `online`; that had re-entered it into routing before its leases finished and defeated graceful drain. Regression test added.
- **`/update` deploy result.** `routes/cluster.py` now `raise_for_status()` on the worker deploy response before reading it, so a failed deploy is not reported as `status: "updating"`.

## Tests

`VramReservationManager` gains an injectable probe for tests. The TOCTOU tests now assert against the shared ledger. Full local run green: arbiter (894 + toctou), cluster, leases, routes/cluster, routes/models, vram_reservation — 167 passed, plus the two new regression tests.

Supersedes #1718 (folds in #1690's worker-auto-update + lock fix via the same branch ancestry). #1725 and #1726 are already merged.

## Summary by CodeRabbit

* **New Features**
  * Added graceful and forced worker draining with drain cancellation and coordinated worker updates.
  * Introduced GPU-aware admission control using shared VRAM reservations, prioritization, and eviction to prevent overcommit.
  * Added GPU resource discovery and extended task resource estimates to include VRAM.
* **Bug Fixes**
  * Ensured draining workers stay draining across heartbeats and are excluded from routing, catalog, and lease claims.
  * Improved GPU eviction/termination to reliably cancel work and release reservations without race conditions.
* **Tests**
  * Added coverage for worker drain behavior and GPU admission/eviction, including reservation TOCTOU and async ordering.


jaylfc:
Addressed the Kilo review in 4839501a:
- WARNING (arch substring): compute_cap is now matched exactly (sm_8 can no longer prefix-match sm_86); the freeform model string keeps a substring match.
- WARNING (resource prefix collision): added _resource_on_worker using the cluster manager's exact _parse_resource_id so gpu-node cannot match gpu-node-2, the same guard #1726 applied on the lease paths; used by both the arch check and cluster admission.
- SUGGESTION (cancel_running_for_leases already_completed count): left as-is; it is a telemetry counter and conflating not-running with evicted there does not affect behavior.
Full suite green locally (arbiter + cluster + leases + routes/cluster + routes/models + vram_reservation).

Issue #1862, opened 2026-07-16
Agent token -> per-project grant lookup (S2a)


jaylfc:
Relax the agent JWT from single-project binding to an identity token whose project access is resolved via per-project grants at request time, so one agent identity can hold membership in several projects. Backend prerequisite for the multi-project Agents-app invite (S5).

jaylfc:
Done: per-project grants (3-col unique + boot-safe legacy migration), grant-gated identity token (Option B), join-existing-agent + admin assign-agent route. Kilo-reviewed: cross-project escalation hardened to safe-by-construction + a concurrent-write race fixed with a serialization lock. Merged in #1866.

PR #1866, opened 2026-07-16
feat(agents): multi-project agent identities (per-project grants, grant-gated token) (#1862)

jaylfc:
## Summary
Enable ONE external-agent identity to be a member of MULTIPLE projects. The token is now the identity (project-agnostic); per-project GRANTS are the sole authorization.

- Grants table now unique on (canonical_id, scope, project_id) instead of (canonical_id, scope). An idempotent boot-path migration rebuilds the table for existing DBs still on the old 2-column unique.
- add_grant is NULL-safe so re-approving the same (canonical_id, scope, project_id) key is idempotent even with a NULL project_id.
- check_agent_scope_for_project no longer hard-pins the token's project_id claim. Project access is grant-gated: authorized purely from a matching active grant. One project-agnostic token authorizes any project the agent holds a grant for; projects without a grant (or with an expired grant) 403. The mint_registry_token project_id claim is now advisory only.
- Extracted add_agent_to_project shared primitive (grants + idempotent project_members row + a2a channel membership). An already-registered agent whose handle collides on an active identity is now ADDED to the new project (identity + token reused) instead of 409ing.
- Added POST /api/projects/{project_id}/members/assign-agent (owner-or-admin gated) for the Agents-app UI to assign an existing agent to a project.

## Notes
- Option B token model: project access is grant-gated, the token project_id claim is advisory.
- No AI attribution, no em dashes.

## Summary by CodeRabbit

- **New Features**
  - Agents can now belong to multiple projects with the same identity.
  - Added an admin/owner-only endpoint to assign an existing agent to another project.
  - Project access is now granted based on active project-specific permissions (including support for global tokens).
- **Bug Fixes**
  - Improved multi-project invitation/auth approval to reuse existing identities without collisions.
  - Hardened approval behavior so grant/membership is bound to the correct project.
  - Upgraded existing grant databases safely to the new per-project uniqueness rules.
- **Documentation**
  - Updated agent API documentation for multi-project identity and grant authorization behavior.

### DIFF
--- a/docs/agent-coordination.md
+++ b/docs/agent-coordination.md
@@ -123,6 +123,17 @@ PIN; see issue #1780). When you change the
 allowlist in `tinyagentos/auth_middleware.py`, update this section in the same
 PR (the doc-gate enforces it).
 
+Multi-project identities (taOS #1862): one agent identity (the registry JWT) may
+belong to several projects at once. The grants table keys a grant on
+`(canonical_id, scope, project_id)`, so the same scope can be held for multiple
+projects. Project access is GRANT-GATED, not claim-gated: the token's `project_id`
+claim is advisory only, and `check_agent_scope_for_project` authorizes a project
+purely from a matching active grant. An already-registered agent is added to a
+further project via `POST /api/projects/{project_id}/members/assign-agent`
+(admin/owner gated) or by redeeming an invite whose handle collides with an
+active identity (the existing canonical_id and token are reused instead of
+409ing).
+
 ## Project invite redeem route (link + PIN)
 
 A project invite lets an external agent join without going through the consent
--- a/tests/test_agent_grants_store.py
+++ b/tests/test_agent_grants_store.py
@@ -1,4 +1,5 @@
 """Tests for AgentGrantsStore — per-agent scope grants persistence."""
+import aiosqlite
 import pytest
 
 from tinyagentos.agent_grants_store import AgentGrantsStore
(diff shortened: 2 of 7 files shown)


jaylfc:
Addressed the Kilo review in 9b117d71:
- WARNING (agent_auth_requests.py:374, multi-project ADD binds to effective_project): the branch now gates AND binds strictly on the EXPLICIT validated project_id, never effective_project. It was safe only by the invariant that effective_project == project_id when project scopes are granted (the guard rejects an absent project_id); it is now safe by construction. Added a regression test: an agent that self-requests project C but is approved for project B lands in B only, never C.
- SUGGESTION (agent_grants_store.py:144, non-atomic DELETE-then-INSERT): accepted as low risk. add_grant runs on the store's single connection and the approval flow serializes grant writes per request, so a concurrent same-key collision is not reachable in practice; the DELETE-then-INSERT is deliberate because a 3-column UNIQUE with a NULL project_id will not dedup via INSERT OR REPLACE (NULL != NULL). Left as-is; can revisit with a transaction wrap if a concurrent path appears.
- SUGGESTION (grant recency): same DELETE-then-INSERT keeps re-approval idempotent and refreshes granted_at, which is the intended behavior.
The Option B token model (project access is grant-gated; the token project_id claim is advisory) is called out in the code + docs/agent-coordination.md - flagging it for your sign-off.

jaylfc:
The CI failure (test_concurrent_mint_same_handle_makes_one_row on 3.12) was the exact race Kilo's SUGGESTION warned about, and I was wrong to dismiss it. Fixed in the latest commit: add_grant now serializes its delete+insert+select behind an asyncio.Lock, so concurrent same-key writes cannot interleave. 304 passing across the grants/mint/registry/auth net locally.

Issue #1765, opened 2026-07-09
fix(worker): manifest path crashes worker on malformed input + availability depends on backend liveness (audit follow-up to #1762)


jaylfc:
Follow-up to #1762 from a code audit of the merged tree. Two real issues plus two smaller ones in the worker manifest path, all in `tinyagentos/worker/worker_manifest.py` + `tinyagentos/worker/agent.py`. Over to @hognek since this is your feature and you know it best.

**1. HIGH: a malformed manifest crashes the worker process.** `load_manifest()` calls `json.loads` unguarded (worker_manifest.py, final line of the loader), so invalid JSON raises `JSONDecodeError`; and the enrichment loop does an unchecked `m["model_id"]` (agent.py, enrichment loop) that raises `KeyError` on an entry missing the key. Both fire inside `detect_backends()`, which `register()` calls with no try/except, and `__main__._run_headless` only catches KeyboardInterrupt, so the whole worker dies (crash-loop under systemd). The docstring says the file may be written by any external platform, exactly the input class that will eventually be malformed. Failure scenario: an admin typos one comma in `/etc/taos/worker-models.json` and the worker crash-loops on start, or, if the file goes bad while running, `heartbeat()`'s blanket `except Exception: return 0` swallows it with no log line and the worker silently drops offline.
Fix shape: wrap the manifest read + enrichment in try/except, log a warning with the parse error, and fall back to no-manifest behaviour. Also add a log line in the heartbeat swallow. Please update the two tests that currently bless the crash (`test_invalid_json_raises`, `test_missing_model_id_handled`).

**2. MEDIUM: manifest models only surface when a matching backend is already live.** The module promises the controller learns what a worker COULD load independently of what is loaded, but enrichment attaches manifest entries only to backends that passed the live probe, so if llama.cpp is stopped (the exact could-load-but-not-loaded case) its manifest entries vanish from the heartbeat. Fix shape: emit manifest entries for declared-but-not-probed backends too (e.g. a synthetic backend entry or a top-level available_models fallback), so availability does not depend on liveness.

**3. Dead mappings:** `SOFTWARE_TO_BACKEND_TYPE` maps `kokoro`/`whisper` to backend types that appear nowhere in the probe candidates list, so those manifest entries are unreachable, despite tests asserting the mappings. Either add the probe candidates or drop the mappings.

[...]

PR #1870, opened 2026-07-17
fix(worker): manifest crash-loop follow-up — availability/liveness decoupling, loaded_models status, unreachable mappings

hognek:
## Summary
Fixes the remaining items from #1765 (follow-up to merged PR #1762, hotfix #1767 already addressed the crash-loop).

### MEDIUM: Availability/liveness decoupling
`detect_backends()` now emits synthetic backend entries for manifest-declared software types that have no running (probed) counterpart. Stopped-but-installed backends declared in the worker manifest now advertise their available models to the controller. Previously, manifest entries vanished from heartbeats when their backend wasn't running.

### Low: Drop unreachable mappings
Removed `kokoro` and `whisper` from `SOFTWARE_TO_BACKEND_TYPE`. These were unreachable because no probe candidates exist for those backend types. Adding probes would require backend support that isn't implemented.

### Nit: loaded_models for status
Changed `detect_backends()` enrichment to use `loaded_models` (from /api/ps, in-memory) instead of the full models catalog (from /api/tags, on-disk) when determining whether a manifest-declared model is 'loaded'. Previously, models merely present on disk were reported as 'loaded' rather than 'available'.

## Files changed
- `tinyagentos/worker/worker_manifest.py` — drop kokoro/whisper from SOFTWARE_TO_BACKEND_TYPE
- `tinyagentos/worker/agent.py` — loaded_models-based status + synthetic backend entries for stopped backends
- `tests/test_worker_manifest.py` — updated mapping tests
- `tests/test_worker.py` — 6 new tests (4 stopped-backend emission + 2 loaded_models status)

## Tests
All 150 worker+manifest tests pass (0 failures).

## Summary by CodeRabbit

* **New Features**
  * Worker backend detection now enriches results with manifest-declared models, including for backend types that aren’t currently running.
  * Manifest-only backends are listed as **stopped**, with **available** models populated from the manifest.
  * Model status reporting now distinguishes **loaded** models from those only **available** in the catalog.

* **Bug Fixes**
  * Worker registration no longer risks sending an empty backend URL when synthetic backends are present.
  * Backend detection gracefully handles manifest load/enrichment failures.

* **Tests**
  * Expanded regression coverage for backend synthesis and model-status enrichment behaviors.

### DIFF
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -252,6 +252,62 @@ class TestRegistration:
 
 
 @pytest.mark.asyncio
+class TestRegistrationUrlFallback:
+    """Regression: when all backends are synthetic/stopped (url=None),
+    the registration URL must fall through to get_worker_url()."""
+
+    async def test_register_url_falls_back_when_all_backends_stopped(self, tmp_path):
+        """Synthetic stopped backends have url=None. The registration URL
+        must skip None-URL backends and fall through to get_worker_url()
+        instead of sending None, which causes a silent registration-failure
+        loop.  Regression test for #1765 review finding."""
+        import json as _json
+
+        save_signing_key(tmp_path, secrets.token_bytes(32))
+        agent = WorkerAgent(
+            "http://controller:6969", name="test-worker", state_dir=tmp_path
+        )
+        mock_response = MagicMock()
+        mock_response.status_code = 200
+        mock_response.raise_for_status = MagicMock()
+
+        # Synthetic stopped backend: url is None
+        stopped_backend = {
+            "name": "ollama",
+            "type": "ollama",
+            "url": None,
+            "capabilities": ["llm-chat"],
+            "models": [],
+            "loaded_models": [],
+            "status": "stopped",
+            "kv_quant_support": {},
+            "available_models": [],
+        }
+
+        with patch("tinyagentos.worker.agent.httpx.AsyncClient") as mock_client_cls:
+            mock_client = AsyncMock()
+            mock_client.__aenter__ = AsyncMock(return_value=mock_client)
+            mock_client.__aexit__ = AsyncMock(return_value=False)
+            mock_client.post = AsyncMock(return_value=mock_response)
+            mock_client_cls.return_value = mock_client
+
+            with patch(
+                "tinyagentos.worker.agent.WorkerAgent.detect_backends",
+                return_value=[stopped_backend],
+            ):
+                result = await agent.register()
+
+            # Registration must succeed AND the payload URL must be a
+            # real reachable URL, not None.
+            assert result is True
+            assert mock_client.post.called
+            call_args = mock_client.post.call_args
+            body = _json.loads(call_args[1]["content"])
+            assert body["url"] is not None
+            assert body["url"].startswith("http://")
+
+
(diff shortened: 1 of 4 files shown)


hognek:
Addressed Kilo findings (all 3) in commit dc9ab0b:

- **WARNING (url:''):** Changed synthetic backend url from `""` to `None` — semantically correct for 'no endpoint', and the register() fallback chain treats None as falsy (falls through to get_worker_url()).
- **SUGGESTION (name without port):** Added comment explaining that synthetic entries use the bare type name (not type:port) since there's no probed instance.
- **SUGGESTION (hardcoded kv_quant):** Changed kv_quant_support from hardcoded fp16 to `{}` — no data is more honest than over-reporting. detect_kv_quant_support() adds the fp16 baseline anyway.

All 51 worker+manifest tests pass. CodeRabbit and CLA are green.

jaylfc:
Deep-reviewed this at code level @hognek. The availability/liveness decoupling, the `loaded_models` status-from-`/api/ps` fix, and the malformed-input guards are all solid, and the 6 new tests give good branch coverage (51 passed locally). One blocking regression before merge though:

**`register()` produces `url: None` for a synthetic-only backend list (`tinyagentos/worker/agent.py`, worker_url selection ~L464-467).**

```python
worker_url = self.advertise_url or adv_url or (backends[0]["url"] if backends else self.get_worker_url())
```
`get_worker_url()` is only reached when `backends` is **empty**. With a synthetic-only list, `backends` is truthy, so the ternary returns `backends[0]["url"]`, which is now `None`. With no `advertise_url` / `TAOS_ADVERTISE_IP` (bare-metal / Mac), `worker_url` becomes `None`. `WorkerRegister.url` is a required non-nullable `str` (`routes/cluster.py:217`) and the handler also 400s on a falsy url (`:335`), so the POST fails -> `register()` returns `False` -> 5s retry loop.

That is exactly issue #1765's item-2 case: a worker whose only backend is installed-but-stopped, no advertise IP. It doesn't crash-loop under systemd anymore, but it converts the crash into a silent registration-failure loop, so the loop moves rather than being eliminated for that config.

Heads up: the earlier note that changing `""` -> `None` fixes this via "the fallback falls through to `get_worker_url()`" isn't right - `None` doesn't reach `get_worker_url()` because that branch is gated on `backends` being empty, not on the url being falsy. Kilo's WARNING on this line is real and still live in the head.

Suggested fix - pick the url from live backends only, then fall back:
```python
live_url = next((b["url"] for b in backends if b.get("url")), None)
worker_url = self.advertise_url or adv_url or live_url or self.get_worker_url()
```
And a regression test asserting `register()`'s payload url is non-None when the only backend is synthetic/stopped would lock it (the current tests assert `detect_backends()` shape but never the registration url, which is why this slipped through).

Non-blocking: (1) the synthetic `name` is the bare backend type rather than `type:port`, so a backend flipping stopped<->running gets two catalog identities - benign churn, fine to defer. (2) A few em dashes in the commit subject/body and test comments - please swap for `--` per the no-em-dash house rule (it's in the taos-development-skill soul.md now).

Everything else looks good - fix the registration url and I'll merge. Nice work on the decoupling.

Issue #866, opened 2026-06-13
Window management: windows lost off-screen on display change + dock right-click controls (minimise/maximise/close)


jaylfc:
## Bug: windows become unreachable when the viewport shrinks
**Repro (Jay):** ran taOS fullscreen on an ultrawide with several windows open, then moved the taOS window/browser to a 16:9 display. The windows opened on the ultrawide were positioned off the smaller viewport - couldn't see them or move them onto the new display.

**Cause:** window positions are absolute and are not re-clamped when the viewport/display size changes (going from a wide to a narrow screen leaves windows beyond the right/bottom edge, with no way to drag them back since their title bars are off-screen).

**Fix:** on viewport resize (and on app focus), **re-clamp every open window** so its title bar stays within the visible bounds (nudge off-screen windows back in by their top-left, like macOS does). Consider a "gather windows" affordance too.

## Feature: dock right-click controls
Right-click an app's dock icon -> a context menu with **Minimise / Maximise / Close** (and pin/unpin already exists). This also gives a way to recover/act on a window whose chrome is off-screen.

- **Close confirmation:** some apps should confirm before closing (unsaved work, long-running task). Add an optional **manifest flag** (e.g. `confirmOnClose: true` in the app manifest); the dock Close (and the window close button) then show a confirm dialog for those apps.

Both live in the window manager / dock (`desktop/src/stores/process-store.ts`, `Dock`, `Window.tsx`). Logged for the queue; not started (wind-down).

jaylfc:
Landed via #1874 (merged to dev). Off-screen windows now clamp back into the visible area on viewport resize. Thanks @hognek - reviewed at code level (clamp math solid across all edge cases, 13/13 tests, tsc clean); added one tiny em-dash polish commit.

PR #1874, opened 2026-07-17
feat(desktop): clamp off-screen windows on viewport resize

hognek:
## Summary
Adds viewport-resize clamping so windows don't become unreachable when the browser moves from an ultrawide monitor to a 16:9 screen.

The dock right-click context menu (minimise/maximise/close) was already present in DockIcon.tsx — this PR completes the remaining piece of \#866.

### `desktop/src/stores/process-store.ts`
- Added `reclampAllWindows()` — iterates visible, non-maximized, non-snapped windows and re-clamps them into the viewport using the existing `safeBounds()` helper. Skips minimized/maximized/snapped windows.

### `desktop/src/components/Desktop.tsx`
- Added a `useEffect` resize listener with 150ms debounce that calls `reclampAllWindows()` on viewport resize.

### `desktop/src/stores/process-store.test.ts`
- 5 new test cases: off-screen clamp, on-screen unchanged, skip minimized, skip maximized, skip snapped.

## Testing
```
npx vitest run --run \
  src/stores/process-store.test.ts \
  src/components/__tests__/DockIcon.context-menu.test.tsx \
  src/components/Dock.test.tsx \
  src/components/DockIcon.test.tsx \
  src/stores/dock-store.test.ts \
  src/components/dock/__tests__/DockVariants.test.tsx
```
**Result:** 6 files, 41 tests — all pass.

Closes \#866

### DIFF
--- a/desktop/src/components/Desktop.tsx
+++ b/desktop/src/components/Desktop.tsx
@@ -1,4 +1,4 @@
-import { useState, useCallback } from "react";
+import { useState, useCallback, useEffect } from "react";
 import { FolderPlus, Image, Monitor, Settings, LayoutGrid, Layers, BookmarkPlus } from "lucide-react";
 import { useProcessStore } from "@/stores/process-store";
 import { useThemeStore } from "@/stores/theme-store";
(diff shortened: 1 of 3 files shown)


jaylfc:
Reviewed at code level @hognek - the clamp logic is solid: I traced `safeBounds`/`reclampAllWindows` across the edge cases (window larger than viewport, partial off-edge, grow vs shrink, minimized/maximized/snapped skipped) and it never produces a negative/NaN position and always keeps the title bar grabbable; focus and z-order are preserved, and the debounced resize listener is cleaned up correctly. 13/13 vitest pass, tsc clean.

Pushed one tiny maintainer-polish commit: replaced an em dash with `--` in a process-store.ts comment (house no-em-dash rule) - that was the only blocker. Merging once CI re-runs green.

Non-blocking nit for a future pass (not gating this): the trickiest paths - window-larger-than-viewport staying fully visible, the y-axis/title-below-dock clamp, and focus/z-order preservation - aren't directly covered by a test even though the math handles them. A single 'oversized window stays reachable' case would lock it in. Nice work.

Issue #40, opened 2026-04-11
Cluster prep for Pi 4 8GB / Fedora box / Mac mini onboarding


jaylfc:
Specific checklist to onboard tomorrow's three new workers:

**Pi 4 8GB (ARM64, Armbian/Raspberry Pi OS):**
- Run scripts/install-worker.sh
- Verify CPU-only embedding backend (llama.cpp) boots
- Run initial benchmark: focus on embedding throughput (the Pi 4's likely useful contribution)
- Register as cpu-aarch64 resource
- Expected role: offload embeddings/rerank from the RK3588 when NPU is busy

**Fedora box (x86_64, Fedora):**
- Run scripts/install-worker.sh (detect Fedora in distro logic)
- Detect NVIDIA / AMD GPU if present
- Register appropriate gpu-* resource
- Run full benchmark suite
- Expected role: primary GPU inference for image gen, LLM chat

**Mac mini (Apple Silicon or Intel):**
- Run scripts/install-worker.sh (macOS branch)
- If Apple Silicon: register gpu-metal with MLX backend
- Run benchmark suite (MLX tok/sec, Core ML image gen if available)
- Expected role: MLX-accelerated LLM inference

Also needed for all three: backend-driven capability reporting on heartbeat (separate issue).

PR #1889, opened 2026-07-17
fix(ci): extend pip-audit scan to proxy and worker extras

### LINKED ISSUE #740: Dependencies: no Python lockfile → non-reproducible installs (>= floors)
**Fact** (audit 2026-06-10): all Python deps use `>=` floors except `taosmd==0.3.0`, and there is no Python lockfile (no requirements.txt/uv.lock/poetry.lock). Two self-installs weeks apart can resolve different transitive trees → 'works on my Pi' bugs + no hash verification. The frontend `package-lock.json` IS committed (good asymmetry to close). Also: `pip-audit` in security.yml only installs `.[dev]`, so the heaviest extras (`proxy`=litellm+prisma, `worker`=pystray) are never CVE-scanned. **Fix:** add a hash-pinned lockfile (uv.lock or pip-tools requirements) and make pip-audit scan `.[proxy,worker]`. Severity: Medium.

hognek:
## Summary

Part 2/2 of #740. The Python lockfile (uv.lock) was committed in June (PR #40), but pip-audit in security.yml still only scanned `.[dev]` — the heaviest extras (proxy=litellm+prisma, worker=pystray) were never CVE-checked.

## Change

`.github/workflows/security.yml`: install `.[dev,proxy,worker]` instead of `.[dev]` before running pip-audit.

## Why

litellm[proxy] and prisma are complex packages with deep dependency trees; they are the most likely source of CVE-worthy transitive issues. Running pip-audit against only the test-and-dev subset was a blind spot.

Task: t_a842feff. Fixes #740.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -23,7 +23,7 @@ jobs:
           python-version: "3.12"
 
       - name: Install dependencies
-        run: pip install -e ".[dev]"
+        run: pip install -e ".[dev,proxy,worker]"
 
       - name: Check for known vulnerabilities
         # CVE-2026-3219 affects pip itself with no fix released yet; revisit once


hognek:
## Kilo review findings — all false positives for GitHub Actions

**WARNING: prisma downloads native engine binaries**
GitHub Actions runners have unrestricted internet access. prisma's engine download is the standard install path and succeeds on every major CI platform including GH Actions. If this ever becomes a concern, prisma can be pre-downloaded via `PRISMA_ENGINES_MIRROR`.

**SUGGESTION: pystray needs GUI system libs**
pystray is a pure-Python package. Its Linux backend (python-xlib) is also pure Python — no native compilation. `pip install pystray` succeeds on headless Linux runners (verified against PyPI wheel metadata).

**SUGGESTION: use uv sync --frozen instead of pip install**
Valid point for a future refactor — the security.yml predates the uv.lock and uv-based CI. This PR is scoped to the specific fix requested in #740 (extend pip-audit extras). Switching to uv sync is a separate improvement tracked in #740 itself.

Issue #1893, opened 2026-07-17
feat(sharing): user-to-user resource-share grant + consent (mirror agent grants)


jaylfc:
Part of #1890 (taOSnet). There is no user->user sharing primitive - all sharing today is agent-scoped (`relationships.py` agent_permissions, `shared_folders.py` folder_access by agent_name). Build the foundational "user A shares resource R with user B" grant, mirroring the agent grant/consent pattern.

## Build (mirror, do not reinvent)
- New store `tinyagentos/user_shares_store.py` (BaseStore) with table `user_shares(id, owner_user_id, resource_type, resource_id, shared_with_user_id, permission, tier, granted_at, expires_at, UNIQUE(owner_user_id, resource_type, resource_id, shared_with_user_id, permission))`. Model it on `tinyagentos/agent_grants_store.py` (NULL-safe write lock, idempotent re-share, the same _post_init discipline). resource_type starts with just `taosmd`.
- Routes `tinyagentos/routes/user_shares.py`: `POST /api/shares {resource_type, resource_id, to_username, permission}` (resolve to_username via `auth.find_user`, owner = session user, admin/owner gated), `GET /api/shares?direction=out|in`, `DELETE /api/shares/{id}`. On share-create, raise a consent decision to the target user (reuse the Decisions app machinery landing in #1883 once merged; until then a notification).
- A helper `user_can_access(resource_type, resource_id, user_id)` for consumers.

## Acceptance
- User A shares taosmd with user B by username -> a pending share the target must accept; once accepted, `user_can_access` returns true for B; revoke removes it; sharing to an unknown username 404s; duplicate share is idempotent.
- Tests mirror `tests/` agent-grant tests: create/accept/revoke/idempotent/unknown-user.

Scoped to the store + routes + consent wiring (NOT the taOSnet app UI - that is a later issue). TDD, single PR to dev.

PR #1897, opened 2026-07-17
feat(user_shares): add UserSharesStore with NULL-safe write lock

hognek:
Task: t_48e0dee4 (taOS #1893 C1).

Implements `tinyagentos/user_shares_store.py` — the foundational store for user-to-user resource sharing, modeled on `tinyagentos/agent_grants_store.py`.

**Schema:** 9-column `user_shares` table with 5-column NOT NULL UNIQUE constraint.

**Methods:**
- `add_share` — idempotent delete+insert+select under `_write_lock`
- `list_shares(owner_user_id)` — all shares by owner, ordered by granted_at
- `list_shares_received(shared_with_user_id)` — all shares targeting a user
- `list_active_shares()` — non-expired shares
- `revoke_share(share_id)` — delete by id (silent on missing)
- `user_can_access(resource_type, resource_id, user_id)` — active-share gateway check

**Matches agent_grants_store.py conventions:** BaseStore, SCHEMA class variable, `_write_lock` serialisation, `_row_to_dict` helper, `_post_init` hook, RuntimeError guards on all methods.

**resource_type starts with just "taosmd".**

All methods verified with smoke tests locally.

## Summary by CodeRabbit

* **New Features**
  * Added persistent resource sharing between users.
  * Supports configurable permissions, sharing tiers, and optional expiration dates.
  * Added APIs to create, view, revoke, and monitor shared resources.
  * Added access checks to determine whether a user currently has permission to access a resource.
  * Prevents duplicate sharing entries and automatically excludes expired shares from active access.

### DIFF
--- a/tests/test_user_shares_store.py
+++ b/tests/test_user_shares_store.py
@@ -1,0 +1,116 @@
+import pytest
+
+from tinyagentos.user_shares_store import UserSharesStore
+
+
+@pytest.mark.asyncio
+class TestUserSharesStore:
+    async def _store(self, tmp_path):
+        s = UserSharesStore(tmp_path / "shares.db")
+        await s.init()
+        return s
+
+    # ── add_share ────────────────────────────────────────────────────
+    async def test_add_share_returns_inserted_row(self, tmp_path):
+        store = await self._store(tmp_path)
+        try:
+            row = await store.add_share("alice", "taosmd", "mem-1", "bob", "read")
+            assert row["owner_user_id"] == "alice"
+            assert row["resource_type"] == "taosmd"
+            assert row["resource_id"] == "mem-1"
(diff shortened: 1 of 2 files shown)


jaylfc:
Added the tests #1893 asked for (9 passing, mirrors test_agent_grants_store.py) so this lands as the C1 store chunk. One design note for the NEXT chunk (routes + consent), not a blocker here: the store has no status/accepted_at column, so user_can_access grants access as soon as a share row exists - there is no pending->accept gate yet. #1893's acceptance ('a pending share the target must accept; once accepted user_can_access returns true') needs that column, which will be a _post_init migration when the consent flow lands. The _post_init stub is correctly positioned for it. Merging the store now.

PR #1905, opened 2026-07-17
fix(scheduling): enable WAL mode + fix sync-in-async in SQLite stores

### LINKED ISSUE #653: SQLite: enable WAL mode, fix sync-in-async, add migration system
## Context

Three related SQLite concerns that affect every subsystem:

**1. No WAL mode anywhere** (`base_store.py:13-18`) — All stores default to DELETE journal mode. Concurrent reads/writes from WebSocket chat + background ingest will block each other.

**2. Synchronous sqlite3 in async context** (`scheduling/job_queue.py:100`, `leases.py:50`, `worker_heartbeat.py`, `mesh_sync.py`) — These block the event loop during every DB operation.

**3. No migration system** (`base_store.py`) — Schema changes via ad-hoc `ALTER TABLE ... ADD COLUMN` wrapped in try/except. No version tracking or rollback.

## Suggested Approach

1. Add `await db.execute("PRAGMA journal_mode=WAL")` in `BaseStore.init()` — single line, benefits everything
2. Add `await db.execute("PRAGMA busy_timeout=5000")` to prevent immediate `SQLITE_BUSY`
3. Switch scheduling stores to `aiosqlite` or wrap with `asyncio.to_thread()`
4. Formalize migrations with a `schema_version` table. The existing `migrations/persona_v2.py` shows the pattern — just needs version tracking

hognek:
Fixes #653.

Three related SQLite fixes across the scheduling subsystem:

**1. WAL mode** — `job_queue.py` and `leases.py` now call `apply_wal_pragmas()` on init (WAL + NORMAL synchronous + busy_timeout=5000). `mesh_sync.py` and `worker_heartbeat.py` already had this. Together with `base_store.py`, all SQLite stores in taOS now use WAL mode.

**2. Sync-in-async** — All three files (`job_queue.py`, `leases.py`, `mesh_sync.py`) now dispatch blocking `sqlite3` calls through `asyncio.to_thread()` following the pattern already established by `worker_heartbeat.py`. Public async API is unchanged.

**3. Concurrency safety (review fix)** — Added `asyncio.Lock` to `JobQueue`, `LeaseManager`, and `MeshSync` to guard the shared `sqlite3.Connection` against concurrent `asyncio.to_thread` workers. This prevents interleaved `BEGIN IMMEDIATE … COMMIT` transactions when multiple callers operate on the same singleton instance (e.g., `app.state.job_queue` reused across concurrent requests).

**4. Migration system** — Verified robust. `db_migrations.py` has version-tracked migrations with `schema_migrations` table, baseline-at-latest semantics for pre-existing DBs, and both sync/async variants. Tests cover fresh DB, existing DB baseline, idempotent re-run, callable migrations, and WAL pragma assertions.

**Also:**
- `apply_wal_pragmas` now sets `busy_timeout=5000` to match every other store (litellm_keystore/agent_budget_store/broker).
- `_import_delta_sync` no longer swallows insert failures with bare `except: pass` — logs a warning instead.

**Tests: 60/60 pass** (`test_db_migrations.py` + `tests/scheduling/`).

### DIFF
--- a/tinyagentos/db_migrations.py
+++ b/tinyagentos/db_migrations.py
@@ -92,6 +92,7 @@ CREATE TABLE IF NOT EXISTS schema_migrations (
 
 def apply_wal_pragmas(conn: sqlite3.Connection) -> None:
     """Enable WAL journal mode and NORMAL synchronous on *conn*."""
+    conn.execute("PRAGMA busy_timeout = 5000")
     conn.execute("PRAGMA journal_mode = WAL")
     conn.execute("PRAGMA synchronous = NORMAL")
 
--- a/tinyagentos/scheduling/job_queue.py
+++ b/tinyagentos/scheduling/job_queue.py
@@ -101,6 +101,7 @@ class JobQueue:
         self._db_path = str(db_path)
         self._conn: sqlite3.Connection | None = None
         self._limits: dict[str, int] = dict(DEFAULT_LIMITS)
+        self._lock = asyncio.Lock()
 
     def _sync_init(self) -> None:
         self._conn = sqlite3.connect(self._db_path, isolation_level=None, check_same_thread=False)
@@ -133,7 +134,8 @@ class JobQueue:
 
     async def init(self) -> None:
         Path(self._db_path).parent.mkdir(parents=True, exist_ok=True)
-        await asyncio.to_thread(self._sync_init)
+        async with self._lock:
+            await asyncio.to_thread(self._sync_init)
 
     async def close(self) -> None:
         if self._conn:
@@ -174,10 +176,11 @@ class JobQueue:
         estimated_seconds: int = 0,
     ) -> str:
         """Add a job to the queue. Returns job ID."""
-        return await asyncio.to_thread(
-            self._sync_enqueue,
-            job_type, payload, agent_name, priority, resource_type, estimated_seconds,
-        )
+        async with self._lock:
+            return await asyncio.to_thread(
+                self._sync_enqueue,
+                job_type, payload, agent_name, priority, resource_type, estimated_seconds,
+            )
 
     # ------------------------------------------------------------------
     # Dequeue (pull-based)
@@ -236,7 +239,8 @@ class JobQueue:
 
         Returns the job dict or None if nothing is eligible.
         """
-        return await asyncio.to_thread(self._sync_dequeue, resource_types)
+        async with self._lock:
+            return await asyncio.to_thread(self._sync_dequeue, resource_types)
 
     # ------------------------------------------------------------------
     # Complete / fail
@@ -252,7 +256,8 @@ class JobQueue:
 
     async def complete(self, job_id: str, result: dict | None = None) -> bool:
         """Mark a job as completed."""
-        return await asyncio.to_thread(self._sync_complete, job_id, result)
+        async with self._lock:
+            return await asyncio.to_thread(self._sync_complete, job_id, result)
 
     def _sync_fail(self, job_id: str, error: str) -> bool:
         now = time.time()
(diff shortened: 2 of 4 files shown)


jaylfc:
WAL pragmas land correctly and don't conflict with #1900's `BEGIN IMMEDIATE`, but there's a **concurrency-correctness blocker** in the sync-in-async fix. Each store keeps a single shared `self._conn` (`check_same_thread=False`); the pre-PR design was implicitly safe because sync sqlite ran inline on the single-threaded event loop. Moving every method to `asyncio.to_thread` lets concurrent callers on the same instance run `_sync_*` in parallel worker threads on **one connection**, interleaving `BEGIN IMMEDIATE … COMMIT` transactions.

Reproduced: 20 concurrent `JobQueue.dequeue()` on one instance over 50 jobs → 0-2 claimed, ~19 `OperationalError: cannot start a transaction within a transaction`. Same for `LeaseManager.acquire()`. This is the primary usage pattern — `routes/jobs.py` caches one `JobQueue` in `app.state.job_queue` reused across all concurrent requests, so the documented 'two callers can't both claim the same job' contract now crashes instead. Kilo flagged the same at `job_queue.py:106`, `leases.py:56`, `mesh_sync.py:184`.

**Fix:** guard each instance with an `asyncio.Lock` around the dispatch (still frees the loop during fsync), OR open a per-op connection inside the thread, OR use aiosqlite.

Also: add `busy_timeout=5000` to `apply_wal_pragmas` (db_migrations.py:82) to match every other store (litellm_keystore/agent_budget_store/broker all set it) — without it `BEGIN IMMEDIATE` contention returns SQLITE_BUSY immediately instead of waiting. Nit: `_import_delta_sync` (mesh_sync.py:537) swallows insert failures with bare `except: pass` while the async twin logs a warning.

hognek:
Addressed the concurrency-correctness blocker:

- **asyncio.Lock** added to JobQueue, LeaseManager, and MeshSync — each instance serializes  dispatch through its own lock, so the single shared  is never accessed by two parallel worker threads. This fixes the interleaved  bug you reproduced with 20 concurrent  calls.

- **busy_timeout=5000** added to  (db_migrations.py:82) to match litellm_keystore/agent_budget_store/broker.

- **_import_delta_sync** bare  replaced with  to expose per-record sync failures.

Tests: 60/60 pass (test_db_migrations.py + tests/scheduling/).

hognek:
Converted from `sqlite3` + `asyncio.Lock` + `asyncio.to_thread` to native `aiosqlite` — the established taOS pattern used by BaseStore and 20+ other stores.

**Why:** The `asyncio.Lock` approach serialized all access (including reads), and `check_same_thread=False` on a shared `sqlite3.Connection` is fundamentally unsafe under concurrency. `aiosqlite` handles thread safety internally — each `await self._db.execute(...)` dispatches to its own internal thread, and WAL mode allows concurrent readers.

**Changes:**
- `job_queue.py`: Removed asyncio.Lock + 13 `_sync_*` helpers. All SQL operations are now native `await self._db.execute(...)`. Net -118 lines.
- `leases.py`: Same treatment — removed Lock + 9 `_sync_*` helpers. Inlined `_renew_locked` helper. Net -107 lines.
- `mesh_sync.py`: `self._db` uses aiosqlite for peer/sync-log metadata. Caller-supplied `sqlite3.Connection` methods (`export_delta`, `import_delta`, `prepare_agent_bundle`) remain unchanged — they work on the caller's connections. `pull_from_peer` dispatches `_export_delta_sync`/`_import_delta_sync` via `asyncio.to_thread` (those operate on caller connections). Net -29 lines.
- `db_migrations.py`: Added `busy_timeout=5000` to `apply_wal_pragmas_async()` — was already on the sync variant but missing from the async one.

**Tests: 60/60 pass** (`test_db_migrations.py` + `tests/scheduling/`)

**Net: +265 / -383 lines** — simpler, follows the BaseStore convention, no shared-connection thread-safety concerns.

PR #1903, opened 2026-07-17
feat(cluster): worker-initiated graceful pause + drain protocol (taOS #890 C2)

hognek:
Task: t_33babf4b. Fixes #890 (partial — C2 drain protocol).

## What
Implements worker-initiated drain: the worker can self-drain via heartbeat
instead of waiting for the controller to push a drain command.

### Controller side (`cluster/manager.py`)
- `heartbeat()` accepts optional `status` and `drain_reason` fields
- Workers can report `"update-available"`, `"draining"`, or `"updating"`
- `_worker_for_resource`, `get_workers_for_capability`, `aggregate_catalog` include `"update-available"` alongside `"online"`
- Monitor loop handles stale `update-available` workers
- Worker-initiated drain emits notifications to activity feed

### Route side (`routes/cluster.py`)
- `HeartbeatBody` extended with `status` and `drain_reason` fields
- Route handler passes them through to `cluster.heartbeat()`

### Worker side (`worker/agent.py`)
- New convenience methods: `report_update_available()`, `initiate_self_drain()`, `notify_drain_complete()`
- `heartbeat()` accepts optional `status` and `drain_reason` parameters

### State machine
```
online → update-available → draining (self-initiated) → updating → restarting → online
                                                                       ↓ (fail)
                                                                   rollback → online
```

### Tests
11 new unit tests covering:
- Self-drain via heartbeat
- Update-available status + routability
- Catalog inclusion for update-available workers
- Lease claims on update-available workers
- Full state transition pipeline
- Drain preservation on subsequent heartbeats
- Update-available re-online on normal heartbeat
- Monitor loop staleness for update-available
- Updating status routing exclusion

### Test results
```
tests/test_cluster.py::TestWorkerDrain — 38/38 passed (0.42s)
tests/test_cluster_worker_protocol.py — 27/27 passed
tests/test_cluster_optimiser.py (unit only) — 17/17 passed
``"

Depends on: C0 (design doc #1896), C1 (version detection, triggers the drain)
Next task: C3 (self-install + restart + re-register + rollback)

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

- **New Features**
  - Workers can now report lifecycle states (update available, graceful draining, and draining complete) along with an optional drain reason.
  - “Update available” workers remain eligible for resource routing and can claim leases.
  - “Updating” workers are excluded from new routing and are handled as a distinct lifecycle stage.

- **Bug Fixes**
[...]

PR #1907, opened 2026-07-17
feat(worker): add WorkerUpdateService for version-detection polling

hognek:
Task: t_2f23c276
Fixes: #890 (C1 child task)

## Summary

Adds a worker-side version-detection service (`WorkerUpdateService`) that mirrors the controller's `AutoUpdateService` but runs on worker nodes.

## Changes

- **`worker/update_check.py`** — new module with:
  - `WorkerUpdateService` — async background service with start/stop/lifecycle
  - `WorkerUpdateConfig` — persisted config (enabled, channel, pinned version, interval)
  - Version comparison, channel filtering, and pin logic utilities
  - Anonymous install-count ping (mirrors `send_version_ping()`)

- **`worker/agent.py`** — wired in:
  - Service started during `run()`, stopped in `stop()`
  - `update_state` field added to heartbeat payload (controller surfaces in cluster view)

- **`tests/worker/test_update_check.py`** — 60 tests:
  - Version parsing, comparison, channel/pin logic
  - Config serialisation and persistence
  - Service start/stop, state export, disabled/no-op paths

## How it works

1. Worker agent starts `WorkerUpdateService` during `run()`
2. Service polls `TAOS_UPDATE_CHECK_URL` (default: `https://taos.my/api/v1/version-check`)
3. Compares installed version vs latest, respecting channel (stable/beta/dev) and pin
4. Sets `update_available` flag; agent includes `update_state` in heartbeat
5. Controller surfaces available updates in Resource Manager / cluster view

Tests: 60/60 pass (new) + 44/44 pass (existing worker tests, no regressions)

## Summary by CodeRabbit

* **New Features**
  * Added automatic worker update checks with configurable release channels, version pins, and polling intervals.
  * Workers now report update availability and status through heartbeat information.
  * Update notifications are tracked to avoid repeated alerts.
  * Update settings are persisted and safely recover from missing or invalid configuration.

* **Bug Fixes**
  * Improved resilience when update-check requests or responses are unavailable.

* **Tests**
  * Added comprehensive coverage for version comparisons, configuration handling, filtering rules, and service lifecycle behavior.

### DIFF
--- a/tests/worker/test_update_check.py
+++ b/tests/worker/test_update_check.py
@@ -1,0 +1,380 @@
+"""Tests for tinyagentos.worker.update_check — WorkerUpdateService logic."""
+from __future__ import annotations
+
+import asyncio
+import json
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+import pytest_asyncio
+
+from tinyagentos.worker.update_check import (
+    WorkerUpdateService,
+    WorkerUpdateConfig,
+    load_config,
+    save_config,
+    is_newer_version,
+    version_matches_channel,
+    version_matches_pin,
+    _parse_version,
+    _default_state_dir,
+)
+
+
+class TestParseVersion:
+    """Tests for _parse_version — version string → numeric tuple."""
+
+    def test_simple_semver(self):
+        assert _parse_version("1.2.3") == (1, 2, 3)
+
+    def test_two_component(self):
+        assert _parse_version("2.0") == (2, 0)
+
+    def test_single_component(self):
+        assert _parse_version("3") == (3,)
+
+    def test_strips_v_prefix(self):
+        assert _parse_version("v1.2.3") == (1, 2, 3)
+
+    def test_strips_prerelease(self):
+        assert _parse_version("1.0.0-beta.40") == (1, 0, 0)
+
+    def test_strips_build_metadata(self):
+        assert _parse_version("1.2.3+build.42") == (1, 2, 3)
+
+    def test_unparseable_returns_empty(self):
+        assert _parse_version("not-a-version") == ()
+
+    def test_empty_string(self):
+        assert _parse_version("") == ()
+
+    def test_whitespace_only(self):
+        assert _parse_version("   ") == ()
+
+
+class TestIsNewerVersion:
+    """Tests for is_newer_version — version comparison."""
+
+    def test_newer_major(self):
+        assert is_newer_version("2.0.0", "1.0.0")
+
+    def test_newer_minor(self):
+        assert is_newer_version("1.3.0", "1.2.0")
+
+    def test_newer_patch(self):
+        assert is_newer_version("1.2.4", "1.2.3")
+
+    def test_equal(self):
+        assert not is_newer_version("1.2.3", "1.2.3")
+
+    def test_older(self):
+        assert not is_newer_version("1.2.2", "1.2.3")
+
+    def test_different_length_newer(self):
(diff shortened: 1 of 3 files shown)


jaylfc:
Clean — self-contained version-detect + anonymous ping, 142 tests pass, Kilo 'No Issues'. Just needs a **rebase onto post-#1903 dev** (trivial payload-dict conflict: both this and #1903 insert keys into the same `heartbeat()` literal in `agent.py`). Land #1903 first, then this. Note (non-blocking): `update_state` is added to the heartbeat payload but the controller never consumes it (not in `HeartbeatBody`, silently dropped by Pydantic) — forward-plumbing for a consumer that doesn't exist yet.

jaylfc:
Held on merge order, not code. The polling loop is leak-free (all `_run_once` exceptions caught, interval floored at 300s, `_stop_event`-driven exit), and 3 of 5 Kilo items are folded. But this is still gated on #1903 landing first (my earlier ask), and #1903 is open, so this needs a rebase onto post-#1903 dev before it can go in. While rebasing, please also address:
- worker/agent.py stop(): the create_task guard fixes the no-running-loop case, but the coroutine can still be left unawaited if run()/the process exits mid-poll, and task exceptions surface only as "never retrieved" warnings.
- worker/update_check.py:783: `self._latest_version` is set before the pin/channel/newer checks, so get_state() can report a `latest_version` that was actually rejected while `update_available=False`.
- Nit: the 90s startup delay at :723 hard-codes past `config.check_interval`.

hognek:
Pushed fix commit c57cdc7 addressing jaylfc's 3 items:

1. **stop() unawaited coroutine**: Moved update-service cleanup into try/finally in run(); simplified stop() to just signal the stop event. No more create_task() fire-and-forget.

2. **Stale latest_version**: Moved `_latest_version` assignment after pin/channel/newer checks. get_state() now correctly reports None until a version passes all gates.

3. **Hard-coded 90s delay**: Replaced with `min(config.check_interval, 90)` so the initial delay respects configured intervals.

Tests: 122/122 targeted pass (update_check + worker agent). Local gate running.

hognek:
Kilo suggestion fix: added `max(config.check_interval, 10)` floor to the initial_delay calculation at `update_check.py:280`. 

Before: `initial_delay = min(config.check_interval, 90)` — had an upper cap but no lower floor, so a misconfigured 0/negative/tiny check_interval could yield a ~0s initial delay.

After: `initial_delay = min(max(config.check_interval, 10), 90)` — now floored at 10s, matching the pattern the main loop already uses (`interval = max(config.check_interval, 300)` on line 293).

All 60 update_check tests pass.

jaylfc:
Folds noted, thanks. This is still gated on #1903 landing first (my earlier ask) and #1903 is still open, so this needs a rebase onto post-#1903 dev before it can merge. Land #1903, then ping me here.

jaylfc:
Chain-head, near-ready (the mkdir-before-write is already folded at head, thanks). One real robustness bug: worker/agent.py:694 stop() is sync and fire-and-forgets the update-service cancellation via asyncio.ensure_future, which raises/leaks if called with no running loop or during shutdown - await the shielded cancellation or guard the no-loop path. Optional low-value nits if quick: update_check.py:135 lstrip('v') strips a char-set not a prefix (vv1.2.3 -> 1.2.3, use removeprefix); :325 follow_redirects as a per-request kwarg breaks on httpx <0.20; :276/:280 hard-coded 90s startup delay with no floor on check_interval. Fix the stop() leak and I merge (unblocks #1907's downstream chain).

PR #1912, opened 2026-07-17
fix(app): remove redundant app.state assignments from lifespan

### LINKED ISSUE #642: App state double-initialization during startup
## Context

`create_app()` in `app.py` sets ~60 attributes on `app.state` twice — once eagerly (lines 1038-1103) and again inside the lifespan (lines 556-770). The lifespan creates NEW instances that shadow the eagerly-set ones. Any request arriving during the startup window sees stale objects.

Objects affected include `MCPSupervisor`, `RestartOrchestrator`, `TraceStoreRegistry`, `BridgeSessionRegistry`, `CopilotTicketStore`, `CopilotHub`, `WantsReplyRegistry`, and `TypingRegistry`.

## Suggested Approach

Option A (simple): Move all initialization into the lifespan. Remove the eager block. Requests during startup get a 503 (correct behavior — app is not ready).

Option B (robust): Keep eager init as a warming phase but share object references with the lifespan. Do not create new instances — just call `.init()` / `.start()` on the existing ones.

hognek:
## Summary
Removes ~100 redundant `app.state.X = local_X` reassignments from the lifespan that were already set by the eager section in `create_app()`.

## Problem (closes #642)
`create_app()` set ~100+ app.state attributes twice — once eagerly (after app creation, for test compatibility) and again in the lifespan. The lifespan also created duplicate instances of `IdempotencyCache`, shadowing the eager one.

## Fix
- Removed redundant `app.state.X = local_X` lines from the lifespan (106 lines removed)
- Preserved all `.init()` calls — stores are still initialized in the correct order
- Removed duplicate `IdempotencyCache`/channel_hub_connectors/deploy_tasks creation in lifespan
- Lifespan-only objects (MCPSupervisor, TraceStoreRegistry, browser stores, etc.) keep their assignments — the eager section correctly sets them to None
- Eager section is unchanged — tests that use `create_app()` without running the lifespan continue to work

## Testing
- All config tests pass (34/34)
- Router registration tests pass (4/4)
- Agent DB/heartbeat/activity/a2a/admin tests pass
- `create_app()` smoke test: all attributes correctly populated
- Syntax check: clean

## Summary by CodeRabbit

* **Refactor**
  * Improved application startup initialization so core services and configuration are available earlier.
  * Streamlined startup sequencing for authentication, agents, storage, monitoring, and background services.
  * Preserved health monitoring and service startup behavior while reducing redundant initialization work.
* **Bug Fixes**
  * Improved reliability for components accessed before full application startup completes.

### DIFF
--- a/tinyagentos/app.py
+++ b/tinyagentos/app.py
@@ -479,22 +479,13 @@ def create_app(data_dir: Path | None = None, catalog_dir: Path | None = None) ->
         # Arm the startup guard: block non-exempt requests until init completes.
         app.state._startup_complete = False
         await agent_registry_store.init()
-        app.state.agent_registry = agent_registry_store
-        app.state.agent_registry_keypair = agent_registry_keypair
         await auth_requests_store.init()
-        app.state.auth_requests = auth_requests_store
         await agent_grants_store.init()
-        app.state.agent_grants = agent_grants_store
         await app_grants_store.init()
-        app.state.app_grants = app_grants_store
         await license_acceptances_store.init()
-        app.state.license_acceptances = license_acceptances_store
         await agent_model_key_store.init()
-        app.state.agent_model_keys = agent_model_key_store
         await cluster_pairing_store.init()
-        app.state.cluster_pairing = cluster_pairing_store
         await capability_map_store.init()
-        app.state.capability_map = capability_map_store
         await metrics_store.init()
         await notif_store.init()
         await notif_push_store.init()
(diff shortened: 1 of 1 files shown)


jaylfc:
Holding this for a boot-verified review before merge — it deletes 106 lines from the most critical boot path (`lifespan`), and 14 of the removed `app.state.X = store` assignments are actively read via `request.app.state.X` by routes (decisions, user_shares, routines, feedback, client_log, install_registry, store_submissions, cluster_pairing, agent_model_keys, project_invites, etc.). A wrong call here is a runtime `AttributeError` on those endpoints, not a no-op.

Static analysis looks like the claim holds: the count=3 attrs (decision_store/shared_docs_store/coding_session_store) are re-assigned later in `lifespan` itself (~L849), and the count=2 attrs are assigned unconditionally in `create_app` (~L1619) at construction time, before any request. **But** please confirm with an actual boot: start the app and hit one endpoint per removed attr (at minimum `GET/POST /api/decisions`, `/api/shares`, `/api/routines`, `/api/feedback`) and show they don't 500 with `AttributeError: ... has no attribute 'decision_store'`. The `create_app` block runs before `init()`, so also confirm no reader depends on the store being initialized at the moment it reads `app.state` (init mutates in place, so this should be fine — just verify). Once you've shown a clean boot + those endpoints respond, happy to merge. Zero rush — this is a cleanup, not a fix.

PR #1918, opened 2026-07-17
feat(agents): invite external agent with no project + name/alias (#1913)

### LINKED ISSUE #1913: Invite external agent: make project optional + set name/alias (mint OS-level invite link)
From live use. Two changes to the Agents-app 'Invite external agent' flow:

1. **Project should be optional.** Today `InviteExternalAgentPicker` (desktop/src/apps/agents/RegistryPanel.tsx) forces a project before it opens `InviteAgentDialog`, which mints a project-scoped invite via `POST /api/projects/{id}/invites`. Users want to add an external agent with **no project** — it becomes available in chat immediately, and projects are assigned later via the existing 'Assign to project' dialog.
2. **Set a name/alias** at add-time.

**Decision (owner):** no-project add still **mints a redeemable invite link** (URL+PIN), just not tied to a project — same external-agent security model, project-optional. The remote agent redeems it to get its identity/token and appears in chat.

**Backend:** `POST /api/agents/registry/register` already takes optional `project_id` and `display_name`; `PATCH /api/agents/registry/{id}` already edits display_name; grants can be global (project_id=None). Needs an OS-level invite endpoint (or extend the invite flow) so an invite can be minted without a project — mint identity + global (chat) availability, no project grant.

**Frontend:** replace the forced project-picker with one 'Add external agent' form: Name/alias field + optional Project select defaulting to 'None — available in chat, assign later' + Mint invite → show URL+PIN. Keep the portal + z-[10001] (already correct).

Relates to [[external-agent-auth-consent]], multi-project agent identities (#1862).

jaylfc:
Closes #1913.

## What

Makes the "Invite external agent" flow project-optional and adds a name/alias. When no project is chosen the agent becomes chat-available immediately (projects assigned later via the existing Assign-to-project dialog), while still minting a redeemable URL+PIN invite link under the same external-agent security model.

## Backend

- **`invite_store`**: `mint()` accepts `display_name` and no longer force-appends `project_tasks` for project-less invites. The pending cap now uses an `IS NULL` branch so it stays enforced for OS-level invites (SQL `= NULL` never matches). Adds a `display_name` column with a boot migration for legacy DBs, plus `list_os_level()`.
- **`project_invites` routes**: new admin-gated `POST` / `GET` / `DELETE /api/agents/invites`. The shared redeem endpoint (`POST /api/projects/invites/redeem`) branches on a null `project_id` to mint a chat-available identity with **global** grants and **no** project membership. A dedicated OS-level connection bundle + guide is returned (a2a bus surface only, `project: null`).
- **`approve_request_record`**: optional `display_name` override so the human alias is preserved even when the handle is slugified/deduped.
- **Router order**: `project_invites` is registered before the `agents` router so the literal `/api/agents/invites` paths win over `/api/agents/{name}`.

## Frontend

- The invite entry is now one form: a **Name / alias** field + an optional **Project** select defaulting to *"None — available in chat, assign to projects later"*. "None" mints via the new OS-level endpoint and shows the URL+PIN result inline; choosing a project keeps the existing project-scoped mint dialog. Portal + `z-[10001]` preserved.

## Security notes

- PIN / expiry / attempt-cap / single-use are all still enforced by `store.redeem()` before any branch runs; the OS-level path only differs in that the grant is global (not project-scoped) and no membership row is written.
- The OS-level mint/list/revoke routes are **admin-only** (there is no per-resource owner to check against, matching `mint-internal` / `seed-internal`). The OS-level revoke route refuses to act on project-scoped invites.
- OS-level scopes are never widened to `project_tasks`, so `approve_request_record`'s `needs_project` guard is not triggered and no project binding is fabricated.

## Tests

- Python: `tests/projects/test_invite_store.py` (+6) and `tests/test_routes_project_invites.py` (+8) cover project-less mint (no `project_tasks`, pin returned), independent pending cap, list/revoke, admin gate, redeem to a chat-available identity with global-only grants + no membership, display_name application + handle slugify/fallback, and the advert page. Existing project-scoped tests unchanged and green.
- Vitest: `RegistryPanel.invite-assign.test.tsx` (+2) covers minting an OS-level invite with the alias (URL+PIN result) and chaining to the project dialog when a project is chosen.
- `desktop` `tsc -b` clean.

## Summary by CodeRabbit

* **New Features**
  * Added project-less agent invites that can be created with an optional display name.
  * Admins can create, view, revoke, and redeem OS-level invites.
  * Invite results now display a redeemable URL, one-time PIN, and copyable setup instructions.
  * Redeemed OS-level invites provide global access without project membership.
* **Improvements**
  * Project selection now defaults to “None” instead of automatically choosing a project.
  * Display names are preserved during invite redemption.
  * Added validation, single-use redemption, and pending-invite limits for OS-level invites.

### DIFF
--- a/desktop/src/apps/agents/RegistryPanel.tsx
+++ b/desktop/src/apps/agents/RegistryPanel.tsx
@@ -414,6 +414,16 @@ function GovernanceAuditPanel({ isAdmin }: { isAdmin: boolean }) {
 /*  InviteExternalAgentPicker                                           */
 /* ------------------------------------------------------------------ */
 
+/** OS-level (project-less) mint result: the redeemable URL + PIN. */
+interface OsInviteResult {
+  invite_id: string;
+  pin: string;
+}
+
+function osInviteUrl(inviteId: string): string {
+  return `${window.location.origin}/i/${inviteId}`;
+}
+
 function InviteExternalAgentPicker({
   onCancel,
   onPick,
@@ -424,7 +434,11 @@ function InviteExternalAgentPicker({
   const [projects, setProjects] = useState<import("@/lib/projects").Project[]>([]);
   const [loading, setLoading] = useState(true);
   const [err, setErr] = useState<string | null>(null);
+  const [alias, setAlias] = useState("");
+  // "" is the default "None" option: available in chat, assign to projects later.
   const [selected, setSelected] = useState("");
+  const [submitting, setSubmitting] = useState(false);
+  const [result, setResult] = useState<OsInviteResult | null>(null);
 
   useEffect(() => {
     let active = true;
@@ -436,7 +450,6 @@ function InviteExternalAgentPicker({
         if (!active) return;
         const list: import("@/lib/projects").Project[] = Array.isArray(rows) ? rows : [];
         setProjects(list);
-        if (list.length > 0) setSelected((prev) => prev || list[0]!.id);
       })
       .catch((e: unknown) => {
         if (active) setErr(e instanceof Error ? e.message : "Failed to load projects");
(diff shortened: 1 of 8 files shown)


jaylfc:
Rebased onto dev (clears the stale conftest `_noop_verify_csrf` CI failure) and folded both Kilo findings:
- WARNING: `mint_os_invite` now strips project-bound scopes (`project_tasks`, `canvas_read`, `canvas_write`) before minting, so an OS-level invite can never hand out authority that resolves to no project. Added `_PROJECT_SCOPED` + a test asserting they are dropped.
- SUGGESTION: `display_name` gets `max_length=64` + a validator that trims, rejects control chars (422), and collapses all-whitespace to None. Test added.
27/27 invite tests green locally.

PR #1921, opened 2026-07-17
feat(agents): request additional scopes for an existing agent identity

### LINKED ISSUE #1920: Let an existing registry agent request additional scopes on its own identity
## Problem

The external-agent consent flow (`POST /api/agents/auth-requests` -> `/approve`) MINTS A NEW identity on approval and 409s on an active-handle collision. An agent that ALREADY holds an active registry identity therefore cannot use it to gain more scopes without duplicating itself.

## Proposal

Add a "request more scope for an existing agent" flow that adds grant(s) to the EXISTING canonical_id (never a new identity):

- `POST /api/agents/registry/{canonical_id}/scope-requests` `{requested_scopes, project_id?, reason?}` - create a pending scope request. Verify the canonical_id exists and is active; validate requested_scopes as a subset of the closed VALID_SCOPES vocabulary; persist a pending record; notify the owner/admin.
- `POST /api/agents/registry/{canonical_id}/scope-requests/{req_id}/approve` `{granted_scopes, project_id?}` - owner/admin only; add_grant per granted scope (idempotent), mark accepted. Admin may narrow but never widen the requested scopes.
- `POST /api/agents/registry/{canonical_id}/scope-requests/{req_id}/deny` - owner/admin only.

## Security

Because the agent already has credentials, creation is gated to the agent's OWN registry bearer token (sub == canonical_id) OR an owner/admin session/local token - never an anonymous caller (unlike the new-agent auth-request, which is unauthenticated because the agent has no creds yet). Approve/deny are owner/admin only.

## Notes

- Reuse existing scopes only; VALID_SCOPES stays in sync with _ALLOWED_SCOPES.
- decisions_read / decisions_write may be granted globally (project_id=None) or per-project; project_tasks and canvas scopes still require an explicit project_id.
- Grants are idempotent via UNIQUE(canonical_id, scope, project_id), so re-approving is a no-op.

jaylfc:
Fixes #1920.

## What

Adds a scope-request flow so an ALREADY-REGISTERED agent can gain additional scope grants on its SAME canonical_id. The existing auth-request flow (`POST /api/agents/auth-requests` -> `/approve`) mints a NEW identity on approval and 409s on an active-handle collision, so an agent that already holds an active identity could not use it to request more scope without duplicating itself.

## Endpoints (routes/agent_auth_requests.py)

- `POST /api/agents/registry/{canonical_id}/scope-requests` `{requested_scopes, project_id?, reason?}` - create a pending request.
- `POST /api/agents/registry/{canonical_id}/scope-requests/{req_id}/approve` `{granted_scopes, project_id?}` - approve, owner/admin only.
- `POST /api/agents/registry/{canonical_id}/scope-requests/{req_id}/deny` - deny, owner/admin only.

## Auth gates (security-critical)

- **Create**: allowed for the agent's OWN registry bearer token (JWT `sub` == path `canonical_id`) OR the owning user / an admin (session or admin local token). An anonymous caller can never escalate an existing identity - unlike the new-agent auth-request, which is unauthenticated only because the agent has no creds yet. The middleware allowlist (`auth_middleware.py`) exposes ONLY the create path to a registry JWT (anchored regex, so the approve/deny subpaths are excluded); the route then re-checks identity == canonical_id, so an agent can only self-request.
- **Approve / deny**: owner/admin only (`require_owner_or_admin` against the agent's registry-record owner). An agent's own token cannot reach these - the middleware does not pass a registry JWT through to them, so it falls to the session gate and is rejected.

## Behaviour

- Approval writes `add_grant(canonical_id, scope, project_id)` per granted scope. No new identity is registered. The admin may narrow but never widen the requested scopes.
- Idempotent via `UNIQUE(canonical_id, scope, project_id)`: re-approving an already-granted scope is a no-op.
- `decisions_read` / `decisions_write` (and other global scopes) may be granted globally (`project_id=None`) or per-project; `project_tasks` and canvas scopes still require an explicit `project_id`.
- Creation surfaces a bell notification (`source: agent_scope_requests`) to the owner/admin, retired on decision.
- Requested scopes are validated against the closed `VALID_SCOPES` vocabulary; `test_allowed_scopes_matches_canonical_valid_scopes` still passes.

## New pieces

- `AgentScopeRequestsStore` (parallel to `AuthRequestsStore`; pending -> accepted|refused, atomic conditional-UPDATE decision).
- `check_agent_identity` in `agent_token_auth.py`: scope-agnostic identity verify (proves who, not what), so an agent does not need a scope it lacks in order to request one.

## Tests

`tests/test_agent_scope_requests.py` (14 tests): create + owner-approve lands a grant on the existing canonical_id with the identity count unchanged; agent self-request via own token; agent cannot request for another identity; unknown / inactive id 404; scope outside VALID_SCOPES 400; granted must be subset of requested; global decisions_write grant; project scope requires project_id; non-owner cannot approve; agent cannot approve its own request; idempotent re-approve; deny + conflict.

Local: `pytest tests/ --ignore=tests/e2e -k "scope_request or auth_request or internal_mint or registry"` = 469 passed. New file = 14/14 passed.

## Summary by CodeRabbit

- **New Features**
  - Added a scope-request workflow for already-registered (active) agents to request additional permissions on the same identity.
  - Added endpoints to create, approve, and deny scope requests, with request notifications and retirement after a decision.
  - Added persistent scope-request storage and enhanced registry token verification.
- **Bug Fixes**
  - Enforced ownership/admin authorization for approvals, idempotent approvals, correct project-scoped vs global scope rules (including required project identifiers), and tighter input/rate validation.
- **Documentation**
  - Documented the end-to-end scope-request and scope/project validation behavior.
- **Tests**
  - Added end-to-end coverage for the scope-request flow, including edge cases and error handling.

### DIFF
--- a/docs/agent-coordination.md
+++ b/docs/agent-coordination.md
@@ -145,6 +145,36 @@ further project via `POST /api/projects/{project_id}/members/assign-agent`
 active identity (the existing canonical_id and token are reused instead of
 409ing).
 
+## Requesting more scope for an existing identity (scope requests)
+
+The auth-request flow (`POST /api/agents/auth-requests`) MINTS A NEW identity on
+approval, so an agent that already holds an active registry identity cannot use
+it to gain more scopes without duplicating itself. A scope request adds grants to
+that SAME canonical_id instead:
+
+- `POST /api/agents/registry/{canonical_id}/scope-requests`
+  `{requested_scopes, project_id?, reason?}` — create a pending request. Unlike
+  the new-agent auth-request (unauthenticated, since the agent has no creds yet),
+  this is CREDENTIALLED: the caller must be the agent's OWN registry bearer token
+  (`sub` == `canonical_id`) OR the owning user / an admin. An anonymous caller can
+  never escalate an existing identity. The middleware allowlist exposes only the
+  create path to a registry JWT; the route re-checks identity == canonical_id.
+- `POST /api/agents/registry/{canonical_id}/scope-requests/{req_id}/approve`
+  `{granted_scopes, project_id?}` — owner/admin only. The admin may narrow but
+  never widen the requested scopes; each granted scope is added via
+  `add_grant(canonical_id, scope, project_id)` (idempotent on the
+  `(canonical_id, scope, project_id)` UNIQUE key, so re-approving is a no-op). No
+  new identity is created.
+- `POST /api/agents/registry/{canonical_id}/scope-requests/{req_id}/deny` —
+  owner/admin only.
+
+Requested scopes are validated against the same closed `VALID_SCOPES` vocabulary
+as the consent flow. `project_tasks` and the canvas scopes still require an
+explicit `project_id`; `decisions_read` / `decisions_write` (and the other global
+scopes) may be granted globally (`project_id=None`) or per-project. Creation
+surfaces a bell notification (`source: agent_scope_requests`) to the owner/admin,
+retired when the request is decided.
+
 ## Project invite redeem route (link + PIN)
 
 A project invite lets an external agent join without going through the consent
--- a/tests/test_agent_scope_requests.py
+++ b/tests/test_agent_scope_requests.py
@@ -1,0 +1,402 @@
+"""Scope-request flow: an EXISTING registry agent asks for MORE scopes on its
+own identity, the owner/admin approves, and the grant lands on the SAME
+canonical_id (no new identity is minted).
+
+Mirrors the auth-request route tests (tests/test_routes_agent_auth_requests.py):
+each test wires fresh registry / grants / scope-request stores + keypair onto
+app.state via monkeypatch, so the flow is exercised end to end through the real
+routes and middleware.
+"""
+from __future__ import annotations
+
+import pytest
+import pytest_asyncio
+from httpx import ASGITransport, AsyncClient
+
+from tinyagentos.agent_registry_store import (
+    AgentRegistryStore,
+    load_or_create_signing_keypair,
+    mint_registry_token,
+)
+from tinyagentos.agent_grants_store import AgentGrantsStore
+from tinyagentos.agent_scope_requests_store import AgentScopeRequestsStore
+
+
+class _Env:
+    """Fresh stores + a registered active agent, wired onto app.state."""
+
+    def __init__(self, registry, grants, scope_store, priv, pub, admin_uid):
+        self.registry = registry
+        self.grants = grants
+        self.scope_store = scope_store
+        self.priv = priv
+        self.pub = pub
+        self.admin_uid = admin_uid
+
+    def agent_token(self, canonical_id: str, framework: str = "claude") -> str:
+        return mint_registry_token(
+            canonical_id, self.priv, user_id=self.admin_uid, framework=framework
+        )
+
+    async def close(self):
+        await self.registry.close()
+        await self.grants.close()
+        await self.scope_store.close()
+
+
+async def _wire(client, monkeypatch, tmp_path, *, owner_uid=None):
+    """Build fresh stores, register one ACTIVE agent, monkeypatch app.state."""
+    app = client._transport.app
+    admin = app.state.auth.find_user("admin")
+    admin_uid = admin["id"] if admin else ""
+    owner_uid = owner_uid if owner_uid is not None else admin_uid
+
+    registry = AgentRegistryStore(tmp_path / "reg.db")
+    await registry.init()
+    grants = AgentGrantsStore(tmp_path / "grants.db")
+    await grants.init()
+    scope_store = AgentScopeRequestsStore(tmp_path / "scope.db")
+    await scope_store.init()
+    priv, pub = load_or_create_signing_keypair(tmp_path / "keys")
+
+    monkeypatch.setattr(app.state, "agent_registry", registry)
+    monkeypatch.setattr(app.state, "agent_grants", grants)
+    monkeypatch.setattr(app.state, "agent_scope_requests", scope_store)
+    monkeypatch.setattr(app.state, "agent_registry_keypair", (priv, pub))
+
+    env = _Env(registry, grants, scope_store, priv, pub, admin_uid)
+    env.owner_uid = owner_uid
+    return env
+
+
+async def _register_active(env, *, handle="@worker", display="worker", framework="claude"):
+    rec = await env.registry.register(
+        framework=framework,
+        display_name=display,
+        user_id=env.owner_uid,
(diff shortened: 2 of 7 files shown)


hognek:
## CI fix: rebased onto current dev

**Root cause:** The branch `feat/agent-scope-requests` was 20+ commits behind `origin/dev`. The `create_app()` lifespan in `app.py` had diverged, causing the CI test (3.12) failure. The test failure was likely from merge conflicts during CI checkout or from integration breakage between the PR changes and newer upstream commits.

**Fix:** Rebased `39bc29d4` onto current `origin/dev` (`0e6e6fab`). Resolved one conflict in `tinyagentos/app.py` — the lifespan function now correctly includes:
```python
app.state.auth_requests = auth_requests_store
await agent_scope_requests_store.init()
app.state.agent_scope_requests = agent_scope_requests_store
```

**Tests verified:**
- 14/14 scope-request tests pass (serial)
- 31/31 auth-request tests pass (serial)
- Broader related suite (118 tests across auth_requests, scope_requests, registry, token_auth, internal_mint): 109/118 pass before timeout, no failures seen

**Rebased branch:** `hognek/tinyagentos:feat/agent-scope-requests` (commit `78908cda`)

To apply: force-push this commit to the PR branch, or merge the fix into this branch.

jaylfc:
Self-review before merging, and the headline is that this does not yet deliver the capability it promises. **HOLD.**

**The auth model is correct and I would keep it as-is.** There is no self-approval path: the middleware allowlist regex is `$`-anchored so `/scope-requests/{id}/approve` cannot match an agent token, and approve/deny additionally require a session with `require_owner_or_admin`. Two independent layers, with cross-agent request (403) and cross-agent approve (404) both tested. `check_agent_identity` proving identity without demanding a scope is the right call for a request-for-scope endpoint.

**But merging this today would not let Jay approve task creation for an agent**, which is the actual reason it is being looked at now. Two independent blocks, both present on this head and on current dev:

1. `routes/projects.py:532-541` — `create_task` uses `_get_owned_project` (owner or admin) and never consults grants. That is #2095.
2. `auth_middleware.py:51-61` — `_AGENT_TASK_ROUTES` deliberately omits `POST /api/projects/{pid}/tasks`, with a comment on line 45 naming it as session-only. An agent JWT cannot reach the handler at all.

So #2095 needs widening: it is not one route oversight, the middleware allowlist encodes the same exclusion and both layers must change. I have noted that there.

**Correcting my own earlier claim:** `project_tasks` is not a dead scope. `_authorize_task_actor` is wired into ten routes (list, get, ready, context, comments, claim, release, close, reopen) and `check_agent_scope_for_project` does real grant-gated enforcement. `decisions_write` also reaches a live enforcement point. Task **creation** is the specific hole.

**Findings to fix before merge:**

1. **Privilege escalation via an agent-controlled authorization parameter.** For non-project scopes the approve path falls back to the **agent-supplied** `project_id` when the approver omits one, and that project is never validated to exist or to belong to the approver. Reproduced: agent requests `decisions_write` with `project_id: prj-not-validated`, admin approves with no project_id, and the grant is written bound to the agent's string. Per `decisions.py:154-158` a project-bound `decisions_write` grant resolves the decider to that project's owner, so an agent can name someone else's project and gain the ability to post into that owner's inbox, with nothing in the approval UI hinting at it. Either require an explicit `body.project_id` for any project binding, or validate against a project the approver owns. Untested either way.

2. **Grants are written before the decision flip, with no rollback.** `add_grant` / `add_agent_to_project` / `set_permission` all run before `store.set_decision(...)`. If that returns `None` (concurrent decision) the route raises 409 having already granted the scopes, so the agent keeps permissions from a request recorded as decided by someone else. Flip the decision first, or take `_get_approve_lock` the way the sibling approve path does.

3. **No UI, so the flow is unapprovable from the desktop.** `NotificationCentre.tsx:42` gates approve/deny rendering on `source === "auth_requests"`; this emits `source: "agent_scope_requests"`, so the bell shows a plain info line and approving requires hand-rolled curl with a session cookie. That fails the no-terminal-for-end-users bar and should either land here or be an explicit tracked follow-up.

4. **Existence disclosure**: the authorize check runs after the registry lookup and scope validation, so an agent can distinguish "does not exist" (404) from "exists, not yours" (403) and enumerate the scope vocabulary from a 400 body. Move the authorize call ahead of the lookup's error branch.

5. **Test gaps.** No test asserts that an approved grant actually unlocks anything — a single end-to-end assertion on a task route would have surfaced #2095 before this PR existed, and that is the test that matters most. Finding 1 is untested. `test_agent_cannot_approve_its_own_request` asserts `status_code in (401, 403)`, a tolerance assertion (pitfall 20); the follow-up assertions do prove the intent, but pin the code. Deny has no agent-token negative test. App wiring is never exercised.

**Clean:** store discipline is fine (new table in a new DB file, so SCHEMA indexes are correct and no retrofit is owed), `set_decision` uses a conditional UPDATE with a rowcount check, and `VALID_SCOPES` matches `_ALLOWED_SCOPES` exactly with no drift.

**Also:** this is CONFLICTING against dev (mechanical, `app.py` lifespan wiring only) and about 20 commits behind. hognek pushed a rebase to his fork on 2026-07-18 that was never force-pushed here. The two red test matrices are **base noise, not this PR** — the failures are `_noop_verify_csrf()` argument errors in unrelated test files. `tests/test_agent_scope_requests.py` passes 14/14 on this head. CodeRabbit never actually reviewed it (rate limited), so its green is fake.

Path: fix 1, 2 and 4, add the enforcement-reaches-reality test, apply the rebase, and either add the NotificationCentre source branch or track the UI explicitly. The PR text should also stop implying `project_tasks` yields task creation until #2095 lands.

jaylfc:
Bot findings folded on the latest head:

- Kilo (approve/deny lock): approve_scope_request already takes `_get_approve_lock`; added the same per-request lock to deny_scope_request so a concurrent approve+deny of the same request is serialized (1389f1f4).
- CodeRabbit (authorize before the registry 404 on approve/deny): the owner check is `require_owner_or_admin(user, record["user_id"])`, which needs the record, so it cannot run before the record fetch. create_scope_request is the same: it fetches the record and 404s on unknown before authorizing (the create reorder moved authz before the SCOPE-VOCABULARY check, not before the record existence check). The 404-unknown vs 403-not-owner distinction is the file-wide pattern and matches the sibling consent approve/deny, so making only these two routes existence-hiding would introduce a new inconsistency rather than remove one. Both paths are session-gated. Leaving as-is for consistency; happy to switch the whole agent-registry surface to existence-hiding 404 as a separate pass if preferred.

PR #1925, opened 2026-07-17
fix(auth): XSS hardening, session race protection, and User-Agent binding

### LINKED ISSUE #646: Auth: brute-force protection, password policy, hashing, XSS, session race
## Context

Several authentication hardening opportunities:

1. **No brute-force protection on login** (`routes/auth.py:264-329`) — login POST is not rate-limited
2. **Minimum password length = 4** (`auth.py:299`) — very short
3. **Password hashing uses single-round SHA256** (`auth.py:74-78`) — fast, GPU-brutable. `argon2-cffi` is already a dependency!
4. **Session file race condition** (`auth.py:30-68`) — concurrent async requests can lose session updates
5. **Session tokens not bound to client** (`auth.py:458-466`) — stolen cookies work from any IP
6. **Reflected XSS in auth pages** (`routes/auth.py:119,128,163`) — `error` and `next_url` query params interpolated into HTML without escaping

## Suggested Approach

1. Add login rate limiting: max 5 attempts per IP per minute
2. Increase minimum password to 8 characters
3. Switch to argon2id (argon2-cffi is already installed!)
4. Use `fcntl.flock()` or `asyncio.Lock` for session file access
5. Bind sessions to User-Agent hash as a basic check
6. Add `html.escape()` to the 3 HTML template locations in `routes/auth.py`

hognek:
## Summary

Addresses remaining items from #646 that weren't yet implemented:

1. **XSS hardening** — Add `html.escape()` to `error` and `next_url` parameters in `_login_page` and `_setup_page` template functions (defense-in-depth; route handlers already sanitize these inputs)
2. **Session race protection** — Add `threading.Lock` to `_PersistentSessions` to prevent lost updates from concurrent read-modify-write cycles on the sessions JSON file
3. **User-Agent binding** — Sessions now store a SHA-256 hash of the User-Agent header; `validate_session()` checks it when both the session has a hash and the caller supplies a User-Agent. Backward compatible: sessions without a hash (legacy) and calls without User-Agent (internal lookups) skip the check

## Changes

- `tinyagentos/auth.py`: Lock on `_PersistentSessions`, UA hash in `create_session`/`validate_session`
- `tinyagentos/routes/auth.py`: `html.escape()` in templates, pass `user_agent` to all `create_session()` calls
- `tinyagentos/auth_middleware.py`: Pass `user_agent` to `validate_session()`
- `tests/test_auth.py`: 9 new tests (TestXSSHardening, TestPersistentSessionsLock, TestSessionClientBinding)

## Test results

```
127 passed in tests/test_auth.py tests/test_auth_middleware.py tests/test_auth_email_login.py
```

Fixes #646

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -1009,3 +1009,122 @@ class TestConcurrentHashUpgrade:
         stored = json.loads(mgr._user_file.read_text())
         final_hash = stored["users"][0]["password_hash"]
         assert final_hash.startswith("$argon2"), f"Hash was not upgraded: {final_hash[:40]}"
+
+
+class TestXSSHardening:
+    """html.escape() is applied to error and next_url in auth page templates."""
+
+    def test_login_page_escapes_error_param(self):
+        from tinyagentos.routes.auth import _login_page
+        html = _login_page(error='<script>alert(1)</script>')
+        assert "<script>" not in html
+        assert "&lt;script&gt;" in html
+
+    def test_login_page_escapes_next_url(self):
+        from tinyagentos.routes.auth import _login_page
+        html = _login_page(next_url='"/><script>alert(1)</script>')
+        assert "<script>" not in html
+        assert "&lt;script&gt;" in html
+
+    def test_setup_page_escapes_error_param(self):
+        from tinyagentos.routes.auth import _setup_page
+        html = _setup_page(error='<img src=x onerror=alert(1)>')
+        assert "<img" not in html
+        assert "&lt;img" in html
+
+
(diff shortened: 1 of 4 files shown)


jaylfc (approved):
Reviewed: XSS escaping (html.escape quote=True), the per-instance session Lock around the read-modify-write cycles, and the UA-binding are sound and backward-compatible. UA binding is documented defense-in-depth (bypassable by omitting the header, and enforced only at the middleware gate) which is fine for its stated intent. Merging.

PR #1926, opened 2026-07-17
feat(browser): default agent sessions to mobile viewport flag (latent, not live)

### LINKED ISSUE #635: Agents default to mobile browser view (cleaner DOM) with live desktop fallback
## Why
Mobile site layouts are typically far less noisy than desktop — no sidebars/hover menus, fewer ads/popups/interstitials, single-column DOM. For browser-driving agents that means fewer mis-clicks and "couldn't find the element" failures (the dominant web-agent failure mode), and lighter rendering on the Pi when several agent sessions run at once.

## Scope
- Default `owner_type=agent` browser sessions to **mobile** presentation (`is_mobile=True`); user sessions keep adapting to the connecting device.
- **Per-agent setting** (pin a default) + **per-task override**.
- **Live desktop fallback:** the agent can flip a running session to desktop without a relaunch via the CDP `Emulation.*` layer — try mobile, fall back to desktop when a needed control isn't present on the mobile layout.
- Surface the toggle in the agent's browser settings.

## Depends on
- The CDP input layer (#632) — provides runtime `Emulation.setDeviceMetricsOverride` / `setUserAgentOverride` so mobile↔desktop is a live toggle, not a container relaunch.
- The per-session `is_mobile` field (already shipped in #633).

## Caveat (motivates the fallback)
Mobile sites hide features behind hamburger menus, and some desktop-only flows are absent on mobile — so the default must be switchable and the agent should be able to escalate to desktop on demand.

hognek:
Task: t_86558074. Fixes #635.

## Summary
Sets the default `is_mobile` flag for new agent sessions in the database. **This is a latent default only** — actual mobile rendering does not work yet because `is_mobile` is not read when the container starts. Activation at start remains a follow-up.

The rationale: mobile site layouts are typically far less noisy than desktop — no sidebars/hover menus, fewer ads/popups, single-column DOM. This reduces mis-clicks and "could not find the element" failures (the dominant web-agent failure mode).

The flag is stored in the session row but every start path in `routes/browser_sessions.py` currently derives `mobile` from the request `device` param, not from the session row. A follow-up is needed to wire `is_mobile` into `runner.start(mobile=...)`.

User sessions keep the existing desktop default. Explicit `mobile=True`/`False` overrides both.

## Changes
- `browser_sessions.py`: `create_session()` resolves `mobile=None` to `owner_type=="agent"` (True) vs `"user"` (False). `migrate_agent_browsers()` now sets `is_mobile=1`.
- Tests: 4 new tests for agent/user mobile defaults + explicit override, and `migrate_agent_browsers` `is_mobile` assertion.

## Verification
- ✓ 48/48 browser session tests pass (`.venv/bin/python -m pytest tests/test_browser_sessions.py -xvs`)
- ✓ All CI checks green (lint, shards, doc-gate, SPA build)
- ✓ No regression to human sessions (user sessions keep desktop default)

### DIFF
--- a/tests/test_browser_sessions.py
+++ b/tests/test_browser_sessions.py
@@ -51,6 +51,8 @@ async def test_migrate_agent_browsers_idempotent(mgr):
     sessions = await mgr.list_sessions("agent", "agent-A")
     assert {s["profile_name"] for s in sessions} == {"default", "work"}
     assert all(s["status"] == "stopped" for s in sessions)
+    # Agent sessions default to mobile viewport (cleaner DOM, fewer mis-clicks)
+    assert all(s["is_mobile"] is True for s in sessions)
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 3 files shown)


jaylfc:
Not merging yet. The core change is correct and I confirmed no regression to human sessions, but two things need fixing first, both verified against the code.

**1. Dead config field.** `config.py:65` adds `browser: {"agent_viewport": "mobile"}` (serialized at :87, loaded at :198), but `agent_viewport` is **never read to drive behavior** anywhere. `create_session` hardcodes `owner_type == "agent"` and ignores config, so flipping `agent_viewport` to "desktop" does nothing. That is a misleading config surface, and CLAUDE.md says no configurability that was not requested. Either wire it into `create_session` (use the config value instead of the hardcoded literal) or drop the `browser` config field and keep the hardcoded default. Dropping it is the simpler correct move unless the knob is a stated requirement of the parent issue.

**2. The mobile default is latent, not live.** The stored `is_mobile` column is never read at container-start time. Every start path in `routes/browser_sessions.py` derives `mobile` from the request `device` param (lines 229/270/281), not from the session row. So `is_mobile=True` on an agent session does NOT produce the mobile UA/viewport container the PR body describes; no path feeds the stored column into `runner.start(mobile=...)`. Not a live bug, since agent rows currently only arrive via migration, but the PR summary reads as if mobile rendering is active when it is not.

**To land this as the incremental step it actually is:** drop the dead config field, and reword the description to "sets the default flag" rather than implying mobile rendering works now. Then the eventual agent-start flow needs a follow-up that actually reads `is_mobile` (or derives `mobile` from `owner_type`) or this default never renders.

The `create_session` logic itself is right and well-tested (agent default, user default, both overrides, migration assertion, 48/48). This is purely about not shipping a knob that does nothing and a claim that overstates the effect.

PR #1928, opened 2026-07-17
feat(cluster): worker registry persistence, split-brain protection, circuit breaker

### LINKED ISSUE #640: Cluster: persistence, split-brain protection, circuit breaker
## Context

Four reliability concerns for multi-node deployments:

**1. In-memory worker registry** (`cluster/manager.py:32`) — `self._workers: dict[str, WorkerInfo] = {}`. Lost on controller restart. Workers must re-register; in-flight routing fails silently during the gap.

**2. No split-brain protection** — No fencing, no generation counter, no quorum. Two controllers can run simultaneously and route independently.

**3. No circuit breaker** (`cluster/router.py:23`) — Tries all workers sequentially. A slow worker (60s timeout) blocks for `60 * N_workers` before failing.

**4. Non-atomic service migration** (`cluster/service_migrator.py`) — Stop source -> tar -> deploy target. Crash between stop and deploy = service down with no auto-recovery.

## Suggested Approach

1. Persist worker registrations to SQLite on heartbeat. On restart, load last-known state and mark stale until re-heartbeat
2. Add a `generation` counter incremented on each controller start; workers reject stale generations
3. Track per-worker failure counts; skip workers failing N times in M seconds
4. Take a snapshot before stopping source; auto-rollback to snapshot if deployment fails

hognek:
Closes #640.

Four interconnected changes for multi-node reliability:

**1. Worker registry persistence** — New `WorkerRegistryStore` (SQLite-backed via `BaseStore`) persists worker registrations so the in-memory `ClusterManager._workers` dict survives controller restarts. On startup, loaded workers are marked `"stale"` until they re-heartbeat. The `"local"` worker is skipped during load (it re-registers on every boot).

**2. Split-brain protection via generation counter** — A generation counter (stored in the same SQLite table) is incremented on each controller `start()`. Workers echo it in heartbeats; stale generations are rejected. Legacy workers that don't send generation get a pass (`None`).

**3. Circuit breaker for failing workers** — New `FailureTracker` with sliding window (default: 3 failures in 60s). `TaskRouter` skips tripped workers instead of serially blocking `60s * N_workers`. Success clears the circuit; failures recorded per-worker.

**4. In-flight routing recovery** — Addressed by Fix 1: persisted workers survive controller restart; stale-until-heartbeat prevents routing to workers that haven't reconnected.

### Files changed
- **NEW**: `tinyagentos/cluster/worker_registry_store.py` — SQLite store for worker records + generation counter
- **NEW**: `tinyagentos/cluster/failure_tracker.py` — Sliding-window circuit breaker
- **MODIFIED**: `tinyagentos/cluster/manager.py` — init/start/heartbeat/register/unregister with persistence, generation counter, failure_tracker wiring
- **MODIFIED**: `tinyagentos/cluster/router.py` — Circuit breaker skip + record on failure/success
- **MODIFIED**: `tinyagentos/app.py` — Wire new stores into ClusterManager creation
- **NEW**: `tests/test_cluster_persistence.py` — 24 tests covering store CRUD, generation counter, stale-on-load, heartbeat revival, split-brain rejection, circuit breaker tripping/skipping/clearing

### Testing
- 55 existing cluster tests + 24 new tests = **79 passed**
- Stores are optional (None = no persistence); all existing tests pass unchanged
- Full gate running in background

### Backward compatibility
- `ClusterManager.__init__` accepts new optional `worker_registry_store` and `failure_tracker` params — defaults to None
- `heartbeat()` accepts new optional `generation` param — None = legacy worker, accepted
- No DB migration needed — new tables created on first `store.init()`
- Router gracefully handles `failure_tracker is None` (no circuit breaker)

## Summary by CodeRabbit

* **New Features**
  * Worker registrations and status are now preserved across controller restarts.
  * Added generation tracking to reject stale worker registrations and heartbeats.
  * Added circuit-breaker protection to skip repeatedly failing workers and resume routing after successful requests.
  * Worker registration and heartbeat requests now support generation information.

* **Tests**
  * Added comprehensive coverage for persistence, split-brain protection, failure tracking, and circuit-breaker routing.

### DIFF
--- a/tests/test_cluster_persistence.py
+++ b/tests/test_cluster_persistence.py
@@ -600,6 +600,48 @@ class TestClusterManagerPersistence:
         await mgr.stop()
 
 
+    async def test_controller_fenced_when_generation_advanced(self, tmp_path):
+        """Controller A at gen 2 fences itself when controller B advances to gen 3."""
+        store = await self._new_store(tmp_path)
+        mgr_a = ClusterManager(worker_registry_store=store)
+        await mgr_a.start()
+        assert mgr_a.generation >= 1
+        gen_a = mgr_a.generation
+
+        # Register a worker so we can verify it gets fenced
+        w = _make_worker("gpu-box", capabilities=["chat"])
+        await mgr_a.register_worker(w)
+        assert mgr_a.get_worker("gpu-box").status == "online"
+
+        # Controller B advances the durable generation past A's
+        await store.increment_generation()
+        durable_gen = await store.current_generation()
+        assert durable_gen > gen_a
+
+        # Simulate the monitor loop's fence detection
+        mgr_a._fenced = False  # reset just in case
+        durable_check = await store.current_generation()
+        if durable_check > mgr_a._generation:
+            mgr_a._fenced = True
+            for worker in list(mgr_a._workers.values()):
+                if worker.name != "local" and worker.status in ("online", "update-available"):
+                    worker.status = "offline"
+
+        assert mgr_a._fenced is True
+        assert mgr_a.get_worker("gpu-box").status == "offline"
+
+        # After fencing, heartbeat should be rejected
+        ok = mgr_a.heartbeat("gpu-box", load=0.1, generation=gen_a)
+        assert ok is False
+
+        # After fencing, new registration should be silently rejected
+        w2 = _make_worker("new-worker", capabilities=["chat"])
+        await mgr_a.register_worker(w2)
+        assert mgr_a.get_worker("new-worker") is None
+
+        await mgr_a.stop()
+        await store.close()
+
 # ── TaskRouter circuit breaker tests (Fix 3) ───────────────────────────
 
 @pytest.mark.asyncio
@@ -631,7 +673,7 @@ class TestTaskRouterCircuitBreaker:
         assert mock_client.post.call_count == 1
 
     async def test_records_failure_on_error(self):
-        ft = FailureTracker(failure_threshold=5)
+        ft = FailureTracker(failure_threshold=1)
         mgr = ClusterManager(failure_tracker=ft)
         await mgr.register_worker(_make_worker("w1", capabilities=["chat"], load=0.1, url="http://w1:8000"))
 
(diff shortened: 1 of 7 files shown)


hognek:
All 5 Kilo findings addressed + rebased onto origin/dev (8b97eb5a):

**WARNING (3 fixed):**
1. `manager.py:271` — Added debug log for None-generation heartbeats from legacy workers (with `self._generation > 1` guard to avoid noise on first boot)
2. `manager.py:109` — register_worker now accepts `generation` parameter + performs the same generation check as heartbeat()
3. `manager.py:391` — Wrapped fire-and-forget `_persist_worker` in try/except with background task tracking (`_background_tasks.add` + `add_done_callback`)

**SUGGESTION (2 fixed):**
4. `failure_tracker.py:26` — Comment now says 3 failures (matches DEFAULT_FAILURE_THRESHOLD)
5. `worker_registry_store.py:178` — Uses `RETURNING value` for atomic increment (fetchone before commit, matching existing RETURNING patterns in message_store.py)

**Routes wired:** `generation` added to WorkerRegister + HeartbeatBody models, passed through to register_worker() and heartbeat()

96/96 cluster/persistence/worker-protocol tests pass.

jaylfc:
Reviewed (deep-dive, all findings file:line-verified) + merging. Highlights: split-brain fencing's primary path is sound — durable atomic generation (`UPDATE...RETURNING`, WAL + busy-timeout) with a bounded ~5s simultaneous-primary window, verified live against a real ClusterManager; merge materially keeps dev's Aug-14 node-revoke work (checked the materialized merge tree, not just the clean-merge signal); breaker is per-worker, time-decaying, can't wedge open.

Three gaps are real and get cards (none architecture-level, all 'coded vs proven/complete'):
1. **Kilo's WARNING is CONFIRMED**: the fenced branch of `_monitor_loop` (manager.py:1000-1007) only flips workers offline and `continue`s — it never releases `self._leases` nor calls `cancel_running_for_leases`, unlike all three sibling termination branches (offline/drain/stuck-updating), and the `continue` also skips the lease sweep. Stranded leases + running arbiter tasks collide with the winning controller. Highest-priority follow-up.
2. The worker-echoed `generation` check (manager.py:110-115, 294-299) is currently dead code — `worker/agent.py` never sends it and the responses never surface it; all real protection rests on the 5s DB poll.
3. The fence test (test_cluster_persistence.py:603-643) manually replicates the loop's conditional instead of driving the real `_monitor_loop` — the create_task+sleep(0.05)+cancel pattern at test_cluster.py:106-129 proves that's trivially testable; also no breaker cooldown-recovery test.

Net: feature adds protection where dev had none; the gaps must not block that but must not be forgotten either.

PR #1930, opened 2026-07-17
fix(notifications): persist archive to backend so it survives reload

### LINKED ISSUE #1916: Notification archiving does not persist across reload
## Summary

Archiving or dismissing a server-origin notification in the desktop PWA does not persist. After closing and reopening the app (or a hard reload), archived notifications reappear in the active inbox.

## Root cause

`desktop/src/stores/notification-store.ts`. The client archive actions `dismiss`, `archiveRead`, and `clearAll` only set a local `archived: true` flag and add the server id to a module-scoped in-memory `Set` (`dismissedServerIds`). They never tell the backend.

`dismissedServerIds` is lost on reload, and the store re-fetches `GET /api/notifications` (the active list), which still returns those rows as active because the server was never informed. Archived notifications therefore return to the inbox.

The backend already exposes `POST /api/notifications/{notif_id}/archive`, which persists the archive and excludes the row from the active feed (covered by `test_notification_api_archive`). The client simply never calls it.

## Expected

Archiving a server notification persists on the backend. After a reload, `GET /api/notifications` no longer returns it and it appears under History (`GET /api/notifications/archived`).

## Scope

Client-origin notifications (`notif-N`, ephemeral session toasts with no server row) remain local-only, since there is no server row to persist.

hognek:
## Summary

Wire the client's notification archive actions (`dismiss`, `archiveRead`, `clearAll`) to call the existing backend endpoint `POST /api/notifications/{id}/archive` for server-origin (`srv-*`) notifications.

## Problem

Archiving or dismissing a server-origin notification in the desktop PWA only set a local `archived: true` flag and added the id to a module-scoped `dismissedServerIds` Set. Neither persisted across a reload, so archived notifications reappeared in the active inbox.

## Root Cause

The client never called `POST /api/notifications/{id}/archive`. The backend endpoint already existed and was fully tested (`test_notification_api_archive`).

## Fix

1. **`server-notifications.ts`** — Added `archiveServerNotification(id)` following the established `markServerRead` pattern. Calls `POST /api/notifications/{n}/archive` with CSRF header, best-effort (swallows failures).

2. **`notification-store.ts`** — `dismiss()`, `archiveRead()`, and `clearAll()` now call `archiveServerNotification()` for `srv-*` ids. Client-origin (`notif-N`) notifications remain local-only — no server row to persist.

## Tests

- **`server-notifications.test.ts`** — 3 new tests for `archiveServerNotification`: POSTs the correct endpoint, no-op for client ids, swallows fetch failures
- **`notification-store.test.ts`** — 4 new tests verifying the mock is called for srv-* ids on dismiss/archiveRead/clearAll, and NOT called for client-origin ids
- All 53 notification-related tests pass (store, lib, centre, toast)

## Scope

- **Not affected:** Client-origin notifications (`notif-N`), which remain ephemeral
- **Not affected:** The `dismissedServerIds` Set — it still prevents re-adding dismissed items mid-session from SSE polls

Task: t_b2a6a624. Fixes #1916.

### DIFF
--- a/desktop/src/lib/server-notifications.test.ts
+++ b/desktop/src/lib/server-notifications.test.ts
@@ -3,6 +3,7 @@ import {
   fetchServerNotifications,
   markServerRead,
   markAllServerRead,
+  archiveServerNotification,
   sourceToTarget,
 } from "./server-notifications";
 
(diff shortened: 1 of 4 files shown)


jaylfc:
Thanks hognek, this is a solid fix and the code is essentially identical to #1917 (same root cause: the client archive actions never called POST /api/notifications/{id}/archive, so a reload re-fetched them). Closing as a duplicate of #1917, which I am landing, purely to avoid a double-merge. The only differences were tie-breakers: #1917 also updates the now-stale dismissedServerIds comment, and its full CI matrix has run (yours sits in action_required as an external-contributor gate). Appreciate the work, and #1916 will close when #1917 merges.

PR #1931, opened 2026-07-17
feat(settings): add connection-test + reachability reporting to taOSmd memory URL

### LINKED ISSUE #1911: taOSmd memory URL: add connection-test + reachability reporting (GET {url,is_local,reachable})
Follow-up from #1904 (merged, implements #1891 core). The persist/proxy/local-fallback core landed and is green, but one #1891 acceptance criterion is not yet implemented:

- `GET /api/settings/memory-url` should return `{url, is_local, reachable}` (currently returns `{url}` only).
- `PUT /api/settings/memory-url` should probe the target taOSmd (health/stats) so an admin pointing at a dead/typo'd remote host gets immediate feedback instead of silent fallback to local FTS5.
- Add direct route tests for GET/PUT `/api/settings/memory-url` (config-layer + `_taosmd_base` precedence are covered; the endpoints themselves are not).

Parent epic: #1890. Scope: observability half of the remote-memory switch.

hognek:
Closes #1911.

Adds GET /api/settings/memory-url returning `{url, is_local, reachable}` with a health-probe against the configured taOSmd. Adds PUT endpoint that validates the URL, probes the new target, and persists it to `app.state.taosmd_url`.

**Changes:**
- `tinyagentos/routes/settings.py`: Add `_is_local_url`, `_probe_taosmd`, `_taosmd_default_url` helpers + `GET` and `PUT /api/settings/memory-url` endpoints
- Fix `parsed.netloc` → `parsed.hostname` in hostname validation (line 1149) — `http://:7900` was bypassing validation because netloc returns `:7900` (non-empty) while hostname correctly returns empty string
- `tests/test_routes_settings.py`: Added `http://:7900` regression test to `test_put_rejects_url_without_hostname`
- 14 tests covering default URL, local/remote detection, PUT validation (empty/invalid scheme/missing hostname/http://:7900), trailing-slash stripping, URL persistence, and `_is_local_url` unit coverage

Tests: targeted MemoryUrl validation tests (3/3 pass). Full gate run timing out on pre-existing probe-hang issue — CI will run full matrix.

### DIFF
--- a/tests/test_routes_settings.py
+++ b/tests/test_routes_settings.py
@@ -450,10 +450,18 @@ class TestMemoryUrlSettings:
 
     @pytest.mark.asyncio
     async def test_put_rejects_url_without_hostname(self, client):
-        """A URL without a valid hostname must return 400 (netloc check)."""
+        """A URL without a valid hostname must return 400 (hostname check)."""
+        # path-only: http:///path  (hostname is empty string)
         resp = await client.put(
             "/api/settings/memory-url",
             json={"url": "http:///path"},
+        )
+        assert resp.status_code == 400
+
+        # port-only: http://:7900  (hostname is empty, netloc non-empty — regression for netloc→hostname fix)
+        resp = await client.put(
+            "/api/settings/memory-url",
+            json={"url": "http://:7900"},
         )
         assert resp.status_code == 400
 
--- a/tests/test_routes_settings.py
+++ b/tests/test_routes_settings.py
@@ -330,3 +330,153 @@ class TestRebuildResultStructured:
         assert r.rebuilt is True
         assert r.success is False
         assert r.message == "x"
+
+
+class TestMemoryUrlSettings:
+    """Contract: GET /api/settings/memory-url returns {url, is_local, reachable}
+    and PUT probes the new target before returning (#1911)."""
+
+    @pytest.mark.asyncio
+    async def test_get_returns_default_url(self, client):
+        """Default URL is localhost:7900, is_local=True, reachable depends on env."""
+        resp = await client.get("/api/settings/memory-url")
+        assert resp.status_code == 200
+        data = resp.json()
+        assert data["url"] == "http://localhost:7900"
+        assert data["is_local"] is True
+        assert isinstance(data["reachable"], bool)
+
+    @pytest.mark.asyncio
+    async def test_get_returns_is_local_true_for_localhost(self, client, app):
+        """Setting the config URL to a loopback address must report is_local=True."""
+        app.state.config.memory_url = "http://127.0.0.1:7900"
+        resp = await client.get("/api/settings/memory-url")
+        assert resp.status_code == 200
+        data = resp.json()
+        assert data["url"] == "http://127.0.0.1:7900"
+        assert data["is_local"] is True
+
+    @pytest.mark.asyncio
+    async def test_get_returns_is_local_false_for_remote(self, client, app):
+        """A remote URL must report is_local=False."""
+        app.state.config.memory_url = "https://memory.example.com:7900"
+        resp = await client.get("/api/settings/memory-url")
+        assert resp.status_code == 200
+        data = resp.json()
+        assert data["url"] == "https://memory.example.com:7900"
+        assert data["is_local"] is False
+
+    @pytest.mark.asyncio
+    async def test_put_sets_url_and_probes(self, client, app):
+        """PUT persists the URL to config and returns probe result."""
+        resp = await client.put(
+            "/api/settings/memory-url",
+            json={"url": "http://192.168.1.100:7900"},
+        )
+        assert resp.status_code == 200
+        data = resp.json()
+        assert data["url"] == "http://192.168.1.100:7900"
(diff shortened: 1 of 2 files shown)


jaylfc:
Thanks hognek. The reachability code is right (GET returns {url,is_local,reachable}, PUT probes {url}/health with an admin-gated 3s timeout, 10 tests pass), BUT do not merge or plain-rebase as-is: this branch predates #1904, which already shipped a GET+PUT /api/settings/memory-url pair on dev. Merging onto current dev registers the route TWICE (ruff F811 flags both get_memory_url/set_memory_url as redefinitions), and FastAPI serves your first-registered handlers, which makes #1904's config-backed pair dead code. That is a regression: your PUT only sets app.state.taosmd_url in memory and never calls save_config_locked, so the URL stops persisting across restart, and your GET ignores config.memory_url. The green tests miss it because #1904 shipped no route tests.

Fix: rebase onto current dev (b0f39932) and FOLD the reachability logic INTO the existing #1904 endpoints, do not add a second pair. Add is_local/reachable to #1904's GET, add the probe to #1904's PUT that already persists via save_config_locked + the Pydantic MemoryUrlUpdate model. While there, fold the Kilo nits: the docstring says validate-before-accept but it persists even when reachable=False (decide: accept-anyway or 400), and use the Pydantic model instead of the unguarded await request.json() (avoids a 500 on a bad body). Keeping #1911 open until this lands.

hognek:
**Reconciled with #1904** — pushed commit `a728593` with all feedback addressed:

### jaylfc feedback (route duplication)
- Folded reachability logic INTO #1904's existing GET+PUT `/api/settings/memory-url` — no duplicate route registration.
- GET returns `{url, is_local, reachable}` with 30s TTL probe cache.
- PUT validates via Pydantic `MemoryUrlUpdate` model, persists via `save_config_locked`, returns `{url, is_local, reachable}`.

### Kilo 2W+2S (all fixed)
- **WARNING (docstring)**: Docstring now explicitly states URL is always accepted/persisted; `reachable` reports but does not gate the save.
- **WARNING (unguarded json)**: Uses Pydantic `body: MemoryUrlUpdate` — FastAPI handles parsing, 422 on bad JSON.
- **SUGGESTION (rstrip)**: Uses `url.endswith("/")` + `url[:-1]` instead of `rstrip("/")` character-set behavior.
- **SUGGESTION (caching)**: `_MEMORY_URL_PROBE_CACHE` with 30s TTL on GET, populated by PUT.

### Tests
13 tests (10 original + 3 new: malformed JSON 422, `_is_local_url` unit, response-shape assertion). Local pytest blocked by pre-existing aiosqlite conftest issue — CI will run the matrix.

Closes #1911.

jaylfc:
The original hold is resolved, thanks: the reconcile folds into #1904's single GET+PUT pair (no route duplication), PUT now persists via `save_config_locked`, and the four Kilo nits are gone. One real (minor, admin-gated) bug left before merge:

- routes/settings.py:1150: validation checks `parsed.netloc` not `parsed.hostname`, so `http://:7900` passes (non-empty netloc, no hostname), probes unreachable, and is still persisted. Switch to `parsed.hostname` and add a `http://:7900` regression test (the existing test only covers `http:///path`).

Nits (optional): `_MEMORY_URL_PROBE_CACHE` is an unbounded module-level dict with no eviction; the persistence test only asserts the in-memory config value rather than reloading from disk.

jaylfc:
The netloc->hostname fix + the http://:7900 regression test look right, thanks. One nit from my earlier review is still open: `_MEMORY_URL_PROBE_CACHE` is an unbounded module-level dict with no eviction. It only grows by distinct admin-entered URLs so the risk is tiny, but per our fold-every-finding rule please add a small cap (e.g. an OrderedDict with popitem(last=False) beyond ~32 entries, or just clear when it exceeds a bound). Once that is in and CI is green this is good to merge.

jaylfc:
Close - most earlier findings were resolved in your rewrite. Two left: routes/settings.py ~1109 the _MEMORY_URL_PROBE_CACHE module dict grows unbounded (never evicted) on a frequently-polled endpoint - add an eviction/size cap. And confirm the old persist-on-probe-fail vs docstring mismatch survived the rewrite: the docstring says the URL is validated before accepting, but taosmd_url was persisted even when reachable==False - either gate the write on reachable or fix the docstring. Fold + green and I merge.

hognek:
All three findings are now addressed:

1. **`_MEMORY_URL_PROBE_CACHE` cap** (abdabfa6): `OrderedDict` with 32-entry FIFO eviction via `popitem(last=False)` — prevents unbounded growth.
2. **Docstring/behavior match** (a728593): docstring now explicitly states "The URL is always accepted and persisted. The *reachable* field reports the probe result but does not gate the save."
3. **Narrow exception catch** (6736f24): `except Exception` → `except (httpx.RequestError, httpx.InvalidURL)` in `_probe_taosmd` — unexpected failures (e.g. missing `http_client`) now propagate instead of being silently swallowed as "unreachable". Closes the CodeRabbit finding.

Validation tests pass (empty URL, non-http scheme, missing hostname). Ready for review.

Issue #858, opened 2026-06-13
GitHub support: connect a GitHub account in Secrets, share with agents, agents post PRs (GitHub App + injected short-lived token)


jaylfc:
**Goal (Jay):** let a user connect their GitHub account in the **Secrets app** and share it with agents, so e.g. the **taOS agent can open PRs / issues** on the taos repo.

## What already exists (≈60%)
- **Secrets app + store** (`SecretsApp.tsx`, `tinyagentos/secrets.py`, `routes/secrets.py`) with **per-agent grants**: `secret_access(secret_id, agent_name)` + `get_agent_secrets(agent_name)`.
- **GitHub read layer** (`routes/github.py`): `_get_token()` reads `github_token` from the SecretsStore (or `gh auth token`), used for auth/status, repos, issues, notifications, starred.

## Decisions (Jay)
- **Auth = GitHub App** (not a raw PAT). A taOS GitHub App is installed by the user on chosen repos; taOS mints **short-lived, per-repo installation tokens**. Revocable per-repo from GitHub; no long-lived credential.
- **Posting = agent-direct git/gh**: the short-lived token is injected into the agent's container as `GITHUB_TOKEN`; the agent uses its own `git` + `gh` to push a branch and open the PR (works under any framework). Token is short-lived to bound the exposure.

## Gaps to build
1. **taOS GitHub App** registration (App ID + private key) and an **installation/connect flow** in the Secrets app (replaces the raw `github_token` key with a first-class GitHub connection: which repos, which installation).
2. **Per-agent grant UI** wired to `secret_access`: pick which agents may use which repo installation.
3. **Short-lived token minting + injection**: at agent deploy/runtime, mint an installation token scoped to the agent's granted repos and inject as `GITHUB_TOKEN` (with refresh before expiry). Granted secrets are not currently injected into the agent runtime - that wiring is needed.
4. **Agent container has git + gh** + a small ops skill/guidance for the branch -> PR flow.

## Prerequisite (Jay's action)
Register a **taOS GitHub App** under the JAN LABS / jaylfc account (App ID + private key, permissions: contents RW, pull_requests RW, issues RW). taOS stores the App private key as a host secret and uses it to mint installation tokens.

Relates to #454 (agent-facing GitHub events skill + channel adapter). Framework-agnostic per the agent-harness-swappable rule. Needs a spec before build.

jaylfc:
## Refined design (Jay) - OAuth identities + guided, time-scoped sharing

[...]

PR #1932, opened 2026-07-17
feat(github): add GitHub App installation flow (C1)

hognek:
## Problem

taOS supports GitHub OAuth for agent identity, but there is no first-class GitHub App installation path. External repos and organization-level workflows need a cleaner integration than personal OAuth tokens.

## Implementation

- **Backend (tinyagentos/github_app.py):** Generates per-installation GitHub tokens via the App installation API. Handles JWT creation, token exchange, and permission scoping.
- **Store (tinyagentos/github_app_installations.py):** Persistent metadata for installed apps (installation_id, account name/avatar, repo list).
- **Routes:** POST /api/github/installations (start installation), DELETE /api/github/installations/:id (revoke), GET /api/github/installations/:id/repos (list repos).
- **Desktop UI (GitHubConnect.tsx):** New "GitHub App" section in Secrets page with install button, installations list with repo count, and uninstall.
- **Config:** github_app_id, github_app_private_key_path, github_app_webhook_secret — all optional, feature is inert when unconfigured.

## Token Resolution for User-Scoped Endpoints

The `_get_token()` helper in `tinyagentos/routes/github.py` was extended with a `user_scoped` parameter to prevent GitHub App installation tokens from being used on personal user endpoints. Installation tokens authenticate as the *App installation*, not as a *user* — endpoints like `/user/starred`, `/notifications`, and `/auth/status` require a user identity and return empty or misleading results when called with an installation token.

### Token preference order

Three sources are tried in order, but **step 2 is skipped for user-scoped endpoints**:

| Priority | Source | User-scoped | Repo-scoped |
|----------|--------|-------------|-------------|
| 1 | PAT from SecretsStore (user's own GitHub identity) | ✓ tried first | ✓ tried first |
| 2 | GitHub App installation token | ✗ **SKIPPED** | ✓ fallback |
| 3 | `gh auth token` subprocess | ✓ fallback | ✓ fallback |

If all sources are exhausted, the endpoint returns a 401 "No GitHub token available" response. For `/auth/status`, it returns `{"authenticated": false, "source": null}`.

### Affected endpoints

**User-scoped** — `_get_token(request, user_scoped=True)`. The App token gate is skipped; PAT is preferred, `gh` CLI is the fallback:
- `GET /api/github/auth/status` — accurately reports whether a personal token is available
- `GET /api/github/starred?page=N` — lists the user's personal starred repos
- `GET /api/github/notifications` — fetches the user's unread notifications

**Repo-scoped** — `_get_token(request)` (default, `user_scoped=False`). App token is used as a fallback when no PAT is configured:
- `GET /api/github/repo/{owner}/{name}` — repo metadata and README
- All other repo-scoped endpoints (backward-compatible, no behavioral change)

### Design rationale

1. **Correctness:** GitHub App installation tokens cannot access user-scoped GitHub API endpoints. Without this gate, user-scoped endpoints would silently return empty results or 401 errors when only an App token was available.
2. **Honest auth reporting:** `/auth/status` now returns `authenticated=false` when only an App token exists, so the UI can prompt the user to connect a personal account.
3. **Backward compatibility:** Repo-scoped endpoints continue to use App tokens as a fallback, so repos accessible to the App installation still work without a PAT.

### Round 1 (7bb09e6a)
Addressed all 12 Kilo findings + CodeRabbit docstring coverage check:
- W1: Reversed token resolution order (github.py)
- W2: Removed dead /user API fetch (github_oauth.py)
- W3: DELETE removes local record after remote deletion
- S4-S12: Various improvements (private_key check, cache TTL, pagination, async save, content-type guard, etc.)

### Round 2 (cbb61eb7)
- **SPA build fix:** Added missing listAppInstallations/beginAppInstallation/deleteAppInstallation mocks to SecretsApp.test.tsx
- **CodeRabbit N1:** Simplified nested installs_store condition in github_oauth.py
- **CodeRabbit N2:** Added refreshInstallations to useEffect deps + window focus listener in GitHubConnect.tsx
- **Error handling:** Moved generate_jwt inside try blocks in get_installation_token and list_installations
- **Error handling:** Added AttributeError/TypeError to _load except clause in github_app_installations.py

## Testing

- Backend: 70/70 tests pass (64 original + 6 new route-level token resolution tests)
- Desktop: npm run build passes, 10/10 SecretsApp tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GitHub auth fallback so user-scoped endpoints try personal credentials first and fall back to the GitHub CLI when unavailable.
  * User-scoped endpoints (auth status, starred, and notifications) no longer mint GitHub App installation tokens.
  * Auth status now correctly reports unauthenticated when neither personal tokens nor GitHub CLI credentials are available.
  * Repository-scoped endpoints continue using GitHub App installation tokens when personal authentication is missing.
* **Tests**
  * Added route-level coverage for these token-resolution and fallback scenarios.

### DIFF
--- a/tests/test_routes_github.py
+++ b/tests/test_routes_github.py
@@ -331,14 +331,12 @@ async def test_repo_upstream_error_returns_500():
 def _build_app_with_app_config(
     token: str | None = None,
     http_client=None,
-    gh_cli_token: str | None = None,
-    install_token: str | None = "ghs_app-token",
 ):
     """Build a minimal FastAPI app with mock GitHub App configuration.
 
-    - PAT is set via `token` (None = no PAT)
-    - gh CLI subprocess returns `gh_cli_token` (None = fails)
-    - App installation token returns `install_token` (None = fails)
+    - PAT is set via ``token`` (None = no PAT)
+    - Callers should mock ``create_subprocess_exec`` and
+      ``get_installation_token`` directly for gh CLI / App token behaviour
     """
     app = FastAPI()
     app.include_router(github_router)
@@ -381,7 +379,6 @@ async def test_starred_uses_gh_cli_when_app_configured_no_pat():
     app = _build_app_with_app_config(
         token=None,  # No PAT
         http_client=http_client,
-        gh_cli_token="gh-cli-token",
     )
 
     async def _fake_subprocess(*args, **kwargs):
@@ -424,7 +421,6 @@ async def test_notifications_use_gh_cli_when_app_configured_no_pat():
     app = _build_app_with_app_config(
         token=None,
         http_client=http_client,
-        gh_cli_token="gh-cli-token",
     )
 
     async def _fake_subprocess(*args, **kwargs):
@@ -458,7 +454,6 @@ async def test_auth_status_false_when_only_app_token_available():
     app = _build_app_with_app_config(
         token=None,  # No PAT
         http_client=http_client,
-        gh_cli_token=None,  # gh CLI fails
     )
 
     transport = ASGITransport(app=app)
@@ -501,7 +496,6 @@ async def test_repo_endpoint_still_uses_app_token():
     app = _build_app_with_app_config(
         token=None,  # No PAT
         http_client=http_client,
-        gh_cli_token=None,  # gh CLI not available
     )
 
     transport = ASGITransport(app=app)
--- a/tests/test_routes_github.py
+++ b/tests/test_routes_github.py
@@ -322,3 +322,204 @@ async def test_repo_upstream_error_returns_500():
     assert resp.status_code == 500
     data = resp.json()
     assert "error" in data
+
+
+# ---------------------------------------------------------------------------
+# User-scoped token resolution: App token must NOT serve user endpoints
+# ---------------------------------------------------------------------------
+
+def _build_app_with_app_config(
+    token: str | None = None,
+    http_client=None,
+    gh_cli_token: str | None = None,
+    install_token: str | None = "ghs_app-token",
+):
+    """Build a minimal FastAPI app with mock GitHub App configuration.
+
+    - PAT is set via `token` (None = no PAT)
+    - gh CLI subprocess returns `gh_cli_token` (None = fails)
+    - App installation token returns `install_token` (None = fails)
+    """
+    app = FastAPI()
+    app.include_router(github_router)
+
+    # SecretsStore (PAT)
+    mock_secrets = MagicMock()
+    if token:
+        mock_secrets.get = AsyncMock(return_value={"value": token})
+    else:
+        mock_secrets.get = AsyncMock(return_value=None)
+    app.state.secrets = mock_secrets
+
+    # App config
+    mock_config = MagicMock()
+    mock_config.github_app_id = "123456"
+    mock_config.github_app_private_key = "fake-private-key"
+    app.state.config = mock_config
+
+    # App installations store
+    mock_installs = MagicMock()
+    mock_installs.list_all = MagicMock(return_value=[{"installation_id": 42}])
+    app.state.github_app_installations = mock_installs
+
+    app.state.http_client = http_client or MagicMock()
+    return app
+
+
+@pytest.mark.asyncio
+async def test_starred_uses_gh_cli_when_app_configured_no_pat():
+    """When App is configured but no PAT, /starred (user-scoped) skips the
+    App token and falls through to gh CLI."""
+    starred_data = [{
+        "name": "repo1", "owner": {"login": "alice"},
+        "full_name": "alice/repo1", "description": "Repo 1",
+        "stargazers_count": 10, "language": "Python",
+        "updated_at": "2026-01-01T00:00:00Z",
+        "html_url": "https://github.com/alice/repo1",
+    }]
+    http_client = _make_http_client(_make_response(starred_data))
+    app = _build_app_with_app_config(
+        token=None,  # No PAT
+        http_client=http_client,
+        gh_cli_token="gh-cli-token",
+    )
+
+    async def _fake_subprocess(*args, **kwargs):
+        proc = MagicMock()
+        proc.communicate = AsyncMock(return_value=(b"gh-cli-token", b""))
+        return proc
+
+    transport = ASGITransport(app=app)
+    with patch(
+        "tinyagentos.routes.github.asyncio.create_subprocess_exec",
+        side_effect=_fake_subprocess,
+    ):
+        # Also ensure get_installation_token is NOT called (user-scoped skips it)
+        with patch(
+            "tinyagentos.github_app.get_installation_token",
+        ) as mock_mint:
+            async with AsyncClient(
+                transport=transport, base_url="http://test"
+            ) as c:
+                resp = await c.get("/api/github/starred?page=1")
+
+    assert resp.status_code == 200
+    # App token minting must NOT have been called for a user-scoped endpoint
+    mock_mint.assert_not_called()
+
+
+@pytest.mark.asyncio
+async def test_notifications_use_gh_cli_when_app_configured_no_pat():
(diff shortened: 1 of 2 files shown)


hognek:
All 4 remaining Kilo SUGGESTIONS addressed in 74f11188:

1. **Repo-list cache** (github_app.py) — added 5-min TTL cache for installation repo lists to avoid N sequential round-trips per page load
2. **Non-200 callback surfacing** (github_oauth.py) — redirects with `?install_error=1` query param instead of failing silently
3. **Action allow-list** (gh_webhook.py) — explicit `_HANDLED_INSTALLATION_ACTIONS` frozenset with early return
4. **fetchJson non-2xx** (github.ts) — verified already correct; `fetchJson` returns fallback on `!res.ok` before reaching `res.json()`

All bot checks green. 59/59 targeted tests pass.

hognek:
## Bot-fix round 2 (cbb61eb7)

Addressed remaining CodeRabbit findings and the SPA build CI failure:

### SPA build fix (CRITICAL)
- **SecretsApp.test.tsx**: Added missing `listAppInstallations`, `beginAppInstallation`, `deleteAppInstallation` mocks. The Secrets app now renders `GitHubConnect` which imports these from `@/lib/github` — the vitest mock was incomplete, causing the spa-build CI job to fail with "No listAppInstallations export is defined on the mock."

### CodeRabbit nitpicks
- **github_oauth.py**: Simplified nested `installs_store` condition — consolidated `if not installs_store or not installs_store.get(iid): if installs_store:` into `if installs_store and not installs_store.get(iid):`
- **GitHubConnect.tsx**: Added `refreshInstallations` to `useEffect` dependency array + window focus listener so installations list refreshes automatically when user returns from the external GitHub App install flow

### Error handling improvements
- **github_app.py**: Moved `generate_jwt()` inside try blocks in both `get_installation_token()` and `list_installations()` — malformed keys now return `None`/`[]` per contract instead of raising unhandled exceptions
- **github_app_installations.py**: Added `AttributeError` and `TypeError` to `_load()` except clause to handle malformed JSON shapes (e.g., "installations" as a list instead of dict)

### Testing
- Backend: 64/64 targeted tests pass (test_github_app, test_github_oauth, test_config)
- Desktop: `npm run build` passes (tsc + vite), 10/10 SecretsApp tests pass

jaylfc:
Deep-reviewed at head cbb61eb7. The skeleton is genuinely solid (mutating routes carry verify_csrf, SPA uses withCsrf, webhook add/remove only after enforced HMAC, installation tokens never reach the SPA, no open redirect) and most earlier findings are verified folded. Held on four things before this can merge - important because #858 C2 token-injection builds directly on it:

1. BLOCKER (CodeRabbit CRITICAL, posted after your last commit): github_app_installations.py:70 - _save() serializes the live self._installations dict via asyncio.to_thread while webhook + route paths mutate it concurrently: dict-changed-during-iteration aborts the save, and concurrent saves collide on the single .tmp file and can persist stale state out of order. Snapshot the dict before handing it to the thread and serialize saves (single flush task or a lock).

2. BLOCKER (regression): routes/github.py:38-114 - _get_token now prefers the App installation token, but installation tokens cannot serve user-scoped endpoints, so /api/github/starred, /notifications, /auth/status break for gh-CLI users the moment an admin configures the App. The PAT-first reorder fixed only the PAT case. Order user-credential sources (PAT, gh CLI) before the App token for user-scoped calls, or route per-endpoint.

3. SECURITY: the App RSA private key is stored plaintext in config.json (config.py:58-59) and re-serialized on every config save. Our own posture encrypts mere PATs at rest (test_token_encrypted_at_rest); the far more powerful signing key must live in the SecretsStore (owner-only, encrypted) or at minimum be a path to a permission-checked key file. Related: config.py:56-57 contract says path but the implementation expects inline PEM - pick one and align config/JWT/tests.

4. SECURITY: GET /api/github/app/callback (github_oauth.py:323-392) is CSRF-exempt (GET) and never verifies the installation belongs to the acting user. GitHub setup callbacks carry no state param, so an attacker who installs the same public App on their own account can lure a logged-in user to .../app/callback?installation_id=<attacker's>; the mint succeeds and _get_token's first-active-installation fallback can then route operations through attacker-scoped credentials. Verify ownership via the installation's account login against the connected identity, or add an explicit confirm step.

Also: finish the generate_jwt-inside-try fix on the remaining 2 of 4 helpers (_get_app_slug, delete_installation still 500 on a malformed key), and add a short self-hoster doc for registering the App (app id, PEM, webhook secret, setup URL) - runtime degrades gracefully when unset, but nobody can enable this without reading source.

Fold those and I'll merge promptly - this unblocks the #858 C2 chain.

hognek:
## Round 3: Security fixes (t_af81fd5c)

**1. Ownership verification in callback** — GET /api/github/app/callback now verifies the installation account login matches a connected GitHub identity before persisting. Mismatches redirect with `?install_error=ownership_mismatch`. This prevents an attacker from luring a user to the callback with their own installation_id and installing a foreign App on the victim's instance.

**2. generate_jwt-inside-try on remaining 2 helpers** — `_get_app_slug` and `delete_installation` now have `generate_jwt()` inside the try block, matching the fix already applied to `get_installation_token` and `list_installations` in round 2. Malformed keys now return None/False per contract instead of 500.

**3. Self-hoster documentation** — `docs/github-app-setup.md` covering App registration (app id, PEM, webhook secret, callback/setup URL), config keys, security notes, and the full installation flow.

**Tests:** 35/35 pass (16 test_github_app + 19 test_github_oauth):
- 4 new malformed-key tests (TestMalformedKeyGraceful)
- 2 new callback ownership tests (mismatch → reject, match → allow)

hognek:
## Rebase to resolve merge conflict (t_16675997)

Rebased `feat/github-app-installation` onto `9c4add9d` (PR base SHA). Resolved one conflict in `tinyagentos/config.py`: both `wallhaven_api_key` (from dev) and `github_app_id`/`github_app_private_key` (from this branch) need to coexist as adjacent fields.

**Test results:** 91/91 pass (test_routes_github, test_github_app, test_github_oauth, test_config).

**Mergeable:** true (was `dirty`/`CONFLICTING` before rebase).

jaylfc:
Good progress - I verified all four of my findings against the 8f1dd86 head:
- Token-source regression (finding 2): FIXED. _get_token(user_scoped=True) skips the App token so /auth/status, /starred, /notifications keep working for gh-CLI/PAT users. Correct.
- Callback ownership + generate_jwt-in-try (finding 4): FIXED. The callback now fetches /app/installations/{id} and rejects unless the installation account.login matches a connected device-flow identity (fails closed), and generate_jwt is inside the try in all 4 helpers. The foreign-installation_id attack is closed.

Two blockers remain, and both are because the fixes live in SEPARATE PRs - #1932 cannot merge alone or it ships the bug + the plaintext secret:
- Save-race (finding 1): still present at this head. github_app_installations.py _save_sync still iterates the LIVE self._installations dict inside the to_thread worker (line ~60), with add()/remove() mutating it on the event loop from both the webhook and callback paths -> dict-changed-during-iteration + fixed .tmp collision. The intended fix is in #1997 (snapshot before to_thread + a save lock). Land #1997 first (or fold its snapshot+asyncio.Lock into #1932), else this ships broken.
- Plaintext RSA key (finding 3): #1932's own config.py:61 + to_dict still write the PEM into config.yaml. The SecretsStore move is in #2009. #1932 must merge AFTER #2009 (or fold it in), otherwise the plaintext-key exposure ships.

Cleanest path: land #1997 + #2009 to dev, rebase #1932 on top (it then inherits both), and it is a clean merge. One non-security note for the #858 follow-up: the ownership check matches installation account.login against USER device-flow identities, so an Organization installation (account.type Organization) gets rejected - fine to fail safe for now, but worth handling org installs later. CI is genuinely green here, nice.

PR #1933, opened 2026-07-17
feat(desktop): auto-reload SPA when new build is deployed

### LINKED ISSUE #1914: SPA: auto-reload desktop when a new build is deployed (kill the hard-refresh papercut)
After a Settings-update + SPA rebuild, users keep seeing stale UI until they manually hard-refresh (e.g. the invite-dialog z-index fix looked 'not applied' because the browser served the old cached bundle). 

Add a lightweight build-version check: embed the build SHA/hash in the served app, expose it at a tiny endpoint (or read the hashed asset manifest), and have the running SPA poll/compare on focus or interval; when the deployed build differs, show a subtle 'New version — reload' toast (or auto-reload after idle). Must not disrupt an active session (don't reload mid-typing).

Root cause context: content-hashed bundles change on deploy, but index.html can be cached by the browser, so clients keep loading old asset hashes. A version ping + controlled reload closes the gap for all users.

hognek:
## Summary

Kill the hard-refresh papercut — the desktop SPA now detects when a new build has been deployed and prompts the user to reload via a notification toast.

Closes #1914.

## What changed

- **read-version.mjs**: Now always includes a nanosecond-resolution timestamp alongside the git SHA, so every build produces a unique identifier even from the same commit. This is what the version-check poll compares against.

- **vite.config.ts**: New `spaVersionPlugin` writes `version.json` to the build output with the full `__TAOS_VERSION__`.

- **useSpaVersionCheck hook**: Polls `/desktop/version.json` on mount, on window focus, and every 60 s while the tab is visible. Compares the deployed version with the running `__TAOS_VERSION__`.

- **SpaUpdateToast component**: Pushes a transient notification via the notification store when a mismatch is detected. Dismissable; fires at most once per unique deployed version. Silent in dev builds. Does NOT auto-reload (avoids disrupting active sessions).

- **AppShell**: Wires `SpaUpdateToast` alongside the existing `UpdateAvailableToast`.

## Tests

- **use-spa-version-check.test.ts**: 6 tests covering matching versions, mismatch detection, dev-build suppression, focus/visibility polling, and error/404 handling.
- **SpaUpdateToast.test.tsx**: 5 tests covering silent state, notification firing, de-duplication, and version-change re-firing.
- All 44 existing related tests pass (UpdateAvailableToast, BackendBanner, AppErrorBoundary, etc.).

## Verification

- `npm run build` produces `version.json` with the full build version in `static/desktop/`
- Service worker cache name includes the unique build identifier
- `npx vitest run` — all new + existing tests pass

### DIFF
--- a/desktop/scripts/read-version.mjs
+++ b/desktop/scripts/read-version.mjs
@@ -20,7 +20,7 @@ function readPackageVersion() {
 
 function readBuildId() {
   // Include both the git short SHA (reproducible anchored identifier) and
-  // a nanosecond-resolution timestamp so every build — even back-to-back
+  // a high-resolution timestamp so every build — even back-to-back
   // builds on the same commit — produces a unique identifier.  The SPA
   // version-check poll relies on this to detect redeployed bundles.
   //
@@ -37,6 +37,10 @@ function readBuildId() {
   }
   const ms = Date.now().toString(36);
   const hr = process.hrtime.bigint().toString(36);
+  // Take the last 6 base-36 chars of the nanosecond counter — combined
+  // with the ms component this yields ~2.1B unique values per ms tick,
+  // which is plenty for sub-millisecond build differentiation while
+  // keeping the stamp compact.
   const stamp = `${ms}${hr.slice(-6)}`;
   return sha ? `${sha}.${stamp}` : stamp;
 }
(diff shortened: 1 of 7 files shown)


hognek:
Addressed Kilo review findings — fixed doc/code mismatch (no-cache → no-store), typed vite config (ResolvedConfig), clarified version precision wording in read-version.mjs.

PR #1908, opened 2026-07-17
feat(user_shares): add share routes + consent wiring (POST/GET/DELETE /api/shares)

hognek:
## Summary
Implement `tinyagentos/routes/user_shares.py` with three authenticated routes for user-to-user resource sharing, plus consent wiring (notifications + Decision records).

**Routes:**
- `POST /api/shares {resource_type, resource_id, to_username, permission}` — share a resource with another user by username. Resolves via AuthManager.find_user(), creates share in UserSharesStore, raises notification + Decision record for consent.
- `GET /api/shares?direction=out|in` — list shares (out = what you have shared, in = what is shared with you). Defaults to `out`.
- `DELETE /api/shares/{id}` — revoke a share. Owner or admin only (uses require_owner_or_admin against the share owner_user_id).

**Consent wiring** (mirrors agent_auth_requests.py lines 204-221):
- Notification raised to target user with share_id, owner_user_id, resource_type, resource_id, permission
- Decision record (type=approve_deny) created for target user Decisions inbox

**Also:**
- Module-level `user_can_access()` helper for route consumers
- UserSharesStore registered in app.py lifespan
- Router registered in routes/__init__.py

**Depends on:** PR #1897 (UserSharesStore)

**Tests:** 19/19 pass (test_register_all_routers, test_app_startup_guard, test_main_entry, test_version_endpoint)

---

**Kilo bot fixes (commit 4e52ffe):**
- WARNING x2 (expiry): `_normalise_expiry` parses caller-supplied `expires_at` to UTC-aware datetime on write, so lexical ISO comparisons in `list_active_shares`/`user_can_access` are safe against timezone offsets, Z suffix, naive timestamps, and sub-second precision.
- SUGGESTION: Added `get_share(share_id)` to UserSharesStore (indexed PK lookup); refactored `_find_share_by_id` to use it instead of O(N) full scan.
- WARNING: Documented one-time `CREATE INDEX` cost in NotificationStore._post_init (IF NOT EXISTS makes it no-op on subsequent starts).

## Summary by CodeRabbit

* **New Features**
  * Added user-to-user resource sharing by username with configurable permissions and expiration.
  * Added incoming and outgoing share lists.
  * Recipients can accept or deny pending shares.
  * Owners and administrators can revoke active shares.
  * Added notifications scoped to individual users while preserving broadcast notifications.
  * Added protections against invalid, self-directed, duplicate, and unauthorized sharing actions.

PR #1945, opened 2026-07-17
test(user_shares): add store + route tests with conftest integration

hognek:
## Summary
Tests for the user-to-user resource sharing feature (taOS #1893 C3), targeting the implementation in PRs #1897 (UserSharesStore) and #1908 (share routes + consent wiring).

Depends on: #1897, #1908

### Files
- **tests/test_user_shares_store.py** (new, +207 lines) — 13 store tests mirroring test_agent_grants_store.py patterns
- **tests/test_user_shares_routes.py** (new, +233 lines) — 9 route tests mirroring test_routes_agent_auth_requests.py patterns
- **tests/conftest.py** (+9 lines) — init user_shares store in client fixture with graceful fallback

### Store tests (13)
add_share: returns_inserted_row, with_optional_fields, idempotent_replace, uninitialised_raises
list_shares: returns_all, empty_for_unknown, scoped_by_owner, received
revoke_share: removes_it, nonexistent_does_not_crash
user_can_access: true_for_active, false_for_expired, false_for_none

### Route tests (9)
create_share (200), unknown_username (404), duplicate_idempotent (200), to_self_rejected (400)
list_outgoing (200), list_incoming (200)
revoke (200), nonexistent (404), non_owner_unauthorized (403)

### Test results
Store: 13/13 pass
Route: 9/9 pass
Full suite: known aiosqlite daemon-thread hang on teardown (pre-existing issue)

## Summary by CodeRabbit

* **New Features**
  * Added user-to-user resource sharing with options to create, view, accept, deny, and revoke shares.
  * Added incoming and outgoing share lists, access controls, duplicate handling, expiration support, and owner/admin revocation permissions.
  * Added notifications and decision tracking for sharing requests.

* **Documentation**
  * Documented the sharing workflow and permission rules.

* **Tests**
  * Expanded coverage for sharing, authorization, CSRF protection, expiration, and invalid requests.

### DIFF
--- a/docs/agent-coordination.md
+++ b/docs/agent-coordination.md
@@ -175,6 +175,28 @@ scopes) may be granted globally (`project_id=None`) or per-project. Creation
 surfaces a bell notification (`source: agent_scope_requests`) to the owner/admin,
 retired when the request is decided.
 
+## User resource sharing (share routes)
+
+Users can share resources with each other through the `/api/shares` endpoints
+in `tinyagentos/routes/user_shares.py`. The consent loop mirrors the
+external-agent consent pattern (`agent_auth_requests.py`): on share-create, a
+notification and a Decision record (type `approve_deny`) are raised to the
+target user so the desktop consent actions can approve or deny:
+
+- `POST /api/shares {resource_type, resource_id, to_username, permission}` —
+  share a resource with another user by username. Resolves the target via
+  AuthManager; self-share is rejected (400). Duplicate shares (same owner,
+  resource, target, permission) are idempotent.
+- `GET /api/shares?direction=out|in` — list shares. `out` (default) returns
+  shares the user owns; `in` returns shares where the user is the target.
+- `POST /api/shares/{id}/accept` — accept a pending share (target user only).
+  Once accepted, the module-level helper `user_can_access()` returns True for
+  that resource.
+- `POST /api/shares/{id}/deny` — deny a pending share (target user only).
+  The share row is preserved with `status=denied` for audit.
+- `DELETE /api/shares/{id}` — revoke a share. Owner or admin only
+  (requires `require_owner_or_admin` against the share's `owner_user_id`).
+
 ## Project invite redeem route (link + PIN)
 
 A project invite lets an external agent join without going through the consent
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -431,6 +431,13 @@ async def client(app, tmp_data_dir):
     if council_members._db is not None:
         await council_members.close()
     await council_members.init()
+    # user_shares store (user-to-user resource sharing) is lifespan-owned;
+    # tests that bypass the lifespan must init it so routes can consult it.
+    user_shares = getattr(app.state, "user_shares", None)
+    if user_shares is not None:
+        if user_shares._db is not None:
+            await user_shares.close()
+        await user_shares.init()
     # BrowserApp v2 stores
     from tinyagentos.routes.desktop_browser.store import BrowserStore, BrowserCookieStore
     _browser_store = BrowserStore(tmp_data_dir / "browser.sqlite3")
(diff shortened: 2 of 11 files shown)


hognek:
Kilo bot findings (1W + 3S) fixed in [b8d7a5e](https://github.com/hognek/tinyagentos/commit/b8d7a5e8):

- **WARNING** (merge-order coupling): Added `pytest.importorskip` for UserSharesStore + `_require_method()` guards for `accept_share`/`list_shares_received` — tests now skip gracefully when #1897 isn't merged yet
- **S1** (direction filtering): `test_list_outgoing_shares` now creates an incoming share (bob→admin) and verifies it's excluded from the outgoing list
- **S2** (redundant bob_client): Removed duplicate bob user creation in `bob_client` fixture (already guaranteed by `shares_client`)
- **S3** (hardcoded username): `test_create_share_to_self_rejected` now resolves the current username via `auth.get_user(session_token)` instead of hardcoding 'admin'

Test results: 13/13 store tests skip cleanly when the store module isn't available. Route tests skip at fixture level. Existing aiosqlite daemon-thread hang on teardown is pre-existing.

jaylfc:
Heads up: the real CI workflow (test 3.12/3.13 + lint + spa-build) did not trigger on this one, only CLA + the automerge skip registered. A tests-only change can miss the workflow path filters. An empty commit or a rebase onto latest dev should kick the full run, then I can verify and merge. #1946 I have just approved the pending fork CI run.

jaylfc:
@hognek Held on ordering, not quality. This tests `user_shares`, which is not on dev yet - it arrives in either #1908 or #1958, and those two are competing implementations that need a decision first.

Flagging one thing about the green tick: **every test here is currently skipping.** `pytest.importorskip` plus `pytest.skip("... not available yet")` means CI passes by testing nothing, and it will keep passing silently after merge until the implementation lands. That is the correct guard for landing tests ahead of code, so no objection, but it does mean this PR's green is not evidence of anything yet.

After the implementation merges, worth re-checking these actually run rather than assuming: if the module path in the importorskip does not match what shipped, they stay skipped forever and look identical to passing.

jaylfc:
@hognek Ordering settled: #1908 is the chosen user_shares implementation (see #1958 for the reasoning). Merge order is #1908, then this.

Standing request from the earlier comment: after #1908 lands, confirm these tests actually **run** rather than still skipping. They are green today only because `pytest.importorskip` is swallowing them, and if the module path in the guard does not match what #1908 ships, they stay skipped forever and look exactly like passing.

jaylfc:
Deep review done, this is merge-grade and I want it in. Two things first, then I merge on your push:

1. I retitled it feat(...) since the branch absorbed the #1908 implementation commits and now ships the whole feature (store, routes, app wiring, consent) plus the tests. Release notes need that to be honest. Nothing for you to do on this point.
2. One coverage gap survived my mutation testing: making direction=in return the owned+received union still passes all 24 route tests, so the incoming-direction exclusion is unverified (the outgoing side has its check). Please add the symmetric assertion CodeRabbit suggested. While in there: drop the duplicate app.state.agent_grants assignment in the app.py hunk (dev already sets it later), and the unused user_can_access import in tests/test_user_shares.py.

For the record, the store and route auth tests are real: expiry filter and owner-guard mutations both go red. Kilo's double-close and uninitialised-close warnings are false (BaseStore.close is None-guarded), and its unmerged-dependency warning is obsolete since the implementation ships in this PR.

jaylfc:
Verified and merging. Your new assertion does exactly what it needed to: I re-ran the mutation that survived the first review (making direction=in return the owned set, the union case) and it now FAILS where before it left all 24 tests green. Restored, 24 passed. That was the only gap; the store and route auth tests were already mutation-proven in the first pass.

Two notes, neither blocking. Your branch predates the dependency-groups fix that merged to dev this morning (#2217), so plain uv sync gives no pytest on it and you need --extra dev; once you rebase onto current dev that goes away and uv sync alone works. And the Kilo check here is the rate-limited fake red, no findings, so I am not treating it as a signal.

Thanks for turning this round quickly. Merging now, and #1908 is already closed as superseded by this.

PR #1960, opened 2026-07-17
fix(stores): audit SCHEMA CREATE INDEX vs _post_init ADD COLUMN + regression gate

### LINKED ISSUE #1865: Audit all stores for the SCHEMA-index-before-migration boot brick
BaseStore runs the SCHEMA string before _post_init migrations, so any CREATE INDEX in SCHEMA that references a migration-added column bricks boot on an existing pre-change DB (no such column). This hit the Pi more than once (registry, canvas, task element_id). Audit every store: cross-reference each _post_init ADD COLUMN X against any SCHEMA CREATE INDEX ... (X) and move such indexes into _post_init after the ALTER. Add an existing-DB upgrade test where relevant.

hognek:
Task: t_93592624. Closes #1865.

## What
- Updated BaseStore docstring: documented that CREATE INDEX in SCHEMA
  referencing `_post_init`-added columns bricks boot, with fix guidance
- Added `tests/test_store_upgrades.py`: 21 tests covering all 9 stores
  with `_post_init` column migrations, including a regression class
  (`TestNoIndexOnMigrationColumns`) that smoke-tests init() on v0-schema
  databases
- Audit confirmed: no store currently has the bug — element_id indexes on
  canvas/task stores were already correctly placed in `_post_init`

## Test results
```
tests/test_store_upgrades.py - 21 passed in 12.76s
```

### DIFF
--- a/tests/test_store_upgrades.py
+++ b/tests/test_store_upgrades.py
@@ -1,0 +1,557 @@
+"""Existing-DB upgrade tests for stores with guarded _post_init column migrations.
+
+BaseStore.init() runs SCHEMA (CREATE TABLE + CREATE INDEX) before _post_init
+migrations.  Any CREATE INDEX in SCHEMA that references a column added by
+_post_init bricks boot on existing databases — the column does not exist yet
+when the index is created.
+
+These tests simulate the upgrade path by seeding a "v0" database (only the
+original columns, no migration columns) and then verifying that store.init()
+successfully adds the migration columns via _post_init.
+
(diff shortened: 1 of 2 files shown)


jaylfc (approved):
Reviewed: independently re-audited all 9 stores' SCHEMA vs _post_init ordering, the audit claim holds, and the v0-upgrade regression gate targets the right boot-brick invariant (SCHEMA CREATE INDEX on a migration-added column over an existing DB). Merging.

Issue #124, opened 2026-04-11
qmd client: exponential backoff retry on transient failures


jaylfc:
The qmd client (memory search, embedding, rerank) currently fails hard on transient network blips — a single failed embed call surfaces as an error to the caller. TinyAgentOS needs a retry wrapper so the qmd hot path is resilient to brief controller-side GC pauses, connection resets, and rkllama model loads in progress.

**What needs to ship:**
- [ ] Retry wrapper in \`tinyagentos/clients/qmd_client.py\` around every call
- [ ] Exponential backoff with jitter: 100ms, 300ms, 900ms, then give up
- [ ] Retry only on connection errors and 5xx responses, not on 4xx
- [ ] Per-call timeout override so one slow call doesn't block the whole chain
- [ ] Logging at WARNING when a retry fires, ERROR only after all retries exhausted
- [ ] Tests with httpx mock that verify backoff and give-up semantics

**User impact today:** transient failures look like hard errors to agents, which can make a conversation hiccup or drop a turn.

PR #1961, opened 2026-07-17
feat(qmd): add retry jitter and per-call timeout override

hognek:
## Summary

Adds two features to the qmd client retry story ("#124):

1. **Jitter** — `with_retry()` now applies a `[0.5x, 1.5x]` random factor to each retry delay by default (`jitter=True`), preventing thundering-herd effects when multiple clients retry simultaneously. Callers can pass `jitter=False` for deterministic timing in tests.

2. **Per-call timeout override** — `QmdClient.embed()` and `QmdClient.search()` now accept an optional `timeout` parameter that overrides the client's default 60s timeout for that single request, so one slow model load doesn't block the whole chain.

## Changes

| File | Change |
|------|--------|
| `tinyagentos/clients/retry.py` | Add `import random`, `jitter` kwarg (default True), jitter in all 3 sleep paths, updated docstring |
| `tinyagentos/qmd_client.py` | Add `timeout: float | None = None` kwarg to `embed()` and `search()`, pass through to httpx |
| `tests/test_clients_retry.py` | Pass `jitter=False` to deterministic timing tests; add `test_jitter_varies_delays` |

## Test results

```
tests/test_clients_retry.py    — 14 passed (incl. new jitter test)
tests/test_qmd_client.py        —  3 passed
tests/test_adapter_retry.py     —  9 passed
tests/test_health_module.py     — 23 passed
                                 49 passed total
```

All existing callers of `with_retry()` (9 adapters + qmd_client + scheduler/failure_handler) use default parameters and automatically get jitter — no consumer changes needed.

### DIFF
--- a/tests/test_clients_retry.py
+++ b/tests/test_clients_retry.py
@@ -133,7 +133,8 @@ class TestWithRetry:
         assert call_count["n"] == 2
 
     async def test_exponential_backoff_timing(self):
-        """Verify delays grow exponentially and respect max_delay."""
+        """Verify delays grow exponentially and respect max_delay (jitter disabled
+        for deterministic assertions)."""
         sleep_calls: list[float] = []
 
         original_sleep = asyncio.sleep
@@ -153,6 +154,7 @@ class TestWithRetry:
                     base_delay=0.1,
                     multiplier=3.0,
                     max_delay=0.5,
+                    jitter=False,
                 )
 
         # Should have slept 3 times (between 4 attempts)
@@ -178,6 +180,7 @@ class TestWithRetry:
                     base_delay=1.0,
                     multiplier=10.0,
                     max_delay=2.0,
+                    jitter=False,
                 )
 
         for delay in sleep_calls:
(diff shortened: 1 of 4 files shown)


hognek:
All 3 Kilo findings fixed in 57928ee:

- **WARNING (retry.py:107)** — Jitter now applied *before* the max_delay cap, and the result is capped at max_delay so sleep never exceeds the documented bound.
- **SUGGESTION (retry.py:105)** — Log message now uses the actual computed sleep_time (post-jitter, post-cap).
- **SUGGESTION (test_clients_retry.py:268)** — Replaced no-op `assert d >= 0.0` with meaningful checks: `d > 0` and `d <= max_delay`.

Tests: test_clients_retry.py — 14/14 pass.

jaylfc:
Thanks hognek. The retry.py jitter half is solid: the cap is correctly applied after jitter in all three paths, the logs emit the real sleep_time, and the 3 earlier Kilo findings are genuinely folded into HEAD (57928ee). test_clients_retry.py 14/14. But do not merge yet, there is one blocking regression in qmd_client.py.

The new `timeout: float | None = None` default passed straight into `client.post(..., timeout=timeout)` does NOT fall back to the client default. In httpx, the request-method default is the sentinel `USE_CLIENT_DEFAULT`; an explicit `timeout=None` means NO timeout (infinite). Verified on httpx 0.28.1:

    sentinel -> connect/read/write/pool = 60 (the AsyncClient(timeout=60) default)
    timeout=None -> all None (infinite)

So the default path (embed/search called with no timeout arg) goes from the 60s ceiling to unbounded. That contradicts the PR docstring ("overrides the client default (60 s)") and the stated goal (one slow model load should not block the chain), and it silently strips the 60s protection from the existing caller at routes/projects.py:1015 (`await qmd.search(...)`, the project-search endpoint), which can now hang forever. No test covers timeout so CI would not catch it, and Kilo blessed the passthrough as correct, but it is not.

Fix: default the param to the httpx sentinel so unset keeps the client default and an explicit value overrides:

    import httpx
    async def embed(self, text: str, *, timeout=httpx.USE_CLIENT_DEFAULT): ...
    async def search(self, ..., *, timeout=httpx.USE_CLIENT_DEFAULT): ...

(the existing `timeout=timeout` passthrough then behaves correctly). Update the type hints/docstrings to match, and please add a small test asserting that an unset call keeps the 60s client default while an explicit value overrides. Optional: the 5 ruff errors in tests/test_clients_retry.py are all pre-existing unused imports, not from this PR, but a quick cleanup while here would be welcome. After that, MERGE.

hognek:
Fixed in b8ad7069:

- **Timeout regression**: Changed `timeout: float | None = None` → `timeout=httpx.USE_CLIENT_DEFAULT` in both `embed()` and `search()`. Unset calls now preserve the 60s client default; explicit floats override as documented.
- **Ruff cleanup**: Removed 5 pre-existing unused imports/variables in `test_clients_retry.py` (time, AsyncMock, MagicMock, DEFAULT_RETRY_ON_STATUS, original_sleep). Ruff is now clean.
- **New tests**: `test_embed_timeout_default` (asserts USE_CLIENT_DEFAULT passthrough) and `test_embed_timeout_override` (asserts 30.0 float passthrough).

Tests: retry 14/14, qmd_client 5/5 — all pass.

jaylfc (approved):
Reviewed: the unbounded-timeout regression is fixed (httpx.USE_CLIENT_DEFAULT sentinel + explicit override + two tests), and the three Kilo nits are folded. Merging.

PR #1965, opened 2026-07-17
feat(chat): truncate conversation branch + re-trigger agent on message edit

### LINKED ISSUE #800: Message editing: rewrite conversation branch and re-trigger agent reply
When a user edits a past message in a conversation today, the edit mutates the stored text but the agent never sees the change. The next reply still carries the original transcript. This causes context drift: the agent's mental model diverges from what the user actually meant, and the only recovery is manually re-explaining or starting a new conversation.

The fix is straightforward: an edit to any user message in an agent conversation should truncate everything after that turn, replace the edited message as the canonical user turn, and (optionally) immediately re-trigger the agent. ChatGPT, Claude.ai, and most modern chat UIs do this. It is the right behavior.

## Design considerations

**Scope: two surfaces**

1. **Messages app** (`MessagesApp.tsx`, channels with agents). The edit infrastructure already exists: `MessageEditor.tsx`, `MessageOverflowMenu` with an "Edit" item, `editMessage` API (`PATCH /api/chat/messages/{id}`), and a WS `message_edit` broadcast. What is missing is the truncation step: after saving the edit, delete all messages in the channel that were created after the edited message, then optionally trigger a fresh agent reply.

2. **taOS agent chat panel** (`TaosAssistantPanel.tsx`). History is held in the Zustand `messages` array and rebuilt as `historyPayload` on every send. Edit here is purely client-side: replace the `AssistantMessage` at index N, drop everything from N+1 onward, then call `doSend` with the updated history. No backend persistence needed beyond what already happens.

**UI affordance**

- Hover on a user message bubble in the taOS agent panel should show an edit icon (pencil) or "Edit" in a context menu, consistent with `MessageHoverActions` in the Messages app.
- Clicking edit opens the text inline (same pattern as `MessageEditor.tsx`).
- Saving commits the edit and truncates the visible transcript. A "re-ask" button can optionally appear rather than auto-sending, to let the user review before triggering.
- Mark the edited turn with the existing `edited_at` affordance so it is clear which turn was changed.

**History rewrite semantics**

- For Messages app: truncation deletes rows in the DB after the edited message's `created_at`. The backend already has `delete_message` per-ID; a batch endpoint or a new `POST /api/chat/channels/{id}/truncate-after/{message_id}` is cleaner.
- For the taOS agent panel: pure in-memory splice, no backend call needed.
- Deleted agent turns should be tombstoned (already supported via `MessageTombstone.tsx`) or silently removed depending on whether the thread panel references them.

**Re-trigger behavior**

- Default: show a "Re-ask" button after truncation rather than auto-sending, so the user can adjust the edit before committing.
- Auto-send can be a follow-up toggle in channel settings.

## Dependencies / ties to existing work

- `MessageEditor.tsx` and `MessageOverflowMenu.tsx` already implement the editing affordance for Messages; the `isOwn` gate is in place. The missing piece is the post-edit truncation+re-trigger call.
- `editMessage` in `chat-messages-api.ts` covers the PATCH; a new truncation endpoint is needed alongside it.
- `MessageTombstone.tsx` handles the deleted-message display, relevant if we tombstone removed agent turns.
- Existing `edited_at` field on the `Message` type is already rendered.
- taOS agent panel Zustand store (`taos-agent-store.ts`) already exposes `setMessages`, making the client-side splice trivial.
- #657 (zero tests on `MessagesApp.tsx`) is relevant: any new edit+truncate flow should land with tests given that file's size.
- #156 (cross-node context summarization) touches message history management; the truncation semantics here should be consistent with whatever history window that feature uses.

## v1 scope cuts

- No edit of agent-authored messages (edit is user messages only).
- No "branching" UI showing alternate history trees. One canonical branch only.
- No auto-send on edit save. Re-ask button is explicit.
- No multi-turn partial truncation (select a range to re-do). Just truncate everything after the edited turn.
- taOS agent panel edit is in-memory only; no server-side session reset for the opencode server (the session carries context internally, so a mismatch is possible; resolving that is a follow-up).

## Open questions

1. Should saving an edit in the Messages app also kill an in-flight agent response if one is streaming? Need a cancel mechanism or at least a guard.
2. The opencode session behind the taOS agent panel has its own conversation state server-side. Editing client-side history but not resetting the opencode session id means the model's internal context diverges from what the panel shows. Should we reset the session (and re-send reconstructed history) on an edit, or accept the mismatch for v1?
3. For Messages channels: should agent turns between the edited user message and the current tail be tombstoned (visible but struck through) or silently deleted? Tombstones preserve context for the user but add noise.
4. Should the edit affordance be available during an in-flight agent response, or locked until the response completes?

hognek:
Fixes #800.

## What
When a user edits a past message:
1. Soft-deletes all messages created after the edited message (truncation)
2. Broadcasts per-message delete events to connected clients
3. Re-dispatches the edited message through AgentChatRouter to re-trigger the agent

## Changes
- `tinyagentos/chat/message_store.py`: new `soft_delete_messages_after(channel_id, timestamp)` method
- `tinyagentos/routes/chat.py`: updated `edit_message_endpoint` and WS edit handler
- `tests/test_chat_edit_delete.py`: 3 new tests (truncation, empty-after edge case, cross-channel isolation)

## Tests
```
tests/test_chat_edit_delete.py ..........
tests/test_chat_messages.py ........................
tests/test_routes_chat.py::test_post_message ✓
tests/test_routes_chat.py::test_post_message_with_embeds ✓
```
All 34 targeted tests pass.

## Summary by CodeRabbit

* **New Features**
  * Editing a chat message now removes later messages only within the same DM conversation branch (and thread, if applicable).
  * Updated messages automatically re-trigger agent responses.

* **Bug Fixes**
  * Message edits are now strictly validated (must target an existing, non-deleted message authored by the requester).
  * Subsequent message removal is scoped to the correct channel/thread, leaving other channels unaffected.

* **Tests**
  * Added coverage for edit/delete side effects, empty-results behavior, and DM scoping.

### DIFF
--- a/tests/test_chat_edit_delete.py
+++ b/tests/test_chat_edit_delete.py
@@ -176,3 +176,127 @@ async def test_delete_idempotent(tmp_path):
         r2 = await client.delete(f"/api/chat/messages/{msg_id}")
         assert r1.status_code == 204
         assert r2.status_code == 204
+
+
+@pytest.mark.asyncio
+async def test_edit_truncates_subsequent_messages(tmp_path):
+    """Editing a user message soft-deletes every message created after it
+    in the same channel, and broadcasts delete events for each."""
+    app, client, rec = await _authed_client(tmp_path)
+    async with client:
+        ch_r = await client.post(
+            "/api/chat/channels",
+            json={"name": "g", "type": "group", "description": "", "topic": "",
+                  "members": ["user", "tom"], "created_by": "user"},
+        )
+        ch_id = ch_r.json()["id"]
+
+        # Post three messages: m1, m2, m3
+        m1 = await client.post(
+            "/api/chat/messages",
+            json={"channel_id": ch_id, "author_id": rec["id"], "author_type": "user",
+                  "content": "first", "content_type": "text"},
+        )
+        m2 = await client.post(
+            "/api/chat/messages",
+            json={"channel_id": ch_id, "author_id": "tom", "author_type": "agent",
+                  "content": "second", "content_type": "text"},
+        )
+        m3 = await client.post(
+            "/api/chat/messages",
+            json={"channel_id": ch_id, "author_id": rec["id"], "author_type": "user",
+                  "content": "third", "content_type": "text"},
+        )
+        m1_id = m1.json()["id"]
+        m2_id = m2.json()["id"]
+        m3_id = m3.json()["id"]
+
+        # Edit m1 — this should soft-delete m2 and m3
+        r = await client.patch(
+            f"/api/chat/messages/{m1_id}", json={"content": "first-edited"},
+        )
+        assert r.status_code == 200, r.json()
+        assert r.json()["content"] == "first-edited"
+
+        # m1 should still exist (edited, not deleted)
+        got_m1 = await app.state.chat_messages.get_message(m1_id)
+        assert got_m1["content"] == "first-edited"
+        assert got_m1["deleted_at"] is None
+
+        # m2 and m3 should be soft-deleted
+        got_m2 = await app.state.chat_messages.get_message(m2_id)
+        assert got_m2["deleted_at"] is not None
+        got_m3 = await app.state.chat_messages.get_message(m3_id)
+        assert got_m3["deleted_at"] is not None
+
+
+@pytest.mark.asyncio
+async def test_soft_delete_messages_after_empty_when_nothing_after(tmp_path):
+    """soft_delete_messages_after returns an empty list when there are no
+    messages created after the given timestamp."""
+    app, client, rec = await _authed_client(tmp_path)
+    async with client:
+        ch_r = await client.post(
+            "/api/chat/channels",
+            json={"name": "g", "type": "group", "description": "", "topic": "",
+                  "members": ["user"], "created_by": "user"},
+        )
+        ch_id = ch_r.json()["id"]
+        m = await app.state.chat_messages.send_message(
+            channel_id=ch_id, author_id=rec["id"], author_type="user",
+            content="only message",
+        )
+        store = app.state.chat_messages
+        # After the message's own timestamp there should be nothing.
+        ids = await store.soft_delete_messages_after(ch_id, m["created_at"])
+        assert ids == []
+        # The message itself should still be intact.
+        got = await store.get_message(m["id"])
+        assert got["deleted_at"] is None
+
+
+@pytest.mark.asyncio
+async def test_edit_only_truncates_own_channel(tmp_path):
+    """Editing a message in channel A must not affect messages in channel B."""
+    app, client, rec = await _authed_client(tmp_path)
+    async with client:
+        # Channel A
+        ch_a = await client.post(
+            "/api/chat/channels",
+            json={"name": "a", "type": "group", "description": "", "topic": "",
+                  "members": ["user"], "created_by": "user"},
+        )
+        a_id = ch_a.json()["id"]
+        # Channel B
+        ch_b = await client.post(
+            "/api/chat/channels",
+            json={"name": "b", "type": "group", "description": "", "topic": "",
+                  "members": ["user"], "created_by": "user"},
+        )
+        b_id = ch_b.json()["id"]
+
+        m_a = await client.post(
+            "/api/chat/messages",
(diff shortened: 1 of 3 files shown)


hognek:
**Kilo 2W bot-fix (commit 7dbec8c2):**

1. **WS auth bypass** (routes/chat.py): WS edit handler now validates `msg["author_id"] == user_id` before performing the edit + truncation, matching the HTTP endpoint's 403 guard. Previously, any connected user could wipe channel messages by sending an edit event for another user's message.

2. **Non-atomic truncation** (chat/message_store.py): `soft_delete_messages_after` replaced the rac-y SELECT-then-UPDATE with a single `UPDATE ... RETURNING id` statement (SQLite 3.35+). This closes the concurrent-insert-bypass window.

**Tests:** All 57 targeted tests pass (test_chat_edit_delete, test_chat_messages, test_routes_chat).

jaylfc:
Reviewed at head 7dbec8c2 (verified locally: ruff clean, 34 tests pass, both prior Kilo WARNINGs genuinely fixed - WS edit now enforces author_id==user_id + not-deleted, and the SELECT-then-UPDATE became an atomic UPDATE ... RETURNING). Channel isolation is correct, it is a recoverable soft-delete with per-message delete events, and there is no double-fire. Good. But do not merge yet, the truncation is scoped too broadly in two ways that can delete messages the user did not intend to touch.

1) BLOCKER - truncation ignores thread_id, so it deletes across sibling conversation branches. soft_delete_messages_after (message_store.py:202-218) filters only by channel_id, not thread. In a threaded channel, editing a message in thread A soft-deletes later messages in thread B AND the main timeline. Threads are independent conversations, so this is the exact cross-conversation hazard, just within one channel. Note the re-trigger router already scopes by thread (agent_chat_router.py:93-102), so truncation is asymmetric with re-dispatch. Fix: pass and filter on msg["thread_id"] - delete only within the same thread, or only the main timeline when thread_id is None.

2) BLOCKER - no channel-type guard, so in a multi-human group channel editing an old message soft-deletes everyone else subsequent messages. Callers at chat.py:236 and chat.py:595 truncate unconditionally. The #800 intent is to rewrite the AGENT reply branch (linear/DM). Your own test_edit_truncates_subsequent_messages uses a group channel and asserts another author ("tom") gets deleted - that encodes the risky behavior as expected. Fix: gate truncation to DM channels (or channels whose only other members are agents), and adjust that test.

Nits to fold while you are in there (house rule, no deferred nits):
- WS edit payload (chat.py:217-224) uses direct data["message_id"]/data["content"] - a malformed event raises KeyError caught by the loop broad except and drops the whole WS connection. It matches the other handlers in that loop so it is low severity, but a .get() + guard is cleaner.
- test_edit_only_truncates_own_channel (test_chat_edit_delete.py:283-302) never asserts a2 WAS truncated (only that b1 survived) - add the positive assertion.
- Timestamp tie (message_store.py:212): strict > means a message sharing the edited message exact time.time() survives; at most one stale adjacent message, note only.

Fix 1 and 2 + fold the nits and I will merge.

hognek:
Addressed jaylfc's review blockers and nits on commit 081962b:

**BLOCKER 1** — `soft_delete_messages_after` now accepts and filters on `thread_id`. When provided, truncation is scoped to that thread; when `None`, only main-timeline messages (thread_id IS NULL) are deleted. This matches the router's existing thread-aware dispatch.

**BLOCKER 2** — Truncation gated to DM channels only. Both the WS handler and HTTP PATCH endpoint check `channel.get("type") == "dm"` before calling `soft_delete_messages_after`. Multi-human group edits no longer delete others' messages.

**NITs** (folded into same commit):
- WS edit handler: `.get()` + `isinstance` guard instead of bare `data["message_id"]`/`data["content"]` — malformed events no longer KeyError-drop the connection.
- `test_edit_only_truncates_own_channel`: now asserts a2 WAS truncated (previously only checked b1 survived).
- Both truncation tests switched from group → DM channels.
- Timestamp tie note: strict `>` means at most one stale adjacent message survives a timestamp collision.

Tests: 100/100 pass across affected modules (chat_edit_delete, chat_messages, chat_threads, chat_channels, agent_chat_router, chat_group_policy).

PR #1979, opened 2026-07-18
test: replace always-true assert with issubclass check in test_installer_class_available

hognek:
Kilo suggestion on merged PR #1964: `assert LXCInstaller is not None` is always true at module scope and cannot detect the import-error regression the docstring claims.

**Fix:** Replace with `assert issubclass(LXCInstaller, AppInstaller)` — a meaningful class-hierarchy regression check.

- Added `AppInstaller` import from `tinyagentos.installers.base`
- Changed assertion from identity check to subclass check
- Updated docstring

Tests: 16/16 pass in test_port_allocator.py

----
## Summary by Gitar

- **Test Suite Refactoring:**
    - Added `TestLxcUsesCentralizedAllocator` to enforce that `LXCInstaller` uses the centralized port allocator.
    - Verified `lxc_installer` module no longer exposes `_find_free_port` or `RESERVED_PORTS`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LXC installation port allocation and retry handling when ports are unavailable.
  * Prevented repeated attempts to use ports that previously failed.

* **Documentation**
  * Clarified how container and host ports are mapped during installation.

### DIFF
--- a/tests/installers/test_port_allocator.py
+++ b/tests/installers/test_port_allocator.py
@@ -124,6 +124,9 @@ class TestLxcUsesCentralizedAllocator:
         assert hasattr(mod, "allocate_host_port"), (
             "lxc_installer must import allocate_host_port"
         )
+        assert not hasattr(mod, "RESERVED_PORTS"), (
+            "lxc_installer must not import RESERVED_PORTS; allocate_host_port handles that internally"
+        )
 
     def test_installer_class_available(self):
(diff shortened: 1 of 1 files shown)


jaylfc:
Good catch folding the post-merge Kilo nit on #1964. The branch is CONFLICTING against dev though - please rebase onto origin/dev and I'll merge once green.

PR #1981, opened 2026-07-18
fix(agents): gate force-kill on container run-state, add bounded-hang comment, fix shield orphan

hognek:
Task: t_aef4a61c. Addresses jaylfc's v3 re-review on #1962.

### 1. Run-state gate (the real bug)
Both `bulk_stop_agents` and `stop_agent` previously gated force-kill on
`container_exists()`, which only resolves project membership — it returns
True for existing-but-Stopped containers. A clean graceful stop fires
`incus stop --force` on an already-dead container and reports misleading
`force_killed: false`.

Fix: gate on `list_containers()` with `status == 'Running'`.  The
`list_containers` pattern is already used at agents.py:173-174.

### 2. Bounded-hang comment
Added one-line note that the sleep+force inside asyncio.shield is bounded
by the 120s `_run` cap, to both `_grace_kill` closures.

### 3. Shield orphan
`asyncio.shield(_grace_kill())` left the grace-kill task running detached
if the request was cancelled during the shield window.  Fixed by creating
an explicit Task, catching CancelledError, cancelling the orphan, and
awaiting its cleanup before re-raising.

### Tests
Added `TestKillSwitchRunStateGate` with 3 tests:
- `test_bulk_stop_skips_force_kill_for_stopped_container`
- `test_stop_agent_skips_force_kill_for_stopped_container`
- `test_bulk_stop_force_kills_running_container`

All 3 pass. Existing `test_bulk_stop` still passes.

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -33,6 +33,7 @@ import { useTaosAgentStore } from "@/stores/taos-agent-store";
 import { InstallPromptBanner } from "@/shell/InstallPromptBanner";
 import { EffectsLayer } from "@/theme/effects/EffectsLayer";
 import { SafetyFloor } from "@/components/SafetyFloor";
+import { withCsrf } from "@/lib/csrf";
 
 interface SystemShortcutsProps {
   toggleSearch: () => void;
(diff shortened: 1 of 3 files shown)


hognek:
Closed in favor of updating existing PR #1962 directly.

PR #1983, opened 2026-07-18
fix(gpg): resolve 5 CodeRabbit findings on fingerprint-gate push

hognek:
Fixes 5 CodeRabbit findings on PR #1927 (branch: feat/gpg-signature-verification):

1. **gpg_verify.py** — Switch to `--raw` mode for locale-independent `[GNUPG:] VALIDSIG` parsing instead of the display-only `Primary key fingerprint:` line.
2. **routes/settings.py** — Pin merge_target to the verified `remote_sha` to prevent TOCTOU (concurrent ref update installing a different commit post-verification).
3. **routes/settings.py** — Pass `gpg_required` only when `gpg_prefs.enabled AND required`, so `{enabled: false, required: true}` doesn't block updates.
4. **update_runner.py** — Call `ensure_key_available` before `verify_commit` in both `update_to_master` and `switch_to_branch` so a newly configured fingerprint doesn't fail on first use.
5. **update_runner.py** — Check `git reset --hard` return code and set `ok=False` on failure instead of silently returning success with a potentially wrong SHA.

Tests: 31/31 pass (test_gpg_verify.py, test_update_runner.py)

### DIFF
--- a/tests/test_gpg_verify.py
+++ b/tests/test_gpg_verify.py
@@ -89,6 +89,17 @@ class TestParseFingerprint:
     def test_returns_none_for_empty(self):
         assert _parse_fingerprint("") is None
 
+    def test_parses_fingerprint_when_primary_key_signs_with_extra_fields(self):
+        """When primary key signs but GPG emits fprlen + algo name (13 fields),
+        parts[-1] is not a fingerprint but parts[2] IS the primary key."""
+        output = (
+            "[GNUPG:] NEWSIG\n"
+            "[GNUPG:] VALIDSIG AAAABBBBCCCCDDDDEEEEFFFF0000111122223333 2024-01-01 1704067200 0 4 0 1 8 00 20 EDDSA\n"
+        )
+        fp = _parse_fingerprint(output)
+        # parts[-1] = "EDDSA" (not a fingerprint), parts[2] = primary key fpr
+        assert fp == "AAAABBBBCCCCDDDDEEEEFFFF0000111122223333"
+
 
 class TestParseKeyId:
     def test_parses_rsa_key(self):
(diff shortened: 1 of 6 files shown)


hognek:
**Kilo findings (10:00-10:20) — assessment and fix:**

1. **gpg_verify.py:226** — `--raw` suppresses human-readable `gpg:` lines breaking `_parse_key_id` → ✅ Already fixed in `ce7a1a8b`. `_parse_key_id` now has a VALIDSIG fallback that derives the key ID from the signing-key fingerprint when `--raw` mode is active.

2. **update_runner.py:95** — GPG verification block in `update_to_master` is unreachable in production → 🔧 Fixed in `1e9e4f4f`. Removed the dead `gpg_fingerprint`/`gpg_required` parameters and the 48-line GPG block from `update_to_master`. The production GPG path goes through `verify_remote_commit` (auto_update.py notification) and `switch_to_branch` (settings.py channel switch), both of which are properly wired.

3. **gpg_verify.py:199** — VALIDSIG fallback subkey/primary fingerprint mismatch → ✅ Already fixed in `ce7a1a8b`. When a subkey signs (`len(parts) >= 12`), the key ID is derived from `parts[-1]` (primary-key fingerprint), consistent with `_parse_fingerprint` which also returns `parts[-1]` for subkey signing. When the primary key signs directly, both use `parts[2]`.

**Tests:** 46/47 targeted pass (test_update_runner, test_gpg_verify, test_rollback, test_update_channel_routes). The one timeout is a pre-existing slowness in `test_switches_and_saves_pref` (DB init in test, not related to this change).

hognek:
@coderabbitai review

jaylfc:
Core theme: the VALIDSIG parsing conflates the signing subkey with the user-configured primary fingerprint, causing false-negative verification. Folds: gpg_verify.py:157 VALIDSIG's first field is the signing key (often a subkey), not the primary - valid signatures get rejected; :210 the VALIDSIG fallback derives the long key-id from parts[2] (subkey) while _parse_fingerprint returns the primary (parts[-1]) - inconsistent key identity between paths; :125 the 12-field fallback can mislabel a subkey as primary when parts[-1] is not a valid 40-hex fingerprint; :237 --raw suppresses the human-readable gpg: lines so _parse_key_id (which scans 'using RSA/ECDSA/EDDSA key') always returns None and key_id is never populated; and update_runner.py:158 the recovery hard-reset failure path returns result before result.recovery_tag is set, so the tag is lost on the exact failure it exists for. Lower priority: gpg_verify.py:204 normalize/validate expected_fingerprint before compare. Fold and I merge.

hognek:
**jaylfc findings on VALIDSIG parsing — fix pushed in 7c4ea4cc**

The core bug: when a primary key signs directly but GPG emits extra fields (fprlen + pubkey-algo-name, making `len(parts) >= 12`), the code returned `None` because `parts[-1]` was the algo name (e.g. "EDDSA"), not a 40-char hex fingerprint.

**Root cause:** `len(parts) >= 12` is not a reliable subkey signal — GPG can append `<fprlen> <pubkey-algo>` after the primary fpr even when the primary key signs directly.

**Fix in both `_parse_fingerprint` and `_parse_key_id`:** use the fingerprint shape of `parts[-1]` as the sole subkey heuristic. When `parts[-1]` is a 40-char hex → subkey case, use `parts[-1]`. Otherwise → primary key case, validate and use `parts[2]`.

**Two new tests:** primary key signing with extra GPG fields for both parse functions.

**Tests:** 37/37 pass (test_gpg_verify.py: 32/32, test_update_runner.py: 5/5)

Issue #1864, opened 2026-07-16
Follow-up: model loads as evictable submit_gpu tasks (arbiter Option B)


jaylfc:
The GPU arbiter now shares one VRAM ledger with the model-load path (#185 closed). Next step is routing model loads through GpuArbiter.submit_gpu as real evictable tasks, so a pull that cannot fit queues or evicts a lower-priority running model instead of 503-ing. This is a product/policy decision (which running model gets evicted for a user-initiated pull), so it needs its own design pass. The one-ledger foundation is already in place.

jaylfc:
@hognek — here's a substantial next piece in your wheelhouse (you own the `submit_gpu` wiring, #185).

The **Unified GPU Work Queue** is fully spec'd + planned and owner-approved:
- Design: `docs/design/2026-07-17-gpu-work-queue.md`
- Phased plan (slice-by-slice, TDD, exact file map): `docs/design/plans/2026-07-17-gpu-work-queue-plan.md`

**Your lane: Phase A (slices A1–A10)** — all pure-unit, no hardware needed, right on top of `gpu_arbiter.py`:
- **Start with A1** (`gpu_queue_flag.py` + `gpu_op` trace kind) — smallest, zero behaviour change, unblocks everything.
- Then the arbiter core **A2→A3→A4** (strictly ordered, same file), with **A5→A6** (hooks/events) branching off A2, and **A7→A8→A10** / **A9** (pull-through + queue API + ModelsApp) after.
- Sequencing + parallelism graph is in the plan under "Sequencing and Parallelism".

**Conventions (per the plan):** one PR per slice, cut each branch from `origin/dev`, branch names `feat/gpu-queue-a1-flag … a10-…`. Tests: `uv run pytest <path> -v`; desktop `npm test -- <file>` in `desktop/`.

**What stays with me (needs the physical GPUs):** B1/B4/B5 (Pi `/unload_model` + RK3588 probe + Fedora RTX-3060 eviction) and the C2/C7 Pi/Fedora soak gates. Take Phase A as far as you like and I'll run the live-hardware slices behind you.

No rush to fold your open PRs first — this is independent. Ping me on any slice and I'll review promptly.

jaylfc:
[...]

PR #1984, opened 2026-07-18
fix(gpu-arbiter): op kinds, shadow queue, cancel_op, position, + 4 pre-deploy defect fixes (#1864)

hognek:
## Summary

GPU arbiter operational improvements and pre-deploy defect fixes, all gated behind `TAOS_GPU_QUEUE` (off by default):

### Earlier commits (A2 operations slice)
- Op kinds (`inference` / `load`) — validated on submit, exposed in queue snapshot + position
- Shadow queue dict (`_queued_entries`) — O(1) lookups for cancel_op, position, snapshot
- `cancel_op` — cancel queued tasks with VRAM reservation release; evict running tasks; atomic across physical queue + shadow dict + tombstones
- `queue_position` — 1-based position, global for loads, per-model for inference
- `queue_snapshot` — non-destructive read with position + op/model/backend fields
- Cancellation-toctou fix — re-check `_cancelled_ids` after await windows in drain + retry loops

### New (pre-deploy defect fixes, #1864)
1. **Queue resilience** — `_process_queue` wraps `_drain_queue` in try/except so one bad task doesn't permanently kill the loop.
2. **None guard** — `_check_cluster_admission` skips non-NVIDIA workers (`free_vram_mb is None`).
3. **resource_id threading** — Added to `_QueuedGpuTask`, threaded through `submit_gpu` → `_drain_queue` → `_run_gpu_task`, so drained tasks hold a cluster lease on the correct resource.
4. **Lease renewal** — `_renew_lease_loop` background coroutine calls `renew_lease` every 200 s; stops in finally.

## Testing

- 36/36 GPU arbiter tests pass (894, TOCTOU, queue_ops, queue_flag)
- 26/26 cluster tests pass
- 10/10 scheduler tests pass
- Total: 72/72 related tests, no regressions

Task: t_68584724 (defect fixes), t_c34f79c7 (cancellation atomicity), prior A2 work
Fixes: #1864

## Summary by CodeRabbit

* **New Features**
  * Added queue position tracking for GPU load and inference operations.
  * Added non-destructive queue snapshots with operation, model, backend, submitter, and position details.
  * Added cancellation support for queued and running GPU tasks.
  * Improved GPU lease renewal and cluster admission handling.

* **Bug Fixes**
  * Preserved compatibility with existing GPU submission calls.
  * Prevented cancelled tasks from being re-queued or retaining VRAM reservations.
  * Improved recovery from temporary queue-processing failures.

### DIFF
--- a/tests/test_gpu_arbiter_queue_ops.py
+++ b/tests/test_gpu_arbiter_queue_ops.py
@@ -1,0 +1,112 @@
+"""Tests for GPU arbiter queue ops — op shape, position, cancel, snapshot (taOS #1864 A2)."""
+
+import asyncio
+import pytest
+from tinyagentos.scheduler.gpu_arbiter import GpuArbiter
+from tinyagentos.scheduler.types import Capability, Priority, Task
+from tinyagentos.vram_reservation import VramReservationManager
+
+
+def _mgr(free_mb: int, total_mb: int = 16384) -> VramReservationManager:
+    return VramReservationManager(probe=lambda: (free_mb, total_mb))
+
+
+def _task(priority=Priority.BACKGROUND, submitter="t"):
+    async def payload(_res):
+        await asyncio.sleep(0.05)
+        return "ok"
+    return Task(capability=Capability.LLM_CHAT, payload=payload,
+                preferred_resources=[], priority=priority, submitter=submitter)
+
+
+@pytest.mark.asyncio
+async def test_submit_gpu_defaults_backward_compatible():
+    arbiter = GpuArbiter(vram_reservation=_mgr(8192))
+    result = await arbiter.submit_gpu(_task(), required_vram_mb=1024)
+    assert result == "ok"          # old call shape, no new kwargs
+
+
+@pytest.mark.asyncio
+async def test_queue_position_global_for_loads():
+    arbiter = GpuArbiter(vram_reservation=_mgr(0))   # everything queues
+    t1, t2, t3 = _task(), _task(), _task()
+    f1 = asyncio.ensure_future(arbiter.submit_gpu(
+        t1, required_vram_mb=1024, op="load", model="a", backend_name="b1"))
+    f2 = asyncio.ensure_future(arbiter.submit_gpu(
+        t2, required_vram_mb=1024, op="load", model="b", backend_name="b1"))
+    f3 = asyncio.ensure_future(arbiter.submit_gpu(
+        t3, required_vram_mb=1024, op="load", model="c", backend_name="b1"))
+    await asyncio.sleep(0.05)      # let them enqueue
+    assert arbiter.queue_position(t1.id) == 1
+    assert arbiter.queue_position(t2.id) == 2
+    assert arbiter.queue_position(t3.id) == 3
+    for f in (f1, f2, f3):
+        f.cancel()
+
+
+@pytest.mark.asyncio
+async def test_queue_position_per_model_for_inference():
+    arbiter = GpuArbiter(vram_reservation=_mgr(0))
+    ta, tb, ta2 = _task(), _task(), _task()
+    fs = [asyncio.ensure_future(arbiter.submit_gpu(
+              t, required_vram_mb=1024, op="inference", model=m, backend_name="b1"))
+          for t, m in ((ta, "m-a"), (tb, "m-b"), (ta2, "m-a"))]
+    await asyncio.sleep(0.05)
+    assert arbiter.queue_position(ta.id) == 1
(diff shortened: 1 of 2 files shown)


hognek:
## Kilo review fixes (commit 4374588)

Three issues from Kilo at 09:58:

1. **Shadow dict drain window** — Moved `_queued_entries.pop` from before `_reserve_and_check` to after admission is confirmed. During the `asyncio.to_thread` gap, the entry remains findable by `cancel_op` via the shadow dict.

2. **Orphan task on submitter cancel** — `CancelledError` handler now adds `task_id` to `_cancelled_ids`, so the drain loop skips the entry (which remains in the real `self._queue` since `PriorityQueue` is append-only).

3. **Op validation** — `submit_gpu` now validates `op` against `{"inference", "load"}`, raising `ValueError` for unknown values so typos like `"LOAD"` don't silently fall through to the global branch.

Targeted tests: 29/29 pass.

hognek:
Applied the Kilo SUGGESTION (task t_27501d36): two fixes for the cancel_op/\_drain_queue race.

**Fix 1 — Admission path (line 603):** Re-check `_cancelled_ids` after the `_reserve_and_check` await returns. If `cancel_op` fired during the await window, the cancelled entry is discarded with reservation released rather than admitted.

**Fix 2 — Retry/re-queue path (line 641):** Same `_cancelled_ids` guard in the failed-admission retry loop. Cancelled entries that landed in the retry list during the admission-await window are skipped instead of being resurrected back into `_queued_entries`.

Both fixes follow the existing `_cancelled_ids` + `discard` pattern from the initial check at line 583.

**Tests:** 22/22 targeted (test_gpu_arbiter_queue_ops + test_gpu_arbiter_894) pass.

hognek:
Addressed Kilo WARNING: **queue processor permanently dies after MAX_CONSECUTIVE_FAILURES**.

**What changed:** Wrapped the inner drain loop in an outer restart loop. After 10 consecutive transient failures (NoResourceAvailableError, TimeoutError, OSError), the processor now waits a 300s cooldown then restarts the inner loop with a fresh failure counter — instead of returning permanently. CancelledError propagation is preserved.

**Before:** `return` on line 646 — queue processor dead forever until manual restart.
**After:** `break` from inner loop → 300s cooldown → outer loop restarts.

All 29 GPU arbiter tests + 192 scheduler/cluster tests pass.

hognek:
Rebased onto origin/dev (8a8b6cb5). Zero conflicts. 15/15 targeted tests pass (6 gpu_arbiter_queue_ops + 9 scheduler). Force-pushed to fork.

hognek:
Rebased onto origin/dev (c5b1a6fe). Two conflicts resolved in `_process_queue` — kept upstream's wake-event loop while preserving the PR's try/except error hardening + restart-after-cooldown behavior. 15/15 targeted tests pass (6 gpu_arbiter_queue_ops + 9 scheduler). Force-pushed to fork.

jaylfc:
Deep review of head 12a90ee72 complete — MERGING, with two findings carded rather than held, since neither blocks the fixes this PR ships.

**Verified fixed (commit ac5c94879):** (1) the bare `except Exception` at gpu_arbiter.py:704-710 closes the silent-queue-processor-death path — previously any non-listed exception escaped both loops and `start()` never restarted the task; it now routes into the existing cooldown-restart flow. (2) The lease-loss hang fix is complete and race-safe: `_renew_lease_loop` cancels `arbiter_future` synchronously alongside the running task (:464, :493), so `_propagate`'s cancelled-task short-circuit can no longer orphan the submitter's future. Cancellation is race-safe against admission/retry (`_cancelled_ids` re-checked at :736/:756/:793/:803). All 31 arbiter tests pass locally.

**Confirmed live, carded (not introduced by this PR's headline fixes, but in the touched code):** CodeRabbit's Aug-2 Major is real — reproduced against the unmodified `_drain_queue()`: the queue-full drop branch (:808-812) sets `NoResourceAvailableError` on the future but never pops `_queued_entries`, so `queue_position()`/`queue_snapshot()` report a phantom task forever after any queue-full drop. Only exit path with that omission (every other removal pops at :384/:784/:804/:855). Card includes the missing regression test (no test in the suite drives the drop branch — every queueing test uses 0-VRAM managers and never approaches `max_queue_size`).

**API gap carded for the fence work:** there is no way to discover or bulk-cancel *queued* ops by resource — `queue_snapshot()` omits `resource_id` and only `cancel_running_for_leases` covers the running half. The cluster fence-handling card (tsk-krklez) needs that primitive before wiring to this arbiter; noted there.

PR #1997, opened 2026-07-18
fix(github-app): snapshot installations dict before asyncio.to_thread to prevent data race

hognek:
## Problem
CodeRabbit CRITICAL on PR #1932: `_save_sync` iterates `self._installations` inside `asyncio.to_thread`. Concurrent `add()` / `remove()` calls from the event loop can mutate the dict while the thread is iterating, causing `RuntimeError` or silent corruption.

## Fix
Snapshot the installations dict into a local `data` dict **on the event-loop thread** before dispatching to `asyncio.to_thread`. `_save_sync` now receives the pre-built dict as a parameter instead of reading `self._installations` directly.

## Testing
- `test_github_app.py`: 12/12 pass
- `test_github_oauth.py`: 17/17 pass

Task: t_a3f0a632

----
## Summary by Gitar

- **GitHub App integration:**
    - Implemented `GitHubAppInstallations` registry in `github_app_installations.py` to manage installation state.
    - Added authentication logic in `github_app.py` including JWT generation and installation token minting.
    - Exposed API endpoints for managing installations in `github_oauth.py` and added webhook support in `gh_webhook.py`.
- **Frontend & Config:**
    - Added UI components in `GitHubConnect.tsx` for managing app installations and repository access.
    - Updated `config.py` to support `github_app_id` and `github_app_private_key` configuration fields.
- **Testing:**
    - Added `test_github_app.py` and expanded `test_github_oauth.py` to cover new authentication and management workflows.

### DIFF
--- a/desktop/src/apps/SecretsApp.test.tsx
+++ b/desktop/src/apps/SecretsApp.test.tsx
@@ -9,6 +9,9 @@ vi.mock("@/lib/github", () => ({
   pollDeviceFlow: vi.fn(),
   listIdentities: vi.fn().mockResolvedValue([]),
   deleteIdentity: vi.fn(),
+  listAppInstallations: vi.fn().mockResolvedValue([]),
+  beginAppInstallation: vi.fn(),
+  deleteAppInstallation: vi.fn(),
 }));
 
 function mockFetch(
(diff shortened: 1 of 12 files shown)


jaylfc:
The core race fix is correct - snapshotting the installations dict before asyncio.to_thread makes serialization race-safe, which folds the CRITICAL I flagged on #1932. Before it fully closes that finding, two things: there is a residual add/remove TOCTOU on the event loop (a remove landing between snapshot and write), so confirm the concurrent add+remove path against #1932's exact wording. And this PR also carries the broader install/uninstall UI, which picked up new Kilo WARNs worth folding: a per-installation token minted on every list call (GitHub rate-limit risk - cache or mint lazily), setup_action=='update' skipping store.add so metadata never refreshes, and a stale token cache after uninstall. Fold those and it is a clean merge.

PR #1998, opened 2026-07-18
fix(invites): failed approve no longer burns the invite + scope validation at mint (#1993)

### LINKED ISSUE #1993: Invite flow: failed approve burns the invite + mint accepts unknown scopes (Fable audit M4)
From the beta.41 pre-deploy audit. Affects the external-agent invite flow now in active use.

invite_store.py:292-298 flips the invite status to redeemed (atomic pending->redeemed CAS, race-safe) BEFORE approve_request_record runs in routes/project_invites.py:794+. Any approve failure (unknown scope, handle 409 race) leaves the invite consumed and the agent with an error, unrecoverable without re-minting.

Compounding: mint_invite (project_invites.py:608) never validates payload.scopes against VALID_SCOPES (validation exists only in agent_auth_requests.py:175/675), so an owner can mint an auto-invite whose scope can never redeem successfully but burns on first attempt.

Fix: validate scopes at mint; flip status to redeemed only after approve succeeds, keeping the atomic pending->redeemed CAS as the race gate (claim pending, approve, then commit redeemed / roll back to pending on failure).

jaylfc:
Closes #1993 (P0 for the incoming external game-dev agent onboarding).

**Burn fix:** `store.redeem()` keeps its atomic pending->redeemed CAS as the concurrent-redeem race gate, but both redeem paths (project + OS-level) now call the new `restore_pending()` when `approve_request_record` fails, returning the invite to the pool, and refuse the dangling auth request (`decided_by=system:invite-rollback`) so it does not linger in the consent inbox. `restore_pending` is guarded on `status='redeemed'` so it can never resurrect an expired/revoked invite.

**Mint validation:** both mint endpoints validate `payload.scopes` against the consent system's `VALID_SCOPES` and 400 with the unknown names, so a typo'd scope can no longer produce an invite that fails at every redeem.

Tests: unknown-scope 400 on both endpoints; simulated approve failure -> 400 -> the SAME invite+pin redeems successfully on retry. 29/29 invite tests green.

Note: lands after the beta.41 release train (this is beta.42 material unless the deploy hits invite trouble).

### DIFF
--- a/tests/test_routes_project_invites.py
+++ b/tests/test_routes_project_invites.py
@@ -163,6 +163,64 @@ async def _mint_invite(client, pid, *, approval_mode="auto", scopes=None, check_
 
 
 @pytest.mark.asyncio
+async def test_mint_rejects_unknown_scopes(client, app):
+    # #1993: an invite minted with a bad scope would silently pass and then
+    # fail (and previously burn) at every redeem. Validate at mint on BOTH
+    # endpoints.
+    pid = await _create_project(client, slug="badscope")
+    resp = await client.post(
+        f"/api/projects/{pid}/invites",
+        json={"scopes": ["a2a_send", "not_a_scope"], "approval_mode": "auto"},
+    )
+    assert resp.status_code == 400, resp.text
+    assert "not_a_scope" in resp.json()["error"]
+
+    os_resp = await client.post(
+        "/api/agents/invites",
+        json={"scopes": ["definitely_bogus"], "approval_mode": "auto"},
+    )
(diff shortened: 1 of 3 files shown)


jaylfc:
Superseded by hognek's #2002, which we are keeping - his pending->claimed->redeemed state machine is cleaner and wraps the auth_store.create() failure too, not just approve(). Closing this to avoid a same-files conflict. I have left the one piece worth porting on #2002. Thanks for taking it independently, hognek.

PR #1999, opened 2026-07-18
fix(csrf): merge CSRF token into effectiveInit for Request inputs (bot-fix for #1991)

hognek:
## Summary

Fixes the 2 inline bot findings on PR #1991 (Kilo + CodeRabbit) in `auth-guard.ts`.

### Changes

**auth-guard.ts** — Request-object CSRF handling:
- Compute effective method from `init?.method || input.method` (was: `input.method` only)
- Merge CSRF token into `effectiveInit` headers instead of rebuilding the Request with `new Request(input, { headers })`
- This fixes 3 issues:
  1. **Kilo finding**: Request inputs were previously excluded from CSRF when init was provided (init.headers would override the rebuilt request's headers)
  2. **CodeRabbit #1**: `init.method` overrides were ignored — a GET Request with `init: { method: 'POST' }` would skip CSRF
  3. **CodeRabbit #2**: `new Request(input, ...)` consumes the original body stream

**auth-guard.test.ts** — Updated Request-object assertion:
- Now checks `spy.mock.calls[1]` (init arg) instead of `spy.mock.calls[0]` (Request arg) since the token lives in effectiveInit

### Tests

- tsc: clean
- auth-guard.test.ts: 1/1 pass
- csrf.test.ts: 17/17 pass

### DIFF
--- a/desktop/src/lib/auth-guard.test.ts
+++ b/desktop/src/lib/auth-guard.test.ts
@@ -37,10 +37,11 @@ describe("installAuthGuard CSRF wiring", () => {
     const lookalike = spy.mock.calls.at(-1)?.[1] as RequestInit;
     expect(new Headers(lookalike.headers).get("X-CSRF-Token")).toBeNull();
 
-    // Request-object input for a same-origin mutating call: the wrapper rebuilds
-    // the Request with the header attached (first arg is the Request).
+    // Request-object input for a same-origin mutating call: the wrapper merges
+    // the CSRF token into effectiveInit so native fetch respects any
+    // init-provided method/headers and preserves the original body stream.
     await window.fetch(new Request(`${window.location.origin}/api/x`, { method: "POST" }));
-    const reqArg = spy.mock.calls.at(-1)?.[0] as Request;
-    expect(reqArg.headers.get("X-CSRF-Token")).toBe("tok123");
+    const reqInit = spy.mock.calls.at(-1)?.[1] as RequestInit;
+    expect(new Headers(reqInit.headers).get("X-CSRF-Token")).toBe("tok123");
   });
 });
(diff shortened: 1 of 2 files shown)


hognek:
Rebased onto origin/dev (a3ef7f77). 3 upstream commits auto-dropped (already in dev). Hogne's fix-for-Request-inputs commit re-applied cleanly. Desktop build + vitest pass.

jaylfc:
Kilo's WARNING here was **correct**, so I have folded it directly onto this branch (maintainer edit) rather than leave the PR parked.

## The bug

```js
const baseHeaders = init?.headers || input.headers;
```

This picks one source. If a caller passed headers via both the Request and the init, the Request's own headers were dropped entirely:

```js
fetch(new Request(url, { headers: { Authorization: "Bearer ..." } }),
      { method: "POST", headers: { "X-Trace": "1" } })
```

`Authorization` vanishes. Nothing downstream restores it, because init also wins at the fetch layer when a Request and an init are both supplied.

Now merged, with init winning on conflict, which matches fetch's own resolution.

## Slightly worse than reported

The regression test is verified to fail against the pre-fix code, and it shows the old line dropped **the CSRF token too** in that same call, not just `Authorization`:

```
AssertionError: expected null to be 'Bearer abc'
AssertionError: expected null to be 'tok123'
```

So in exactly the scenario this PR was extending coverage to, the header the wrapper exists to attach was itself being lost. That makes it a correctness bug in the feature, not only a header-preservation nit.

## One note for whoever touches these tests next

The new assertions sit inside the existing `it()` block deliberately. `installAuthGuard` has a module-level `installed` flag, so calling it again from a fresh `it()` is a no-op and `window.fetch` stays the bare spy rather than the wrapper. A test written the obvious way passes or fails for the wrong reasons. I lost a couple of iterations to that before spotting it, so it is called out in a comment in the file.

Thanks for the original fix; the merge-vs-pick distinction was the only thing missing.

PR #2000, opened 2026-07-18
docs(skill): refresh taos-development-skill against current repo reality

jaylfc:
Full audit of .claude/skills/taos-development-skill (SKILL.md + soul.md) against origin/dev. The pitfalls section was almost entirely still accurate; the rot was concentrated in the bot-review section, the fork-CI story, and missing material that recent churn made load-bearing.

Corrections: bot roster (Kilo + Gitar are the reliable gate; CodeRabbit unreliable; Qodo paused), finding-fetch commands fixed (bot logins have no [bot] suffix so the old jq silently dropped CodeRabbit findings and missed gitar-bot entirely, and it never read reviews[] or inline comments where most findings live), fork-CI approval is now first-time-only (returning contributors auto-run), CI description gains spa-build (npm build + tsc + vitest) and the app-start smoke, adapters line corrected, stale counts dropped.

Additions: frontend CSRF contract (withCsrf/taosFetch; raw mutating fetch passes tests but 403s in prod - the #1977 class), agent auth model (Bearer grant-gated JWT = CSRF-exempt vs session + CSRF header), adapter verification_status vocabulary, 4-file version sync note.

Deletions: duplicated Pico/htmx guidance collapsed to one line, time-estimate/difficulty filler cut, issue-triage trimmed to the taOS-specific rules. soul.md rule 3 now extends no-AI-attribution/no-em-dash to all public-facing text.

Skill-only change (docs), beta.42 material.

## Summary by CodeRabbit

* **Documentation**
  * Updated contribution and fork-based CI approval guidance, including clarified first-time vs returning contributor paths and draft/CI deadlock handling.
  * Added explicit CI-required validation steps and revised the bot review cycle to emphasize the required review gates and strengthen checks around current head SHA.
  * Expanded guidance on frontend CSRF behavior, agent authentication model, adapter verification vocabulary, and version sync rules.
  * Refined issue triage rules and simplified legacy template styling and conventional-commit/no-AI-attribution guidance.

PR #2001, opened 2026-07-18
docs(skill): fix bot suffix claim and jq null guard (Kilo findings on #2000)

hognek:
Fixes 2 Kilo Code Review findings on PR #2000:

1. **WARNING (SKILL.md:194):** Corrected the claim that bot logins have NO `[bot]` suffix — reviews DO carry it (`kilo-code-bot[bot]`, `gitar-bot[bot]`), while issue comments may show logins without it. Updated the jq to handle both forms.

2. **SUGGESTION (SKILL.md:199):** Added null-safe `?.author.login?` with `.user.login?` fallback to the jq pipeline — prevents pipeline abort when a review/comment has a null author (deleted user). Also handles the reviews-vs-comments login field difference (.user.login vs .author.login).

Tests: n/a (docs-only change). Target base: docs/skill-refresh (original PR branch).

### DIFF
--- a/.claude/skills/taos-development-skill/SKILL.md
+++ b/.claude/skills/taos-development-skill/SKILL.md
@@ -191,12 +191,15 @@ findings **before** surfacing the PR for human maintainer review.
 ### Procedure
 
 1. **Push PR and mark ready.** Wait ~10 minutes for bot reviews to complete.
-2. **Pull bot findings - reviews AND inline comments, all bots.** Bot logins have NO `[bot]`
-   suffix in the JSON (`kilo-code-bot`, `gitar-bot`, `coderabbitai`, `qodo-code-review`):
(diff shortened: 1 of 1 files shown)


jaylfc:
Thanks hognek, but closing this one - the bot-suffix prose in #2000 is actually correct for the command it documents. The suffix is surface-dependent: gh pr view --json (GraphQL) returns bare logins (kilo-code-bot), while gh api .../comments (REST) returns the [bot]-suffixed form. #2000 documents the GraphQL command, so 'no [bot] suffix' is right there; flipping it to carry [bot] would break that exact command. I will fold your .author.login? // .user.login? // '' null-guard into #2000 since that part is a genuine improvement. Appreciate the catch on the guard.

Issue #1994, opened 2026-07-18
Fable audit: minor findings + rapid-iteration dead code (M5/L1-L5)


jaylfc:
Low-severity findings from the beta.41 pre-deploy audit; batch cleanup. M5: user_shares_store.py (#1897) unwired dead code. L1: _backfill_lead_member_id (project_store.py:88-110) should clear is_lead flags on 0/many branches too. L3: 6-digit invite ids -> collision 500 (add retry); GET /i/{id} auth-exempt/enumerable leaks project name/scopes on LAN (PIN caps keep redemption safe). L4: dead code - _CANCEL_SCOPES (project_invites.py:77), gated-off UsernameStep (~120 lines, calls nonexistent /api/account/username), taosFetch (zero call sites). L5: latent - stores sharing one DB file share one schema_migrations table keyed by int version; first shared-file store to add MIGRATIONS collides. Migration safety / CSRF backend coverage / invite-grant scoping / SSRF all verified CLEAN.

PR #2003, opened 2026-07-18
fix: Fable audit dead-code cleanup (M5, L1-L5)

hognek:
Issue #1994. Low-severity findings from beta.41 pre-deploy audit.

- **M5:** Removed unwired user_shares_store.py + tests (dead code from #1897)
- **L1:** Clear is_lead flags on ALL branches in _backfill_lead_member_id (0/many, not just exactly-1)
- **L3:** Add collision retry (max 5) for 6-digit invite IDs in InviteStore.mint
- **L4:** Removed dead code: _CANCEL_SCOPES (project_invites.py:77), UsernameStep + gating (~140 lines, OnboardingScreen.tsx), taosFetch (BackendStatusContext.tsx)
- **L5:** Namespace-aware schema_migrations table — store_name column scopes migration versions so stores sharing one DB file don't collide

Tests: 58/58 targeted pass (db_migrations, base_store, project_store, invite_store, check_schema_migrations). No behavior changes.

## Summary by CodeRabbit

* **New Features**
  * Database migrations are now tracked independently for each storage area, improving reliability when multiple services share a database.
  * Invitation creation now retries automatically if a temporary identifier conflict occurs.

* **Changes**
  * Onboarding now completes immediately after successful account creation; username claiming is deferred.
  * The user-to-user sharing feature and related storage support have been removed.


jaylfc:
Thanks for taking #1994 - this faithfully covers M5/L1/L3/L4/L5 with no scope creep. One gate before it can merge: the L5 legacy schema_migrations PK-rebuild uses executescript, whose implicit commit makes the DROP/RENAME sequence NON-atomic - a crash mid-rebuild can leave a DB with no schema_migrations table, which is exactly the upgrade-bricks-boot failure mode we have hit repeatedly. Please wrap that rebuild in a single explicit transaction (BEGIN/COMMIT, no executescript) so it is all-or-nothing, and test it over a real pre-namespace DB (copy a beta.40 store, run the upgrade, confirm boot). Also two bot nits worth folding: the invite-id retry catches all IntegrityError + the fragile 'UNIQUE' in str(exc) check (narrow to the unique-constraint case), and the async _ensure_namespace_column_async bare except Exception (the sync twin narrows it). Green + those folded and I will merge.

PR #2026, opened 2026-07-18
fix(security): address 8 CodeRabbit findings — aliasing, permissions, race, snapshot, signing-failure gate, 422 test

hognek:
fix(security): address 8 CodeRabbit findings on PR #2023 store security

### store_signing.py (3 fixes)
- **Docstring correction:** Document detached in-memory signatures (stored in AppRegistry._signatures, not YAML). Clarify the post-load tamper-detection model.
- **\_enforce\_permissions fail-closed:** Raise PermissionError instead of swallowing OSError, so a keyfile with unenforceable permissions is not read.
- **Race fix in keypair creation:** Use os.link() for non-overwriting atomic claim instead of os.replace(). A late contender that loses the race loads the winner's keypair.

### app.py (2 fixes)
- **Import aliasing:** Alias \ as \ to prevent shadowing by the agent_registry_store import of the same name. The catalog-signing path now correctly uses the store-signing key (store_signing_key.json).
- **Move signing key init into lifespan:** Previously ran synchronously during create_app() (triggering registry.reload()). Now runs in lifespan() before \, so tests bypassing the lifespan get store_signing_pubkey=None.

### registry.py (2 fixes)
- **\_CatalogState snapshot:** Bundle catalog, signatures, manifest_dicts, and signing_failures into one immutable dataclass snapshot that is atomically replaced. Prevents TOCTOU where a reader could see a new manifest from \_catalog paired with an old \_signatures dict.
- **Signing failure tracking:** Track manifests that failed sign_manifest() in \. Expose via \ so the install gate can block them (rather than treating them as merely unsigned, which would bypass the gate).

### store_install.py (1 fix)
- **Check is_signing_failure()** in \: When get_signature() returns None, additionally check if the manifest failed to sign during load. If so, return 403 — a signing failure is not the same as "never signed" and must not bypass the gate.

### tests (1 fix)
- **test_unknown_backend_returns_422:** Updated from expecting 500 to 422, matching the production code at store_install.py:975.

## Test results
- store_signing: 13/13 pass
- store_install (key tests): 5/5 pass — unknown_backend_422, tampered_403, valid_signature, real_registry_tampering, no_signing_key_skips

Fixes: CodeRabbit 7 inline + 1 outside-diff findings on #2023
Related: #1924

### DIFF
--- a/tests/test_routes_store_install.py
+++ b/tests/test_routes_store_install.py
@@ -526,6 +526,7 @@ class TestInstallV2:
         # 2. Build a real AppRegistry with a signing key.
         priv, pub = generate_signing_keypair()
         installed_path = Path(tempfile.mkstemp(suffix=".json")[1])
+        installed_path.write_text("[]")  # initialise with valid JSON
         reg = AppRegistry(
             catalog_dir=catalog_dir,
             installed_path=installed_path,
--- a/tests/test_routes_store_install.py
+++ b/tests/test_routes_store_install.py
@@ -430,6 +430,129 @@ class TestInstallV2:
         assert "compat" in body
         assert "chain" in body
 
+    # ------------------------------------------------------------------
+    # Code-signing tests (#647)
+    # ------------------------------------------------------------------
+
+    @pytest.mark.asyncio
+    async def test_tampered_manifest_rejected_403(self, client):
+        """When registry.verify_manifest_signature returns False, the
+        install is rejected with 403."""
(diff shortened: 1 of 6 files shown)


hognek:
Folded into #2023

hognek:
Superseded by #2023. The 500→422 test assertion fix was consolidated into #2023 per jaylfc's fold instruction.

PR #2036, opened 2026-07-19
feat(secrets): per-agent GitHub token grant UI + short-lived token storage

hognek:
Task: t_2f795122. Part of #858 (C2).

## Changes

- **New module:** `tinyagentos/github_token.py` — lifetime-aware token minting cache with `mint_installation_token()` and `get_agent_github_token()`
- **SecretsStore:** Added `github-installation` category + `get_agent_github_installations()` method
- **API:** `GET /api/secrets/agent/{name}/github` endpoint for per-agent GitHub grant queries
- **Frontend:** 
  - `github-installation` category in SecretsApp category filter, styles, and dropdowns
  - Per-agent access toggles for each GitHub App repo in GitHubConnect.tsx
- **Tests:** 7/7 `test_github_token.py`, 6/7 `test_secrets.py` pass

## Summary by CodeRabbit

* **New Features**
  * Added support for GitHub Installation secrets, including creation, editing, filtering, and secure agent access.
  * GitHub repository connections now allow configuring agent grants per repository.
  * GitHub installation details now display repository permissions.
  * Added automatic GitHub installation token generation, caching, and refresh.
  * Secret names can now include path-style characters.

* **Security**
  * Secret values remain encrypted at rest and are masked in listings.
  * Access to agent GitHub grants requires authentication.

### DIFF
--- a/desktop/src/apps/SecretsApp.tsx
+++ b/desktop/src/apps/SecretsApp.tsx
@@ -24,7 +24,7 @@ interface Secret {
   revealed?: boolean;
 }
 
-type CategoryFilter = "all" | "api-key" | "credential" | "token" | "config";
+type CategoryFilter = "all" | "api-key" | "credential" | "token" | "config" | "github-installation";
 
 /* ------------------------------------------------------------------ */
 /*  Constants                                                          */
@@ -35,6 +35,7 @@ const CATEGORY_STYLES: Record<string, string> = {
   credential: "bg-cyan-500/20 text-cyan-400",
   token: "bg-amber-500/20 text-amber-400",
   config: "bg-emerald-500/20 text-emerald-400",
+  "github-installation": "bg-purple-500/20 text-purple-400",
 };
 
 const MASKED = "\u2022\u2022\u2022\u2022\u2022\u2022\u2022\u2022";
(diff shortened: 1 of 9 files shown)


jaylfc:
Good direction on per-agent GitHub tokens. Fold list before merge (test 3.13 is failing too):

MUST-FIX (security):
1. secrets.py:42 - GET /api/secrets/agent/{agent_name}/github returns installation IDs + repo names with no visible auth/authorization guard. This leaks which repos an agent can reach to any caller. Gate it (admin session + CSRF, or the agent-token allowlist) the same way the other secrets routes are gated, and add a test that an unauthenticated/cross-agent caller gets 401/403.

MUST-FIX (correctness):
2. github_token.py:147 - get_agent_github_token silently uses installations[0] and drops every other granted installation. If an agent is granted multiple installs this picks one arbitrarily. Resolve the correct installation for the target repo/owner rather than index 0 (this is the same installations[0] pitfall #1997 addressed on the store side).
3. GitHubConnect.tsx:462 - handleSaveGrants is always called with permissions: [] hardcoded, so grants save with empty permissions. Pass the real selected permissions.

Once 1 + 2 are folded and CI is green I will re-review. Note secrets are security-critical so I will deep-review the auth path specifically.

jaylfc:
Adding to the fold list: your branch has a failing test of its own in CI: tests/test_secrets.py::TestGitHubInstallationCategory::test_update_secret_agents asserts 200 but gets 404, so the new installation-category update route either is not registered where the test expects or the path differs. This is on top of the auth-guard and installations[0] items above. Separately, dev itself had one red test from a #2009/#1932 semantic conflict; fix is up as #2041, so rebase onto dev once that merges and your base will be green again.

PR #1944, opened 2026-07-17
feat(todo): add TodoStore + /api/todo routes with ordering and due dates

hognek:
Task: t_60ad855b. Part of #1923 Notes/Todo split (C1: backend).

Implement Todo data model and store based on C0 design doc:
- **TodoStore** (`tinyagentos/todo/todo_store.py`): BaseStore subclass with todo_lists + todo_items tables, full CRUD, position-based ordering, batch reorder, optional due_at/remind_at
- **Routes** (`tinyagentos/routes/todo.py`): /api/todo/* endpoints — list/create/patch/archive lists, add/patch/delete items, PUT reorder, ISO-8601 date parsing
- **Wired** into app.py (lifespan + all 3 state-assignment points) and routes/__init__.py

Tests: 18/18 pass (CRUD, ownership, reorder, due dates, 404s). Notes tests still green (27/27).

## Summary by CodeRabbit

* **New Features**
  * Added a Todo REST API with todo lists and checklist items, including create, edit, completion toggle, delete, archiving, and item reordering.
  * Added due dates and reminder timestamps (ISO-8601 parsing with validation and clear handling).
  * Enforced ownership/authorization for list and item operations, with consistent 403 behavior.
  * Added not-found responses (404) for missing lists and items.

* **Tests**
  * Added comprehensive end-to-end API tests covering CRUD, reordering, authorization, and due-date validation/error cases.

PR #2033, opened 2026-07-19
feat(todo): Notes/Todo split — final integration pass (C5)

hognek:
Closes #1923.

## What this does
Completes the Notes/Todo split with the final integration pass:

- **TodoApp** — Dedicated checklist component with create/complete/reorder/due-date/per-item persistence
- **TodoStore** — SQLite store (tinyagentos/todo/) with list+item tables and ownership checks
- **/api/todo routes** — Full REST API (CRUD lists, CRUD items, reorder) with CSRF protection
- **NotesApp** — Cleaned to notes-only (todos live in TodoApp), shares the SharedDocsStore
- **App registry** — Both `notes` (launchpadOrder 16.8) and `todo` (16.85) registered

## C5 checklist
- [x] Both apps register in app-registry.ts
- [x] Todo tests — create, complete, reorder, persistence (9 vitest + 18 pytest)
- [x] Notes tests — create, edit, persistence, share (9 vitest + 27 pytest)
- [x] No kind branching remains in shared code
- [x] Desktop build (tsc + vite) passes clean
- [x] CSRF dependency on todo_router (was missing — fixed in C5)

## Commits
- `ad96c709` feat(todo): TodoStore + /api/todo routes
- `b929f6f2` feat(todo): TodoApp component
- `a6260f01` fix(todo): Kilo review fixes
- `4002cf77` fix(todo): reorder + validate
- `e26bade0` fix(todo): duplicate submissions, toggle serialization
- `90141ac8` fix(todo): CSRF dependency on todo_router

## Summary by CodeRabbit

- **New Features**
  - Notes and Todo are now separate apps with dedicated interfaces.
  - Todo lists support creating lists, adding and editing tasks, completion tracking, deletion, due dates, and reordering.
  - Todo task reordering now validates the full list and preserves consistent ordering.

- **Bug Fixes**
  - Opening the Todo app now loads the correct Todo interface.
  - Invalid task reorder requests return a clear error instead of being partially applied.
  - Todo requests now receive standard security protection.

PR #2037, opened 2026-07-19
refactor(desktop): split NotesApp into notes-only component with shared module

hognek:
## Summary

Split the monolithic NotesApp.tsx into a shared module + two thin app wrappers:

1. **notes-shared-base.tsx** — shared DocsApp component, sub-components (EntryHistoryPanel, SharePanel, EntryRow, NoteListItem, CreateNoteForm, NoteDetailPane, MemberBadge), types, and helpers. Extracted from the original NotesApp.tsx.

2. **NotesApp.tsx** — thin notes-only wrapper (33 lines). Hardcodes kind="note" config. No more kind-branching or TodoApp export.

3. **TodoApp.tsx** — standalone todo component. Uses the dedicated /api/todo store. Originally created alongside PR #1944 but never committed.

4. **app-registry.ts** — todo app now points to `@/apps/TodoApp` instead of importing TodoApp from `@/apps/NotesApp`.

## Data migration

Existing docs with kind="note" continue to work without changes. Docs with kind="list" are handled by TodoApp via /api/todo. No database migration required.

## Tests

- All 14 Notes + Todo unit tests pass
- All 9 app-registry tests pass
- Test imports updated: NotesApp from NotesApp.tsx, TodoApp from TodoApp.tsx
- Todo tests updated to mock /api/todo instead of /api/notes

Task: t_f5920047


jaylfc:
Fold list before merge:

1. TodoApp.tsx:257 (functional regression) - the standalone Todo loses the Share/members panel and per-entry revision history that the old DocsApp-based list variant provided. Either carry those over or, if intentionally dropped for slice 1, call it out explicitly in the PR + a follow-up issue so we do not silently lose sharing on todos.
2. TodoApp.tsx:348 - loadLists swallows fetch failures in an empty catch and only flips loading to false, so a 500/network error shows an empty list with no error state. Surface the error (toast or inline) rather than failing silently.
3. NotesApp.test.tsx:24 - Todo tests now live in two files (__tests__/NotesApp.test.tsx and the top-level NotesApp.test.tsx) both importing the same thing. Consolidate to one to avoid drift.

Once folded and CI green I will merge. Big refactor so I will diff the extracted components for behavior parity.

jaylfc:
@hognek Held pending a direction call on #2033. Both this and #2033 **add** `desktop/src/apps/TodoApp.tsx`, so they cannot both land; whichever goes first makes the other an add/add conflict.

For what it is worth, this is the more interesting change of the two: pulling 1160 lines out of NotesApp into `notes-shared-base.tsx` and rebuilding TodoApp at 402 lines against it is a much better factoring than a second 898-line standalone app, **if** the shared behaviour is genuinely stable. The risk is that it is a large refactor of a working app, and the shared base becomes a coupling point the moment Notes and Todo start diverging.

Your call, on #2033. Note this PR has no TodoApp.test.tsx while #2033 carries a 473-line one, so if this direction wins the tests need to come across.

jaylfc:
Closing for now, per the decision on #2033 - not because this is the worse change.

Both PRs **add** `desktop/src/apps/TodoApp.tsx`, so only one can land, and #2033 goes first because it carries the backend and a 473-line test suite while this one is frontend-only with no TodoApp tests.

**I want this factoring back as a follow-up.** Pulling 1160 lines of shared behaviour out of NotesApp into `notes-shared-base.tsx` is the better end state, and a 402-line TodoApp beats a 898-line one. Reopen it against dev once #2033 lands, as a pure refactor carrying `TodoApp.test.tsx` across unchanged. That way the tests prove the refactor preserves behaviour, which is exactly the evidence a change of this size needs and cannot have while it is also introducing the feature.

@hognek this is a sequencing call, not a rejection.

hognek:
## Merge order resolution — jaylfc's decision

This PR is **superseded by #2033**. This approach gutted NotesApp by ~1160 lines; #2033's standalone TodoApp approach is the safer design. The Notes/Todo split is handled more cleanly in #2033 with a dedicated TodoApp component rather than stripping functionality from the existing NotesApp.

Closing in favor of #2033.

PR #2032, opened 2026-07-19
fix(peer): address Kilo WARNINGs — os.environ leak, design doc, rate limiter docs

hognek:
## Summary

Addresses three Kilo WARNINGs carried over on #2025:

1. **app_with_contacts fixture** — mutated `os.environ["TAOS_DATA_DIR"]` without restore. Fixed by using `monkeypatch.setenv` (scoped, auto-restored). Also removed now-unused `import os`.

2. **outbound_token plaintext** — design doc (`cross-user-collaboration.md`) claimed `encrypted at rest` but code says `stored in plaintext; encryption deferred to post-MVP`. Aligned design doc to match code.

3. **Rate limiter** — added explicit `FIXME(post-MVP)` for shared-store backing across workers, and documented the per-worker aggregate limit semantics (60×N/min across N workers).

## Testing

- 37/37 peer tests pass (`pytest tests/test_contacts_peer.py -v`)
- Canonical gate running in background

## Related

- PR #2025 (feat/collab-a1-contacts-peer)
- Kilo inline findings from 2026-07-19

## Summary by CodeRabbit

* **New Features**
  * Added secure peer-to-peer communication through authenticated inbox, chat, and acknowledgement endpoints.
  * Added contact and peer-link management, including token authentication, status tracking, revocation, and replay protection.
  * Added signed message validation with freshness, recipient, sender, and size checks.
  * Added rate limiting for peer requests.

* **Documentation**
  * Clarified peer token storage and encryption plans in the design documentation.

* **Tests**
  * Added comprehensive coverage for contact management, peer messaging, authentication, validation, replay protection, and rate limiting.

PR #2046, opened 2026-07-19
feat(collab): friend-accept creates contact row + peer link + handshake

hognek:
## A2: Friend-accept → contact row + peer-link handshake

On hub friend-accept, creates the contact row (pinning Ed25519/X25519 pubkeys from the directory response), mints the inbound peer token, fires a background handshake envelope to the peer's endpoints, and records the peer_link. Block cascades to revoke_peer_link.

### Changes
- **routes/hub.py**: Extend `accept_friend_request` to create contact row, mint peer token, and deliver handshake (fire-and-forget via `asyncio.create_task`). Extend `block_peer` to cascade to `contacts_store.revoke_peer_link` via fingerprint→username lookup through hub_authors.
- **peer.py**: Add `send_handshake()` and `deliver_handshake()` for building and delivering handshake envelopes to peer endpoints.
- **tests**: 4 new tests in `test_routes_hub_slice3.py` — accept creates contact, accept creates peer_link, accept without contacts_store doesn't crash, block cascades to revoke peer link. All 53 targeted tests pass (hub + contacts + peer).

### Task
- Epic: #2012 (Cross-user collaboration)
- Issue: #2014
- Depends on: #2025 (A1 — merged)

### Test results
```
53 passed in 253.34s — test_contacts_peer.py (37) + test_routes_hub_slice3.py (16)
```

### Design notes
- Handshake delivery is fire-and-forget (`asyncio.create_task`) — we don't block the HTTP response waiting for the peer
- `outbound_token` starts empty in the initial peer_link; it's filled when the peer responds to our handshake
- Block cascade uses `store.get_author(fingerprint)` to resolve peer fingerprint → hub username → contact_id
- If the directory response doesn't include pubkeys, the contact row is still created (with empty strings) — the peer link handshake will fill them

### DIFF
--- a/tests/test_routes_hub_slice3.py
+++ b/tests/test_routes_hub_slice3.py
@@ -275,3 +275,188 @@ class TestPresenceGate:
         assert resp.json()["endpoints"] == ["wss://x", "wss://y"]
         fwd = [c for c in captured if c["url"].endswith("/api/hub/presence?username=alice")]
         assert fwd, "presence was not forwarded to the directory"
+
+
+# ---------------------------------------------------------------------------
+# Milestone A2: friend-accept -> contact row + peer link + block cascade
+# ---------------------------------------------------------------------------
+
+
+class TestFriendAcceptContactCreation:
+    """Friend-accept creates a contact row and peer link."""
+
+    @pytest.mark.asyncio
+    async def test_accept_creates_contact_row(self, client, upstream):
+        """Friend-accept with identity info creates a contact row in contacts_store."""
+        from tinyagentos.hub import identity as idmod
+
+        # Ensure we have a local hub identity so the handler can run.
+        idmod.load_or_create()
+
+        # Initialise contacts_store — the lifespan may not have done this in tests.
+        contacts = client._transport.app.state.contacts_store
+        if contacts._db is not None:
+            await contacts.close()
+        await contacts.init()
+
+        async def handler(method, url, **kw):
+            return _FakeResp(content=json.dumps({
+                "peer": "peerFP",
+                "username": "hogne",
+                "display_name": "Hogne",
+                "signing_pubkey": "ab" * 32,
+                "encryption_pubkey": "cd" * 32,
+                "endpoints": [],
+            }).encode())
+        _set(upstream, handler)
+        resp = await client.post(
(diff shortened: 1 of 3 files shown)


jaylfc:
This and #2043 are both the A2 friend-accept slice from sibling branches (feat/collab-a2-friend-accept vs feat/collab-a2-friend-accept-handshake). This one looks like the superset (adds peer.py wiring + the fuller test file). Please close #2043 if this supersedes it, or tell me what #2043 carries that this does not - one PR per slice (pitfall 13).

jaylfc:
Three Kilo findings to fold before merge (CI is green and the token-hash storage is right, so these are the only gate):

1. hub.py:484 MUST-FIX: the block cascade only revokes the peer link when hub_authors already has a row mapping the blocked fingerprint to a username. In the normal flow that row is not guaranteed to exist, so blocking a peer leaves their link live. Resolve the contact by pinned fingerprint on the contact row itself (contacts store) rather than via hub_authors, and add a test blocking a peer who was never upserted as an author.
2. peer.py:199 MUST-FIX: send_handshake calls resolve_local_identity_id() with no data_dir, falling back to TAOS_DATA_DIR env or the default path. Thread the app data_dir through like the receive path does (same env-coupling class as the #2032 item).
3. peer.py:241: the handshake envelope carries the plaintext inbound_token. The locked design says hub-relayed envelopes are X25519-sealed from day one - confirm this handshake actually goes through the sealed path (and add an assertion or test that the wire form is sealed), or seal it. If it is already sealed at the relay layer, document that at the call site so the next reader does not re-flag it.

Also: please close #2043 or state what it carries beyond this PR - still open from my earlier ask. Fold these and A2 merges.

hognek:
Superseded by #2043.

Both PRs implement the A2 friend-accept slice and Kilo flagged the same block-cascade issue on both (the hub_authors lookup returns None because accept_friend_request never caches the peer fingerprint→username mapping, so revoke_peer_link is silently skipped). #2043 has the more comprehensive implementation with 8 tests (vs 4), additional findings to address, and jaylfc's preference.

Closing in favor of #2043.

hognek:
Superseded by #2043. These are sibling A2 friend-accept branches per jaylfc — #2043 is the canonical one.

PR #2047, opened 2026-07-19
feat(collab): E1 — DM schema migration (remote_msg_id, delivered_at, peer_outbox)

### LINKED ISSUE #2020: Collab E1 (lead): human DM over peer channel (schema migration + outbox)
chat_channels type=dm-remote; chat_messages remote_msg_id + delivered_at (author_type stays user, author_id namespaced hub:{name}); peer_outbox store-and-forward + ack path. Migration MUST be tested over an existing pre-change DB (boot-brick rule).

Part of #2012 (cross-user collaboration). Spec: docs/design/cross-user-collaboration.md.

hognek:
## Summary

Implements E1 of the cross-user collaboration epic (#2012): human DM schema migration and peer outbox for store-and-forward delivery.

Closes #2020.

### Schema migration (boot-brick safe)
- **chat_messages**: Added `remote_msg_id TEXT` and `delivered_at REAL` columns via guarded `_post_init` ALTER TABLE pattern (same as existing `deleted_at`/`expires_at`)
- **chat_messages**: Added `UNIQUE INDEX` on `(channel_id, remote_msg_id) WHERE remote_msg_id IS NOT NULL` for idempotent remote delivery
- The index is created in `_post_init` (not SCHEMA) per BaseStore boot-brick rule

### send_remote_message()
- New method on `ChatMessageStore` — atomic INSERT with dedup
- Catches `IntegrityError` on duplicate `(channel_id, remote_msg_id)` and returns `None`
- Sets `delivered_at` on receipt, `author_id` namespaced as `hub:{username}`

### PeerOutboxStore
- New store: `peer_outbox` table with exponential backoff retry schedule (60s → 120s → 300s → 600s → 1800s cap)
- Methods: `enqueue`, `dequeue_due`, `mark_sent`, `mark_failed`, `purge_for_contact`, `count_for_contact`
- Attached at `app.state.peer_outbox`

### Tests
- 18 new tests in `tests/test_chat_remote_dm.py`
- Upgrade migration tested against pre-change v0 DB (boot-brick regression gate)
- Dedup: duplicate remote_msg_id returns None, same remote_msg_id different channel both succeed
- Outbox: enqueue/dequeue, sent removal, backoff retry, purge, count, limit, future retry

### All tests pass
```
100 passed (18 new + 82 related/regression)
```

## Summary by CodeRabbit

* **New Features**
  * Added support for receiving remote direct messages with duplicate detection and preserved message metadata.
  * Added a persistent peer delivery outbox for queued messages, retry scheduling, delivery tracking, and cleanup.
  * Existing chat databases are upgraded automatically while preserving stored messages.
  * Added reliable startup and shutdown handling for the peer delivery outbox.

* **Bug Fixes**
  * Prevented duplicate remote messages from being stored.
  * Added bounded retries with backoff for messages that cannot be delivered.

### DIFF
--- a/tests/test_chat_remote_dm.py
+++ b/tests/test_chat_remote_dm.py
@@ -227,6 +227,37 @@ class TestSendRemoteMessage:
             metadata={"hops_since_user": 0, "delivered_via": "peer"},
         )
         assert msg["metadata"]["delivered_via"] == "peer"
+        assert msg["metadata"]["contact_id"] == "hub:hogne"
+
+    async def test_contact_id_stored_in_metadata(self, msg_store):
+        msg = await msg_store.send_remote_message(
+            channel_id="dm-remote-1",
+            author_id="hub:jaylfc",
+            contact_id="hub:jaylfc",
+            remote_msg_id="rm-contact-test",
+            content="contact id test",
+        )
+        assert msg["metadata"]["contact_id"] == "hub:jaylfc"
+
+    async def test_empty_remote_msg_id_rejected(self, msg_store):
+        with pytest.raises(ValueError, match="remote_msg_id must be a non-empty string"):
+            await msg_store.send_remote_message(
+                channel_id="dm-remote-1",
+                author_id="hub:hogne",
+                contact_id="hub:hogne",
+                remote_msg_id="",
+                content="empty remote_msg_id",
+            )
+
+    async def test_null_like_remote_msg_id_rejected(self, msg_store):
+        with pytest.raises(ValueError, match="remote_msg_id must be a non-empty string"):
+            await msg_store.send_remote_message(
+                channel_id="dm-remote-1",
+                author_id="hub:hogne",
+                contact_id="hub:hogne",
+                remote_msg_id=None,
+                content="None remote_msg_id",
+            )
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


jaylfc:
Reviewed properly rather than on the green tick. **Merging.**

The migration is done right, and it is worth saying so explicitly because this is the area that has bricked our boot three times: columns added in `_post_init` rather than `SCHEMA`, `ADD COLUMN` idempotent on re-run, the dedup index created after the columns exist with a comment explaining exactly why it cannot live in `SCHEMA`, and index creation failing soft with a warning when pre-existing data would violate uniqueness. The tests seed a **v0 schema** and migrate it, which is the upgrade-over-an-existing-DB case rather than the fresh-install case that usually gets tested and proves nothing.

Two follow-ups, neither blocking:

**1. The bare `except Exception: pass` on both `ADD COLUMN`s is too wide.** It is meant to swallow "duplicate column name" on re-run, but it also swallows a locked DB, a full disk, or a corrupt table. When that happens the store carries on without the column and the failure resurfaces later as an INSERT error inside `send_remote_message`, at which point the cause is a long way from the symptom. Narrow it to the duplicate-column case and let anything else raise. Fail-soft is right for the index (pre-existing data can legitimately violate it) but not for the columns, which the feature cannot work without.

**2. Em dash in the log string** on the dedup-index warning. Public repo, so we keep those out. Use a plain hyphen.

@hognek please fold both into your next PR touching this file rather than opening one for two lines.

PR #2027, opened 2026-07-18
fix(store): update test assertions 500→422 for unknown backend

hognek:
## Summary

PR #2023 changed the `_BACKEND_TO_METHOD` lookup failure from 500 to 422 (jaylfc requested 4xx). Update both test assertions to match.

### Changes
- `tests/routes/test_store_install_v2.py`: `test_unknown_backend_returns_500_not_exception` — assertion 500→422 + docstring update
- `tests/test_routes_store_install.py`: `TestInstallV2::test_unknown_backend_returns_500` — assertion 500→422 + docstring update

### Dependency
Depends on #2023 (which changes the code from 500 to 422). CI will be red until #2023 merges.

### Tests
2/2 targeted tests pass locally against the PR branch.

Kanban: t_7353a27e

----
## Summary by Gitar

- **Security features:**
    - Implemented Ed25519 manifest signing and verification to detect catalog tampering in `store_signing.py`.
    - Added a `_verify_manifest_for_install` gate to `store_install.py` to prevent installing tampered manifests.
    - Added a TOCTOU guard that re-reads and re-verifies the manifest on disk immediately before installation.
- **New API:**
    - Exposed `GET /api/store/signing-pubkey` to allow external auditing of catalog integrity.
- **Refactor:**
    - Updated `AppRegistry` to support manifest signature tracking and re-verification logic.
    - Enabled lazy loading of keypairs in the lifespan to prevent startup failures on read-only filesystems.

<sub>This will update automatically on new commits.</sub>

PR #2050, opened 2026-07-19
fix(store): fail-closed on sign_manifest failure — prevent install-gate bypass

hognek:
## Summary

Fix Kilo WARNING from PR #2027: When `sign_manifest` raises, the exception was logged but no signature was stored. `_verify_manifest_for_install` saw `None` and short-circuited to `(True, None)` — allowing install with zero tamper protection for any manifest whose signing threw. An attacker inducing a signing failure for a target manifest could bypass the Ed25519 gate silently.

### Changes

- **`registry.py`**: Added `_signing_failures: set[str]` tracking + `had_signing_failure()` method. When `sign_manifest` raises during catalog load, the app_id is recorded in the failure set.
- **`store_install.py`**: In `_verify_manifest_for_install`, when `stored_sig is None`, check `had_signing_failure()` first and return `(False, msg)` if signing failed — block the install gate rather than silently allowing it.

### Tests

- 30/30 registry + signing tests pass
- No new test regressions in store_install suite

### Reference

Kilo: `sign_manifest` failures silently downgrade to unsigned (PR #2027 review on registry.py:141)

Kanban: t_7f43f306

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app installation reliability by running the install-time manifest signature re-check without blocking other activity.
  * Install now fails with an authorization error (and includes an `install_id`) if the manifest was modified after the initial verification step.
  * Stricter handling of missing or invalid stored signatures to prevent unsafe installs.
  * Made manifest signing/verification output more consistent.
* **Tests**
  * Added coverage for install-time manifest tampering and refusal-path scenarios.

### DIFF
--- a/tests/test_routes_store_install.py
+++ b/tests/test_routes_store_install.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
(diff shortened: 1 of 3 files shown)


jaylfc:
This has gone to CONFLICT against `dev` and I would like it unblocked ahead of the others, because of what it is: fail-closed on `sign_manifest` failure preventing an install-gating bypass. A security fix sitting in conflict is the one kind of stale PR that gets worse with time, since every day it is not merged is a day the bypass is live on dev.

Please rebase onto current `origin/dev` (it moved tonight: #2171, #2172 and #2174 landed). If the conflict is only in `.gitignore` that will be my #2171/#2173 secret-ignore work, and mine is purely additive so take both sides.

Flagging one thing to watch during the rebase, because it bit #2068 badly tonight: when you resolve, check you are not carrying back anything dev has since changed. A branch cut before a hardening commit can silently revert it during a rebase and the merge will report clean, with no conflict marker to warn you. Worth a `git log <merge-base>..origin/dev` over the files you touch before you push.

jaylfc:
Reviewed properly rather than on the bot verdicts, because two of three say "No Issues Found, Merge" and I think they missed something in `store_signing.py`.

**The core change is right and I want it in.** Converting the TOCTOU re-verify from fail-open to fail-closed is exactly correct: previously an unreadable manifest (`on_disk = None`) or a missing signature (`stored_sig is None`) silently skipped re-verification and the install proceeded. Now each of those returns `False` and the install is blocked with 403. Moving it to `asyncio.to_thread` is also a good call, and I confirmed `asyncio` is imported at `store_install.py:12`, so no NameError.

**But this PR has no tests, and it is a security fix.** Both changed files are production code:

```
tinyagentos/routes/store_install.py
tinyagentos/store_signing.py
```

A fail-closed gate that has only ever been observed passing is unproven exactly where it counts. I want a test per refusal branch that **proves it goes red**: manifest missing, manifest unreadable, `safe_load` returns empty, `stored_sig is None`, and signature mismatch. Five cheap tests, each asserting 403. Without them the next refactor can quietly restore fail-open and CI will stay green, which is precisely how this bug got in.

**The finding the bots missed: `default=str` in `_canonical_manifest_bytes` creates a canonicalisation collision.**

`store_signing.py:238` now passes `default=str` to the function that produces the **bytes that get signed**. I measured what that does:

```
plain manifest identical:  True          <- good, no compat break
date without default=str:  TypeError: Object of type date is not JSON serializable
date WITH default=str:     {"when": "2026-01-01"}

COLLISION CHECK
  {"when": "2026-01-01"}          -> {"when": "2026-01-01"}
  {"when": date(2026,1,1)}        -> {"when": "2026-01-01"}
  IDENTICAL: True
```

The good news first: for any manifest that already serialised, the output is **byte-identical**, so existing signatures are unaffected. `default` is only consulted for otherwise-unserialisable objects. No compatibility break.

The problem is that **a canonical form must be injective**, and this one no longer is. `yaml.safe_load` parses unquoted `2026-01-01` into a `datetime.date` but quoted `"2026-01-01"` into a `str`. Those are two different manifests that now produce the **same signing bytes**, so a signature valid for one is valid for the other. Type erasure in a signing primitive is the thing canonicalisation exists to prevent, and it is a strange thing to introduce in a PR whose entire point is tightening a signature gate.

Worth noting what it is really fixing: without `default=str`, a date-bearing manifest raises TypeError inside `_verify_sig`, which is called **outside** the `try` in `_toctou_reverify`, so it propagates through `asyncio.to_thread` and surfaces as a 500 rather than a clean 403. That is a genuine bug and worth fixing. But the fix should preserve injectivity.

**What I would rather see, in preference order:**
1. Reject non-primitive types explicitly in `_canonical_manifest_bytes` and return a verification failure (fail-closed, consistent with this PR's thesis), instead of coercing them.
2. Or, if dates must be supported, type-tag them so a date and its string form cannot collide.
3. Or, at minimum, move `_verify_sig` inside the `try` so a TypeError becomes `return False` (403) rather than a 500, and drop `default=str` entirely.

Option 3 is the smallest change and fits this PR best: it makes the unserialisable case fail **closed**, which is what the PR is for.

Happy to merge once there are refusal-path tests and the canonicalisation question is resolved. The `store_install.py` half is good work and I do not want it stuck behind this.

jaylfc:
Consolidating what is open on your side so you are not hunting across seven PRs at 1am. You have been pushing steadily (last commit 00:44) and I would rather hand you an ordered queue than a pile.

**Ordered by what unblocks the most:**

1. **#1908** — `doc-gate` red, and it gates **#1945**. I reproduced it locally: `DOC-GATE FAIL: routes -- an API route module was added or removed (edit one of: docs/agent-coordination.md, or add a 'Docs-Reviewed: <why>' trailer)`. One doc note or one trailer. Cheapest unblock in the queue, and two PRs move.
2. **#2165** — rebase (DIRTY since #2174 landed) **plus the real bug I just posted**:  spreads  into a literal, and  returns a  instance, so the CSRF token this PR adds is silently stripped. I measured it. Qodo caught it and I had wrongly dismissed it.
3. **#2050 (this PR)** — refusal-path tests, plus the  canonicalisation question.  and  now produce identical signing bytes.
4. **#2048** — the one that matters most: make  actually skip PIN verification, plus ONE test that mints **and successfully redeems**. That single test kills A1 and A2 together and is the only thing that will show the feature runs.
5. **#2070** — rebase now that P2 landed as #2177.
6. **#2035** — the  DB orphan on existing installs, and the real- test.
7. **#2034, #2043** — untouched since my reviews; findings stand.

**Two things worth knowing that are not asks:**

**#2177 merged** and it was a genuinely good rebase. I verified the silent revert was gone by symbol-diffing against , and your streaming cap fix is correct: client open during the read, content-type gated before the body, cap checked inside . The size-cap test that shrinks  to 100 and yields 200 bytes is exactly the right way to test a limit.

**#2068 closed** as superseded, so nobody builds on the wrong branch.

**On #2180:** do not add it to required status checks yet. Let it run advisory for a few days first. Against the real #2068 head it produced **85 findings**, all technically true but unreadable, and a gate that trains people to waive is worse than none. On a stale branch the answer is always "rebase", not "waive 85 symbols".

jaylfc:
**Ignore my previous comment, it posted mangled.** I wrote it inside a double-quoted shell string and every backticked identifier got eaten by command substitution, so items 2, 3, 4 and 6 lost the exact technical detail that made them useful. My error and a good argument for heredocs. Corrected queue below.

**Ordered by what unblocks the most:**

1. **#1908** — `doc-gate` red, and it gates **#1945**. Reproduced locally: `DOC-GATE FAIL: routes -- an API route module was added or removed (edit one of: docs/agent-coordination.md, or add a 'Docs-Reviewed: <why>' trailer)`. One doc note or one trailer. Cheapest unblock here, and two PRs move.

2. **#2165** — rebase (DIRTY since #2174 landed) **plus a real bug**. `desktop/src/lib/x-monitor.ts:54`:
   ```ts
   const res = await fetch(url, { ...init, headers: { Accept: "application/json", ...init?.headers } });
   ```
   `withCsrf` returns `headers` as a **Headers instance**, and spreading a Headers into an object literal yields `{}`. Measured: `{"Accept":"application/json"}`, token gone. So the CSRF header this PR exists to add is stripped at that line, and every `postJson`/`putJson` in the file inherits it. Fix by normalising: build a `new Headers({Accept:...})`, then `new Headers(init?.headers).forEach((v,k) => headers.set(k,v))`. Qodo flagged this and I wrongly dismissed it after checking `csrf.ts`, which is correct; the bug is at the call site.

3. **#2050** — refusal-path tests, plus the `default=str` question in `_canonical_manifest_bytes`. Measured: `{"when": "2026-01-01"}` (string) and `{"when": date(2026,1,1)}` produce **identical** signing bytes, so a signature valid for one is valid for the other. Existing signatures are unaffected, so no compat break, but a canonical form must be injective.

4. **#2048** — the one that matters most: make `pin_required=False` actually skip PIN verification in `redeem()` (right now it writes a column nothing reads), plus ONE test that mints **and successfully redeems**. That single test kills A1 and A2 together and is the only thing that will show the feature runs at all.

5. **#2070** — rebase now that P2 landed as #2177.

6. **#2035** — the `notes_set_done` DB orphan on existing installs (seeding is `INSERT OR IGNORE` every startup, and `list_tools` does not cross-check `SKILL_IMPLEMENTATIONS`), plus the registry-miss test using a real `AgentRegistryStore` rather than `MagicMock`.

7. **#2034, #2043** — untouched since my reviews; findings stand.

**Two things that are not asks:**

**#2177 merged**, and it was a genuinely good rebase. I verified the silent revert was gone by symbol-diffing against `origin/dev` rather than reading the diff, and the streaming fix is correct: client open during the read, content-type gated before the body is consumed, cap checked inside `aiter_bytes` against a running total. The size-cap test that shrinks `_MAX_WEB_BYTES` to 100 and yields 200 bytes is exactly the right way to test a limit.

**#2068 closed** as superseded so nobody builds on the wrong branch.

**On #2180:** do not add it to required status checks yet. Let it run advisory for a few days. Against the real #2068 head it produced **85 findings**, all technically true but unreadable, and a gate that trains people to waive is worse than none. On a stale branch the answer is always "rebase", never "waive 85 symbols individually".

jaylfc:
**Both findings are fixed properly. Merging.** Verified each against the code rather than the commit message.

**1. `default=str` is gone** from `_canonical_manifest_bytes`, so the canonical form is injective again. A string and a date no longer collide to identical signing bytes.

**2. The fail-closed structure is now correct**, and you took the option I hoped you would. `_verify_sig` is *inside* the `try`, so an unserialisable manifest returns `False` and the install is blocked with 403, rather than raising a TypeError through `asyncio.to_thread` and surfacing as a 500. The whole re-verify path now fails closed on missing, unreadable, empty, unsigned and mismatched.

**3. The tests are the right ones**, and the comment in them says exactly what I wanted to see: *"each test asserts 403 to prove the gate goes red where it counts"*. `test_toctou_manifest_missing_returns_403` and `test_toctou_manifest_unreadable_returns_403` pin the refusal branches, which is what makes the fix survive the next refactor.

**One thing worth recording, and it is not a criticism of this PR.** Every bot comment here is dated **2026-07-19**. Your last commit is **2026-07-28T03:14**. So the "No Issues Found / Recommendation: Merge" verdicts, and the "Address before merge" one, all reviewed a version of this PR that is **nine days old and no longer exists**. No bot has looked at the current head.

I am merging on **my own review**, not on those verdicts, and stating that explicitly so the record is honest. A stale bot approval is worse than no approval, because it reads as a second opinion when it is an opinion about different code. Worth everyone checking comment dates against the head SHA before treating a bot tick as review.

Good fix. This closes a live install-gate bypass on `dev`.

Issue #2065, opened 2026-07-20
BUG: invite dialog closes on successful mint without showing the URL + PIN bundle


jaylfc:
Second live-reported invite bug from Jay (companion to #2064). Repro on the Pi (beta.42): open Invite external agent, select scopes, press Mint invite. Server logs show POST /api/projects/{id}/invites returned 200 OK and the invite was created, but the dialog simply closed - no URL, no PIN, nothing. The one-time PIN is unrecoverable afterward, so the created invite is orphaned (it still counts toward the 6-pending cap).

The success path EXISTS in the component (InviteAgentDialog.tsx:168 builds the instruction string from result.invite_id + result.pin), so the mint handler is either throwing after the fetch resolves (response-shape mismatch swallowed by a catch that closes the dialog - pitfall 8) or a state update races the close. Note the mint response shape is flat {invite_id, pin, expires_ts, scopes, approval_mode, check_interval_secs} - verify the handler is not expecting a nested record (the #2045 store changes altered the record dict; check whether the route response changed with it).

Fix: (1) find and remove the silent failure - on any post-mint render error the dialog must show the error, never close; (2) show the success view with URL + PIN and a copy button; (3) since the PIN is shown once, add a test asserting the success view renders from the FLAT response shape; (4) consider surfacing newly-minted-but-unviewed invites in Pending invites with a cancel action so orphans are recoverable.

Live evidence: Pi journal 17:01:18 POST 200 followed by GET 200, dialog closed with no bundle shown. Jay was unblocked via a direct API mint.

jaylfc:
Root cause found, and it explains all three symptoms: the invite list endpoints return `scopes` as a raw JSON-encoded string (the DB TEXT column passed straight through), while the dialog types it as `string[]` and calls `inv.scopes.join(", ")` at InviteAgentDialog.tsx:349.

- Dialog crashes the whole SPA on open whenever any pending invite exists (error boundary, reload required)
- After a successful mint the pending-list refresh hits the same throw and unmounts the dialog before the URL/PIN success view renders - that is the silent close reported here
- The dialog only survived while the pending list was empty, which is why this stayed hidden

Backend contract fix: #2066 (parses scopes to a list in both serializers, defensive on malformed rows).

Remaining for the dialog rework in this issue:
1. Show the mint success view reliably (URL + PIN, copy buttons) - do not auto-close
2. Harden the pending list rendering (never let a render throw escape to the app error boundary; scope the dialog with its own boundary)
[...]

PR #2071, opened 2026-07-20
fix(invites): allow revoking expired invites, 409 for terminal states

jaylfc:
## Problem
Live repro from Jay's e2e test: both pending invites hit their 15-minute TTL and lazily flipped to `expired` on read. `store.revoke` only matches `status='pending'`, so the DELETE matched zero rows and the route returned 404 "invite not found" - while the dialog kept listing the rows (it labels everything "Pending" regardless of status, tracked in #2065). Net effect: dead invites, displayed as pending, unrevokable.

## Fix
- `invite_store.revoke`: matches `pending` and `expired` (an expired invite is exactly what a user wants to clean up).
- `revoke_invite` route: `redeemed`/`revoked` now return 409 with the actual state instead of a misleading 404; 404 remains only for genuinely missing or wrong-project invites.

## Testing
- New: revoke-expired returns 204 (store + route), revoke-redeemed returns false, double-revoke returns 409.
- Full invite store + route suites: 71 passed.

## Related
- #2065 dialog rework should render status badges (pending vs expired) rather than one "Pending" list - noted there.
- The 15-minute TTL itself is a separate product decision (human-mediated redeems need longer or a mint-time choice) - raising with Jay.

## Summary by CodeRabbit

* **New Features**
  * Expired project invitations can now be revoked successfully.
* **Bug Fixes**
  * Revocation now returns clearer **409 conflict** responses when an invitation is already **redeemed** or **revoked**.
  * If an invitation is mid-redeem (“claimed”), revocation is blocked with a **409** including a **“mid-redeem”** message.
  * If the invitation changes state during the request, the endpoint returns **409** with a retry guidance response.
* **Tests**
  * Added coverage for revoke behavior across expired, redeemed, claimed, and already-revoked scenarios.

### DIFF
--- a/tests/projects/test_invite_store.py
+++ b/tests/projects/test_invite_store.py
@@ -315,6 +315,44 @@ async def test_revoke_nonexistent_returns_false(store):
     assert await store.revoke("000000") is False
 
 
+@pytest.mark.asyncio
+async def test_revoke_expired_invite(store):
+    """An expired invite can still be revoked (cleanup from the pending list)."""
+    result = await store.mint(
+        project_id="prj-1",
+        scopes=[],
+        approval_mode="auto",
+        check_interval_secs=1800,
+        created_by="u",
+    )
+    iid = result["record"]["invite_id"]
+    await store._db.execute(
+        "UPDATE project_invites SET expires_ts = 1 WHERE invite_id = ?", (iid,)
+    )
+    await store._db.commit()
+    row = await store.get(iid)
+    assert row["status"] == "expired"
+    ok = await store.revoke(iid)
+    assert ok is True
+    row = await store.get(iid)
+    assert row["status"] == "revoked"
+
+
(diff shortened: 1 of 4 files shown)


jaylfc:
Both Kilo suggestions folded at the new head:

1. **Route-level 409 for redeemed untested** - added `test_revoke_redeemed_returns_409` (mint, redeem via store, DELETE, assert 409). You were right that the PR description claimed coverage the store-level test did not provide.
2. **Generic 409 for a transient `claimed` invite** - `claimed` now has its own branch returning "invite is mid-redeem, retry once it settles" rather than falling through to "invite state changed, retry", with `test_revoke_claimed_returns_409_with_mid_redeem_message` covering it. Behaviour was already safe (revoke correctly refuses mid-redeem); this just stops the message from misdescribing why.

Suite: 73 passed.

PR #2076, opened 2026-07-21
tsk-midukh  [OPEN]  Add vitest coverage for the Crosswords app

jaylfc:
Autonomous build of board card tsk-midukh.

Files:
 desktop/package-lock.json               | 34 ++-------------------------------
 desktop/src/apps/CrosswordsApp.test.tsx | 24 ++++++++++++++++++++++-
 2 files changed, 25 insertions(+), 33 deletions(-)

----
## Summary by Gitar

- **Test coverage:**
    - Added test case for checking answers in `CrosswordsApp` using `checkButton` interaction.
    - Implemented test for switching puzzles by triggering the `new puzzle` button.

### DIFF
--- a/desktop/package-lock.json
+++ b/desktop/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "tinyagentos-desktop",
-  "version": "1.0.0-beta.41",
+  "version": "1.0.0-beta.42",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "tinyagentos-desktop",
-      "version": "1.0.0-beta.41",
+      "version": "1.0.0-beta.42",
       "dependencies": {
         "@codemirror/lang-markdown": "^6.5.1",
         "@codemirror/language-data": "^6.5.2",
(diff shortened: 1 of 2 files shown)


jaylfc:
Closing and returning the card to the board. The test file work is real, not a stub, but it is not mergeable as written:

1. **The new test is tautological.** `clicks check answers button to enter check mode` clicks the button and then asserts `expect(checkButton).toBeInTheDocument()` - the button obviously still exists after being clicked, so the assertion cannot fail and proves nothing about check mode being entered. This is the same class as #1979 (always-true assert replaced with a real check). A useful version would assert the observable consequence: that incorrect cells become marked, or that the check-mode styling or state actually applies.
2. **Gratuitous unrelated edit**: `a clue's` changed to `a clue\'s` inside a double-quoted string. The escape is unnecessary and the line had nothing to do with the card.
3. **Unrelated `desktop/package-lock.json` diff.** That one is my fault, not the lane's: the lockfile was generated on macOS and the Linux build host rewrites platform-specific entries (libc, os/cpu on optional deps) whenever npm runs, so any lane that merely executes the test suite picks up lockfile churn. I have fixed the executor to strip it, matching the existing uv.lock guard, so it will not ride along on future PRs.

Card goes back to the board rather than being marked blocked - the task is sound, the attempt was not.

PR #2077, opened 2026-07-21
tsk-43msix  [OPEN]  taOStalk s1: snippet ps1 twin + Playwright e2e + a

jaylfc:
Autonomous build of board card tsk-43msix.

Files:
 desktop/src/apps/MessagesApp.tsx        |  20 +-
 desktop/src/apps/chat/ConnectWizard.tsx | 393 ++++++++++++++++++++++++++++++++
 desktop/src/apps/chat/MessageList.tsx   |  23 +-
 desktop/src/apps/chat/blocks.tsx        | 166 ++++++++++++++
 4 files changed, 598 insertions(+), 4 deletions(-)

----
## Summary by Gitar

- **New components:**
    - Added `ConnectWizard` to guide users through agent channel creation and display connection snippets.
    - Created `blocks.tsx` to handle structured message rendering including `TextBlock`, `ThinkingBlock`, `ToolCallBlock`, and `StatusBlock`.
- **Message rendering updates:**
    - Updated `MessageList.tsx` to map `content_blocks` to specialized UI components.
    - Added `session_turn` to `content_type` and extended `Message` interface in `MessagesApp.tsx` to support structured content blocks.

<sub>This will update automatically on new commits.</sub>

## Summary by CodeRabbit

* **New Features**
  * Added a guided three-step wizard for connecting chat sessions to running agents.
  * Users can create a chat channel, verify the connection, and copy Bash or PowerShell setup commands.
  * Chat messages now support richer content blocks, including text, thinking sections, tool activity, statuses, and questions.
  * Added expandable details, previews, status indicators, and selectable options for supported message content.

* **Bug Fixes**
  * Added a fallback display for unsupported message content types.

PR #2079, opened 2026-07-21
docs(skill+pitfalls): three defect classes from the Library P1 review cycle

jaylfc:
Distils this cycle into the contributor docs. Every item is drawn from a real defect, not a hypothetical.

**Pitfall 20: a tolerance is not a test of an invariant.** `assert abs(after - before) <= 2` ran green in #2062 while reprocess destroyed the user's original uploaded file (observed `before=2, after=0`). The same tolerance would have accepted a doubling. Assert equality and assert the resulting status.

**Pitfall 21: shell snippets inside template literals must escape `${`.** In #2077 a bash snippet in a JS template literal produced **225 TypeScript syntax errors from one cause**, in one of four otherwise-clean files, which read like incoherent output rather than a single mechanical mistake. Generalises to any language sharing `${...}` with the shell, and is invisible without compiling.

**Pitfall 22: conflict resolution is a decision, not a mechanical act.** Library P2/P3 both went CONFLICTING when P1 landed, and all four defects P1 fixed are reintroducible by a plausible-looking resolution.

**Skill** gains a short verify-before-you-claim section covering those three plus the mock-from-the-contract rule: a test whose fixture is built from the caller's assumption rather than the service's documented response certifies the bug. That happened three times in one PR cycle (invented `dbPath` shape, un-nested poll response, tolerance assertion).

Docs only.

## Summary by CodeRabbit

* **Documentation**
  * Added contributor guidance for verifying exact test outcomes, using accurate external-service response fixtures, and running type checks and builds before submitting changes.
  * Documented best practices for escaping template-literal shell snippets and resolving merge conflicts carefully.
  * Added new entries covering recurring review pitfalls.

### DIFF
--- a/.claude/skills/taos-development-skill/SKILL.md
+++ b/.claude/skills/taos-development-skill/SKILL.md
@@ -274,6 +274,31 @@ A close without a successor link reads as lost work and forces the maintainer
 into git forensics (this happened with #1927/#1924 - both were legitimate
 "landed via" closures that looked like data loss for hours).
 
+### Asking another team's agent a question
+
+taOS depends on sibling services with their own maintainer agents (taOSmd, the
+website). Their contracts are theirs to state, so ASK rather than inferring from
+their source (pitfall 23). How to reach them depends on where you sit:
+
+- **On the A2A bus** (internal agents): post on the relevant channel, name the
+  agent, and expect a reply inside the hour.
+- **Outside the bus** (external contributors): open an issue on
+  `jaylfc/taos-agent-commons`, the private invite-only coordination repo. Label
+  it `contract-question` and name the service. @taOS-dev sweeps it hourly and
+  relays to the owning agent on the bus, then carries the answer back.
+- `jaylfc/taosmd` is public with issues enabled, so a taosmd contract question
+  can also go straight there.
+- `jaylfc/taos-website` is private, so commons or the relay is the only route.
+
+If a question sits unanswered for more than about two hours, escalate by also
+raising it on the PR. The relay is a person-shaped hop and can stall; silence
+should never be mistaken for progress.
+
+This arrangement is TEMPORARY scaffolding. It retires when an external
+contributor can hold a taOS identity and reach the bus directly, which is the
+same capability as agent sharing. Do not build tooling that assumes it is
+permanent.
+
 ### Verify before you claim, and compile before you PR
 
 - An assertion with a tolerance (`abs(a - b) <= n`) does not test an invariant.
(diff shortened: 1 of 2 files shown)


jaylfc:
Strengthened the mock guidance after a challenge from Jay, who asked whether we should avoid mocks entirely or at least force a follow-up issue.

I checked before answering: **372 of 535 Python test files use mocking**, so a blanket ban is neither actionable nor right. Most of that is mocking our own internals or injecting errors nobody can produce on demand (a 500, a timeout, an ImportError), and none of it is what bit us.

The dangerous class is narrow: **mocks of external service contracts we do not control**, whose fixtures were hand-composed from what the calling code expects. That fixture encodes a belief about someone else's API, and a wrong belief produces a green test that certifies the bug. All three failures in the #2062 cycle were that exact shape.

Added as **pitfall 23** with an enforceable hierarchy rather than a principle:
1. Capture a real response and commit it as the fixture. A recorded response cannot encode a wrong belief.
2. Keep one real integration test per external contract, feature detecting so it skips when the service is unreachable. taosmd's `GET /version` capabilities list exists for exactly this.
3. If neither is possible, mark the mock provisional **in code** with its source and verification date.

On the follow-up-issue suggestion specifically: I argued against it as the primary mitigation and said so in the text. With 260 open issues, filing one detaches the caveat from the code and functions as indefinite deferral. A comment in the file travels with the code and shows up in review; an issue does not.

The reviewer-facing version is one question: **where did this fixture come from?** That is a question I could have asked on #2062 and did not.

jaylfc:
Revised again after Jay pointed out the obvious thing I had missed: **for taOSmd there is no danger, because that is what the A2A bus is for.**

He is right, and it reframes the rule. taOSmd is not a third-party API, it is a sibling service whose maintainer is one message away. Every mock failure in the #2062 cycle was a guess at something that was free to obtain:

- The builder invented a `dbPath` request shape rather than asking what the endpoint takes.
- I verified the envelope against taosmd's **source** rather than asking its owner, which found the right answer but only after the wrong one had already shipped twice.
- When we finally did ask, we got the response envelopes documented, a partially wrong stats key list corrected by its own author, and a `/version` capabilities endpoint built specifically so the contract becomes machine-checkable.

So pitfall 23 now splits by whether the contract has a reachable owner:

**Sibling services: ask, do not guess.** Reverse-engineering a sibling's API from its source is a smell rather than diligence, because source tells you what it does today while the owner tells you what it guarantees. Contributors without bus access ask in the PR and the lead relays, which is what happened here.

**Genuine third parties (GitHub, OpenRouter, Reddit): capture, do not compose.** There is nobody to ask, so record a real response instead of writing a fixture from what the caller expects.

Both then keep one feature-detecting integration test, or mark the mock provisional in code.

Worth stating plainly since it is my own miss: I spent a lot of this cycle verifying contracts against source when a bus message would have been faster and more authoritative. The rule now says so.

PR #2107, opened 2026-07-22
fix(notifications): show approve/deny for scope-request notifications in bell

hognek:
Fixes scope-request notification UI gap from #1921.

## Problem
PR #1921 (agent scope-requests) emits notifications with `source: "agent_scope_requests"`, but `NotificationCentre` and `NotificationToast` only render `ConsentActions` (approve/deny buttons) for `source === "auth_requests"`. Scope-request approve/deny was invisible in the UI — approving required hand-rolled curl.

## Fix
- Branch the source check in `NotificationCentre.tsx` and `NotificationToast.tsx` to also match `"agent_scope_requests"`
- `ConsentActions` accepts optional `source` (defaults to `"auth_requests"`) and `canonicalId` props; routes approve/deny to the scope-request endpoints (`/api/agents/registry/{id}/scope-requests/{req_id}/approve|deny`) when source is `"agent_scope_requests"`
- `consentPayload()` extracts `canonical_id` from notification data for scope-request endpoint path construction
- Backward compatible: existing callers (DecisionsApp, existing tests) continue to work with `source="auth_requests"` default

## Files
- `desktop/src/components/ConsentActions.tsx` — added source/canonicalId props + URL branching + consentPayload canonicalId extraction
- `desktop/src/components/NotificationCentre.tsx` — branched source check + passes source/canonicalId
- `desktop/src/components/NotificationToast.tsx` — same

## Tests
- `npm run build` passes (tsc + vite build)
- `vitest run ConsentActions.test.tsx` — 7/7 pass (backward compatibility confirmed)

Ref: #1921 — PR adding scope-request flow

## Summary by CodeRabbit

* **New Features**
  * Added support for approving or denying agent scope requests from notifications and toast messages.
  * Consent actions now handle additional request metadata to route decisions correctly.
  * Expanded consent information to include canonical request identifiers.

### DIFF
--- a/desktop/src/components/ConsentActions.tsx
+++ b/desktop/src/components/ConsentActions.tsx
@@ -38,11 +38,15 @@ export function ConsentActions({
   scopes,
   requestedProjectId,
   onResolved,
+  source = "auth_requests",
+  canonicalId,
 }: {
   requestId: string;
   scopes: string[];
   requestedProjectId?: string;
   onResolved?: () => void;
+  source?: string;
+  canonicalId?: string;
 }) {
   const needsProject = scopes.some((s) => PROJECT_SCOPES.has(s));
   const [busy, setBusy] = useState(false);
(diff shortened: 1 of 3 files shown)


jaylfc:
Verified CodeRabbit's Major finding against the code. **It is already addressed**, so merging.

The concern was that `canonicalId` could fall back to `requestId` and produce `/registry/{requestId}/scope-requests/{requestId}`, or that an empty value yields `/registry//...`.

Neither happens on this head:

- `consentPayload` does not fall back. It returns `canonicalId` only when `typeof data.canonical_id === "string"`, otherwise `undefined`.
- `ConsentActions` guards before building the URL:
  ```js
  if (source === "agent_scope_requests" && !canonicalId) {
    setError("Missing agent identifier; cannot process this scope request.");
    return;
  }
  ```
  `!canonicalId` is falsy for both `undefined` and `""`, so the empty-string case is covered too.
- The registry URL is constructed in exactly one place (`ConsentActions.tsx:146`), three lines after that guard. Every caller (`DecisionsApp`, `NotificationCentre`, `NotificationToast`) routes through `ConsentActions`, so there is no bypassing path.

One fair point remains from the finding, as a UX nit rather than a defect: CodeRabbit suggested *disabling* the action when the metadata is missing, and this errors on click instead. Disabling up front would be better, but a user cannot reach a malformed request either way. Not worth holding the PR.

Issue #2135, opened 2026-07-26
CI test suite runs ~2.5x its documented time (16 min claimed, 31-41 min actual)


jaylfc:
Split out of #2134, which raised `timeout-minutes` from 45 to 75 so the cap stopped manufacturing false red PRs. That was a containment fix. This is the actual problem.

## The numbers

The workflow comment said 3.12/3.13 PR runs finish in ~16 min. Measured 2026-07-26 across the last 12 job records:

| job | duration |
|---|---|
| test (3.13) | 31 - 35 min |
| test (3.12) | 38 - 41 min (two hit the 45 cap and were killed) |

Roughly 2.5x the documented expectation. Nobody noticed because the failure mode was gradual and the only symptom was occasional reds that looked like flakes.

## Why it matters beyond wall-clock

- Every PR now costs 40 min of feedback latency, so a fold-and-repush cycle is over an hour.
- 3.12 is reliably ~7 min slower than 3.13 on identical work. Worth understanding rather than assuming, since it may point at a specific dependency path rather than general growth.
- The daily cron adds 3.11, which the old comment put at ~30 min. If 3.11 has drifted proportionally it may now be the slowest leg of all.

## Worth checking

- `--durations` output to find whether this is broad growth or a handful of slow tests dominating. The recent additions around agent scope requests and project files are all end-to-end through real routes and middleware, which is the expensive shape.
- Whether anything sleeps or polls with a real clock rather than a fake one.
- Whether the suite can run parallel (`pytest-xdist`), and if not, what specifically prevents it (shared sqlite files are the usual culprit).
- Whether install/provision steps are a meaningful share, or it really is all test execution.

## Done when

Either the runtime is back near its documented figure, or the documented figure is corrected to a measured one and the cap in #2134 is re-tuned to match. What must not persist is a comment asserting a number nobody has verified since it was written, because that is what let the cap silently become a tripwire.

jaylfc:
Additional data from a #2133 run, which narrows this considerably.

**The suite is already parallel.** The step is `pytest tests/ --tb=short --ignore=tests/e2e -n auto`, so xdist is not the missing piece.

**Test count has more than doubled.** The workflow comment next to that step says:

> The serial suite (4845 tests) took ~22 min; -n auto cuts it to a few minutes

A run today reported:

```
==== 2 failed, 10246 passed, 17 skipped, 64 warnings in 2038.94s (0:33:58) ====
```

[...]

PR #2137, opened 2026-07-26
ci: shard the test suite across parallel jobs (#2135)

jaylfc:
Closes the runtime half of #2135.

## The cause is cores, not slow tests

`xdist` parallelises *within* a job. A hosted runner is 2 vCPU, so `-n auto` gives 2 workers however many tests exist. The suite grew from 4,845 to 10,250, and 10,250 tests over 2 workers is ~34 min. That matches the observed 31 to 41 min almost exactly.

There is no hotspot to optimise instead. The largest single test file is 124 tests, so the growth is spread broadly across the suite.

## The fix is free

A public repo gets roughly 20 concurrent jobs and we were using 2. Sharding into 4 gives 2 versions x 4 shards = 8 jobs, still well inside the limit, and should cut wall-clock to about a quarter.

## Verified the partition, did not assume it

The dangerous failure here is silent: a sharding scheme that drops tests leaves CI **green while testing less**. So this was checked explicitly:

```
all tests:       10250
union of shards: 10250
unique in union: 10250   (no duplicates)
missing from every shard: 0
```

## Also

`Verify app starts` is gated to shard 1. It is an import smoke check and was otherwise running identically in all 8 jobs.

## Not addressed here

This makes the suite *finish* faster; it does not make it *cheaper*. Total compute is roughly unchanged. If the growth itself is a concern, that is worth a separate look, but it is no longer blocking merges.

## Summary by CodeRabbit

* **Tests**
  * Improved automated test execution by splitting the test suite across four parallel jobs.
  * Retained application startup verification while avoiding redundant checks.
* **Chores**
  * Added the tooling required to distribute tests consistently across parallel runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,15 @@ jobs:
   # run at once instead of queueing behind the 2 self-hosted boxes. The VPS +
   # Fedora runners stay free for the kilo lane and GPU/bench work.
 
-  test:
+  # Renamed from "test" to "shards" because sharding changes the check-run names
+  # this job reports. Branch protection matches required checks by NAME, so the
+  # sharded jobs report "test (3.12, 1)" etc and the required contexts
+  # "test (3.12)" / "test (3.13)" would never be reported again. Protection then
+  # waits forever on checks that cannot exist and blocks every PR, including the
+  # one that would fix it. The aggregate "test" job below keeps reporting those
+  # two exact names, so no protection change is needed and there is no window
+  # where dev is wedged.
+  shards:
     runs-on: ubuntu-latest
     # Measured 2026-07-26 over the last 12 job records: 3.13 runs 31-35 min and
     # 3.12 runs 38-41 min. The previous comment claimed "~16 min", which had
@@ -61,7 +69,8 @@ jobs:
         # Cores per job were the bottleneck, not the tests.
         #
         # A public repo gets ~20 concurrent jobs free and we were using 2, so
-        # this costs nothing. 2 versions x 4 shards = 8 jobs.
+        # this costs nothing. 2 versions x 4 shards = 8 jobs on PR/push, and
+        # 12 on the scheduled run that adds 3.11.
         shard: [1, 2, 3, 4]
 
     steps:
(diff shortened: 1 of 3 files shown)


jaylfc:
**Not merging: this would break branch protection on `dev` the moment it lands.** Required checks are name-matched, and this PR renames the jobs.

`dev` requires exactly:
```
test (3.12)
test (3.13)
spa-build
lint
```

This PR produces:
```
test (3.12, 1)  test (3.12, 2)  test (3.12, 3)  test (3.12, 4)
test (3.13, 1)  test (3.13, 2)  test (3.13, 3)  test (3.13, 4)
```

`test (3.12)` and `test (3.13)` would then **never be reported again**. GitHub does not treat the shards as satisfying the old contexts, so every subsequent PR would sit forever waiting on two checks that no longer exist - including the PR that tries to fix it. Self-inflicted deadlock at 1am is a bad way to find out.

The sharding work itself looks right and I verified earlier that the four shards form an exact partition of the suite. This is purely a rollout-ordering problem:

1. Land the sharded workflow.
2. Update branch protection contexts to the shard names (or add a single aggregating `test` job that depends on all shards and require only that - which is the more durable shape, because the required-context list stops changing every time the shard count does).
3. Then remove the old contexts.

**My recommendation: add the aggregate gate job in this PR** so the required context becomes one stable name, and protection is updated once rather than every time we re-tune shard count.

Two minor bot findings also stand, both real and both about comments rather than behaviour: the shard-count comment says something inconsistent with the matrix (Qodo) and the smoke-check count is wrong (CodeRabbit). Worth folding while you are in there.

Not a criticism of the work - it is a genuinely good speedup. It just cannot land without the protection change, and I am not changing branch protection unattended overnight.

jaylfc:
Merging. Proven on both paths rather than just the happy one.

**Success path:** the aggregate jobs report `test (3.12)` and `test (3.13)` exactly, so branch protection keeps seeing the contexts it requires and needs no change. That was the whole reason for the aggregate rather than renaming the required checks.

**Failure path:** verified on a throwaway branch (#2162) carrying one deliberately failing test. First run exposed a real defect I would otherwise have shipped: the aggregate is a matrix and fail-fast defaults to true, so the first gate to fail CANCELLED its sibling, giving `test (3.12) failure, test (3.13) cancelled`. Never unsafe, since cancelled is not success, but which version was reported broken depended on which gate lost the race. Fixed with `fail-fast: false` on the aggregate only; the shards keep fail-fast, where stopping early is the point. Re-verified: both gates now report failure independently.

Worth stating plainly because it nearly went the other way: the success path looked perfect and I would have merged on it. The defect only existed in the half that actually protects the branch.

Also folded the two review findings, both correct: the job-count comment ignored the scheduled run's third Python version, and the smoke check runs once per version rather than once overall.

Note the `Kilo Code Review` check is red with "Review failed: Assistant request failed". That is the bot erroring, not a finding, and it means this PR is UNREVIEWED by it rather than approved. Reviewed by me instead.

PR #2145, opened 2026-07-26
tsk-dk5v6d  [OPEN]  PWA install: exempt manifest + icons from auth mid

jaylfc:
Autonomous build of board card tsk-dk5v6d.

Files:
 tinyagentos/auth_middleware.py | 6 +++++-
 1 file changed, 5 insertions(+), 1 deletion(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed PWA installation by allowing required desktop manifest and icon assets to load before authentication.
  * Improved access to installation resources for browsers that fetch them during the install process.

PR #2146, opened 2026-07-26
tsk-gp5xrq  [OPEN]  S2A: GET /api/share/destinations (authorization-fi

jaylfc:
Autonomous build of board card tsk-gp5xrq.

Files:
 tests/test_routes_share_destinations.py | 117 ++++++++++++++++++++++++++++++++
 tinyagentos/auth_middleware.py          |   2 +-
 tinyagentos/routes/share.py             |  63 +++++++++++++++++
 4 files changed, 184 insertions(+), 1 deletion(-)

### DIFF
--- a/tests/test_routes_share_destinations.py
+++ b/tests/test_routes_share_destinations.py
@@ -1,0 +1,117 @@
+import pytest
+from httpx import ASGITransport, AsyncClient
+
+from tinyagentos.app import create_app
+from tinyagentos.device_store import DeviceStore
+from tinyagentos.agent_registry_store import AgentRegistryStore
+
+
+@pytest.fixture
+def app(tmp_data_dir):
+    return create_app(data_dir=tmp_data_dir)
+
+
+@pytest.mark.asyncio
+class TestShareDestinations:
+    async def test_unauthenticated_returns_401(self, app, tmp_data_dir):
+        transport = ASGITransport(app=app)
+        async with AsyncClient(transport=transport, base_url="http://test") as client:
+            resp = await client.get("/api/share/destinations")
+            assert resp.status_code == 401
+
+    async def test_library_always_present_for_paired_device(self, client, app):
+        reg = await client.post("/api/devices/register", json={"platform": "ios"})
+        assert reg.status_code == 200
+        token = reg.json()["scoped_token"]
+
+        resp = await client.get(
+            "/api/share/destinations",
+            headers={"Authorization": f"Bearer {token}"},
+        )
+        assert resp.status_code == 200
+        body = resp.json()
+        assert "destinations" in body
+        kinds = {d["kind"] for d in body["destinations"]}
+        assert "library" in kinds
+        lib = next(d for d in body["destinations"] if d["kind"] == "library")
+        assert lib["id"] == "library"
+        assert lib["label"] == "Library"
+
+    async def test_project_filtering_excludes_other_user_projects(self, client, app):
+        await app.state.project_store.create_project(
+            name="Other Project",
+            slug="other-project",
+            created_by="other-user",
+            user_id="other-user",
+        )
+
+        reg = await client.post("/api/devices/register", json={"platform": "ios"})
+        token = reg.json()["scoped_token"]
+
+        resp = await client.get(
+            "/api/share/destinations",
+            headers={"Authorization": f"Bearer {token}"},
+        )
+        assert resp.status_code == 200
+        body = resp.json()
+        slugs = {d["id"] for d in body["destinations"] if d["kind"] == "project_files"}
+        assert "other-project" not in slugs
+
+    async def test_response_shape_and_correct_kinds(self, client, app):
+        reg = await client.post("/api/devices/register", json={"platform": "ios"})
+        token = reg.json()["scoped_token"]
+
+        resp = await client.get(
+            "/api/share/destinations",
+            headers={"Authorization": f"Bearer {token}"},
+        )
+        assert resp.status_code == 200
+        body = resp.json()
+        assert "destinations" in body
+        assert isinstance(body["destinations"], list)
+
+        for entry in body["destinations"]:
+            assert "kind" in entry
+            assert "id" in entry
+            assert "label" in entry
+            assert entry["kind"] in ("library", "project_files", "agent_chat")
+
+    async def test_agent_chat_includes_active_agents_in_user_channels(self, client, app):
+        registry = app.state.agent_registry
+        if registry._db is not None:
+            await registry.close()
+        await registry.init()
+
+        try:
+            primary = app.state.auth.get_primary_user()
+            user_id = primary["id"] if primary else "admin"
+
+            agent_record = await registry.register(
+                display_name="Test Agent",
+                framework="test",
+                user_id=user_id,
+            )
+            canonical_id = agent_record["canonical_id"]
+
+            await app.state.chat_channels.create_channel(
+                name="dm",
+                type="dm",
+                created_by=user_id,
+                members=[user_id, canonical_id],
+            )
+
+            reg = await client.post("/api/devices/register", json={"platform": "ios"})
+            token = reg.json()["scoped_token"]
+
+            resp = await client.get(
+                "/api/share/destinations",
+                headers={"Authorization": f"Bearer {token}"},
+            )
+            assert resp.status_code == 200
+            body = resp.json()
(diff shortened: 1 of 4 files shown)


jaylfc:
**nemotron-ultra-orB review** 

Now let me do the thorough code review of the diff:
Now I'll output the comprehensive code review:
VERDICT: Approved with minor findings
- tinyagentos/routes/share.py:28 - Missing authorization check: endpoint returns ALL user's projects without verifying device belongs to that user (device has user_id but no cross-check with project ownership)
- tinyagentos/routes/share.py:41-56 - Agent chat discovery logic bug: check `if user_id_str not in members and "user" not in members` uses string "user" but members appear to be user IDs (UUIDs/strings), so condition is likely always true for non-admin users, causing incorrect channel filtering
- tinyagentos/routes/share.py:47-48 - Dead code: `label = member` assignment overwritten on line 52
- tinyagentos/routes/share.py:54 - `seen.add(member)` inside `if agent and agent.get("status") == "active":` block means inactive agents won't be added to `seen` and could be re-processed on subsequent channel iterations (minor duplication risk)
- tinyagentos/auth_middleware.py:10 - Adding `/api/share/destinations` to EXEMPT_PATHS allows unauthenticated access but endpoint calls `require_device()` which enforces auth - inconsistent (remove from EXEMPT_PATHS or make endpoint truly public)
- tests/test_routes_share_destinations.py:34 - Test `test_library_always_present_for_paired_device` uses `client` fixture but also `app` fixture - inconsistent with other tests that create their own client
- tests/test_routes_share_destinations.py:85 - Test `test_agent_chat_includes_active_agents_in_user_channels` mutates global `app.state.agent_registry` (closes/reopens) causing potential test pollution - should use fixture to isolate registry state
- tests/test_routes_share_destinations.py:96 - Test creates channel with members `[user_id, canonical_id]` but doesn't verify channel type is "dm" or that agent is actually in a DM with user - weak assertion

jaylfc:
**nemotron-super review** 

VERDICT: Blocking issues found
- tests/test_routes_share_destinations.py:116: Test modifies app state by closing registry and setting app.state.agent_registry to closed registry, risking test interference
- tinyagentos/routes/share.py:42: Magic string "user" used without explanation reduces clarity; consider constant or comment
- tinyagentos/routes/share.py:45: Same magic string "user" in tuple check; same clarity issue

jaylfc:

VERDICT: Needs fixes - security bypass, logic errors, and test gaps

- **Security bypass in `tinyagentos/auth_middleware.py:10`**: Adding `/api/share/destinations` to `EXEMPT_PATHS` allows unauthenticated access. The route itself uses `require_device()` which validates a device token, but the exemption means the middleware skips auth entirely before reaching the route handler. This creates a window where the endpoint is accessible without any token validation at the middleware layer.

- **Logic bug in `tinyagentos/routes/share.py:44-45`**: The condition `if user_id_str not in members and "user" not in members:` incorrectly filters channels. It should be `or` not `and` - a channel should be included if the user is a member OR if "user" (generic placeholder) is a member. Current logic only includes channels where BOTH are present.

- **Logic bug in `tinyagentos/routes/share.py:47-48`**: The inner loop `if member in (user_id_str, "user"): continue` skips the user themselves, but then tries to treat remaining members as agents. This assumes all non-user members are agents, but channels can have multiple human users. No verification that `member` is actually an agent canonical_id.

- **Silent failure in `tinyagentos/routes/share.py:53-58`**: The `except RuntimeError: pass` swallows all runtime errors from `registry.get()`, including legitimate issues like DB connection failures. Should at minimum log the error.

- **Test gap in `tests/test_routes_share_destinations.py`**: No test for the `agent_chat` filtering logic when:
  - Channel has multiple human users (non-agent members)
  - Agent exists but status != "active"
  - Registry is unavailable/throws non-RuntimeError
  - User has no channels at all

- **Test gap**: `test_unauthenticated_returns_401` will now fail because the path is in EXEMPT_PATHS - the middleware won't check auth, and `require_device()` will likely return 401 but the test expectation may not match actual behavior.

- **Style**: `tinyagentos/routes/share.py:32` hardcodes `"library"` as id/label - consider constants. Line 42 uses `str(user_id)` but `user_id` from device may already be string.
VERDICT: Needs fixes - security bypass, logic errors, and test gaps

- **Security bypass in `tinyagentos/auth_middleware.py:10`**: Adding `/api/share/destinations` to `EXEMPT_PATHS` allows unauthenticated access. The route itself uses `require_device()` which validates a device token, but the exemption means the middleware skips auth entirely before reaching the route handler. This creates a window where the endpoint is accessible without any token validation at the middleware layer.

- **Logic bug in `tinyagentos/routes/share.py:44-45`**: The condition `if user_id_str not in members and "user" not in members:` incorrectly filters channels. It should be `or` not `and` - a channel should be included if the user is a member OR if "user" (generic placeholder) is a member. Current logic only includes channels where BOTH are present.

- **Logic bug in `tinyagentos/routes/share.py:47-48`**: The inner loop `if member in (user_id_str, "user"): continue` skips the user themselves, but then tries to treat remaining members as agents. This assumes all non-user members are agents, but channels can have multiple human users. No verification that `member` is actually an agent canonical_id.

- **Silent failure in `tinyagentos/routes/share.py:53-58`**: The `except RuntimeError: pass` swallows all runtime errors from `registry.get()`, including legitimate issues like DB connection failures. Should at minimum log the error.

- **Test gap in `tests/test_routes_share_destinations.py`**: No test for the `agent_chat` filtering logic when:
  - Channel has multiple human users (non-agent members)
  - Agent exists but status != "active"
  - Registry is unavailable/throws non-RuntimeError
  - User has no channels at all

- **Test gap**: `test_unauthenticated_returns_401` will now fail because the path is in EXEMPT_PATHS - the middleware won't check auth, and `require_device()` will likely return 401 but the test expectation may not match actual behavior.

- **Style**: `tinyagentos/routes/share.py:32` hardcodes `"library"` as id/label - consider constants. Line 42 uses `str(user_id)` but `user_id` from device may already be string.

jaylfc:
Reviewed, including verifying both bot findings against the code rather than taking them at face value. **Good PR** - real tests (401 unauthenticated, cross-user project filtering, response shape), and the design is right.

**The EXEMPT_PATHS addition is NOT a security hole**, to head off that reading: the route calls `require_device(request)` itself, which 401s without a valid device bearer, and the test asserts it. The middleware exemption is what lets a device-token caller past the *cookie* gate to reach the route's own check - the same pattern as the registry pubkey path. Correct approach.

Three things before it merges, all fix-forward on this branch:

**1. doc-gate is red, correctly.** A new API route module was added, so the gate wants `docs/agent-coordination.md` updated or a `Docs-Reviewed:` trailer. Reproduced locally: `DOC-GATE FAIL: routes -- an API route module was added or removed`. Document the endpoint properly rather than using the trailer escape hatch - this is a new agent-reachable surface and the docs are how agents discover it.

**2. Missing CSRF dependency (Qodo #1) - confirmed real.** `app.include_router(share_router)` on line 66 has no `dependencies=_csrf`, while every sibling does (`agents_router` on line 48). No live vulnerability today since the router is GET-only, but the moment S2B adds a POST here it is unprotected by default. Fix it now while it is one word.

**3. Agent chat destinations are silently omitted (Qodo #4) - confirmed real.** `share.py` resolves each channel member via `registry.get(member)`, which keys on canonical_id, but `routes/agents.py:703` creates DM channels with `members=["user", body.name]` - the agent **slug**, not the canonical id. So locally deployed agents never appear as share destinations. That is the main feature of the endpoint failing quietly for the most common case, and the current tests do not cover it.

Also note: **this conflicts with #2145**, which touches the same `EXEMPT_PATHS` line. Whichever lands second rebases - do not resolve it by reverting the other one's paths.

jaylfc:
Merging. I checked the auth-exempt change carefully because that is exactly where a public-data hole hides, and it is sound.

Adding `/api/share/destinations` to EXEMPT_PATHS looks alarming but is correct: this is a paired-DEVICE (taOSc) endpoint, so it is exempt from the SESSION middleware and does its own auth via `require_device(request)`, which 401s without a valid device token (that is why `test_unauthenticated_returns_401` passes). It then scopes every result to the token's user: `list_for_user(user_id)` for projects, and channels filtered by `user_id in members`. No cross-user leak - the project-filtering test confirms it. Same pattern as the other device/agent-token endpoints already in the exempt set.

Clean, gated, scoped, tested. Good.

Issue #2150, opened 2026-07-26
EPIC: taOS as the harness wrapper - agent comms move into taOStalk + Projects


jaylfc:
# EPIC: taOS as the harness wrapper - move all agent comms into taOS

Jay, 2026-07-26: all communication with the lead agents moves out of the Claude
app and into taOS. taOStalk becomes the chat surface, Projects becomes the work
surface, and the A2A realtime stream is the transport. The target is using taOS
to build taOS, on stream.

## What already exists (verified in code, not assumed)

- **`request_decision` agent tool** -> `DecisionStore` -> Decisions app, with
  `single_select`, `multi_select`, `approve_deny`, `free_text`, options
  normalised to `[{label, value}]`, priorities, and a notification on raise.
- **Decisions app** renders and answers those choices today.
- **Chat messages already carry `content_blocks`** (`chat/message_store.py`).
- **A2A realtime stream is live**: authenticated SSE proxy, verified end to end.
- **taOStalk s1 renderer cards** are already on the board (text, thinking,
  tool-call, status/question, live badge, connect-session wizard).

## The actual gap

**Decisions render only in the Decisions app.** No chat component renders a
decision, so a question raised by an agent is invisible in the conversation where
it was asked. That is the seam Jay named, and it is the difference between taOS
being a viewer and taOS being the place the work happens.

## Design: one decision, two surfaces, one resolution

- **`DecisionStore` stays the single source of truth.** Chat does NOT get its own
  question table. A question raised in chat IS a decision; the chat block is a
  *view* of it, keyed by `decision_id`.
- **A decision block in chat** renders the same options as buttons inline.
- **Answering from either surface resolves both**, live, via the existing broker
  so open clients update without a refresh. First answer wins.
- **Idempotency is the hard requirement**: the same decision is visible in two
  places at once, so a second answer must be rejected cleanly (not silently
  overwrite, not 500), and both surfaces must show the resolved state with who
  answered and when. Assume the user has chat open on a phone and Decisions open
  on a desktop, and taps both.
- **Nothing is lost when answered elsewhere**: a decision answered in the
  Decisions app must visibly resolve in the chat thread it came from, in place.

## Slices

- **W1** decision blocks render inline in chat, read-only first (see it, cannot
  answer it) - proves the join without touching write paths.
- **W2** answer from chat, with idempotent first-answer-wins and live
  propagation to every open surface.
- **W3** taOStalk s1 renderers (already carded) so agent output reads correctly:
[...]

Issue #2142, opened 2026-07-26
EPIC: taOS Teams - unified agent team app (dispatcher + benchmarks + org model)


jaylfc:
# taOS Teams (epic)

Jay, 2026-07-26: the dispatcher (#2141) and the Council (role-benchmarked
model team) are the same concern split in two - one knows WHO is good at WHAT,
the other decides WHAT goes to WHOM. Merge them into a single app: **taOS
Teams**. The name is deliberate: this is also the natural home for the
already-built agent org model (roles, titles, reporting lines, delegation),
which has backend but no first-class UI.

## What Teams is

The place a user assembles and runs their agent team:

- **Members**: the user's agents (registry identities), each with role/title
  (org-model data, already in the backend), current model (synced
  model-management source of truth), and project memberships/grants.
- **Routing**: the dispatcher - auto-assign claimable cards across the team,
  per the #2141 S1 policy (priority-then-oldest, blocked-on, per-agent caps,
  grants never bypassed, off by default).
- **Benchmarks**: role-benchmarked model data (Council gauge DB, community +
  local runs). Surfaced per member ("stepflash's model scores X on docs
  work") and fed into routing as a TIEBREAK within eligibility, never a gate.
- **Org view**: roles, reporting lines, delegation - the #161 backend gets
  its UI here.

## Slices

- S1: dispatcher assignment engine (#2141 S1, card tsk-rpxqsj) - IN FLIGHT,
  scope unchanged, it simply lives under the Teams umbrella.
- S2: Teams app shell (desktop): members list w/ role, model, grants, live
  claimed-card count; dispatcher on/off + config.
- S3: benchmark store + role taxonomy (role:<x> card labels) + scoring
  tiebreak in the S1 selection seam (#2141 comment has the design).
- S4: org chart view + delegation surfaces (backend exists).
- Later: community gauge sync (share/consume benchmark results, taOSnet).

## Naming

App name "Teams" in the launcher; "Council" is retired as a product name and
survives only as the internal name for the benchmark methodology docs.

PR #2157, opened 2026-07-27
docs: mark superseded designs after the 2026-07-26 decisions

jaylfc:
Adds supersede banners to eight docs that now contradict decisions made on 2026-07-26. **Nothing is deleted** - the banners point at the current statement and the docs stay readable as history.

Found by an audit sweep, not by chance, which is the point: these decisions were made in conversation and the docs kept saying the old thing.

**The urgent one: `docs/design/taostalk-acp-bridge.md`.** It proposes mirroring a CLI session into taOStalk *including PTY capture as a fallback* - exactly the approach Jay rejected. Free lanes are working taOStalk cards tonight and could reasonably read it as the current plan and build the wrong thing. That is the reason this PR exists now rather than in the morning.

| Doc | Why |
|---|---|
| `taostalk-acp-bridge.md` | session mirroring / PTY capture - rejected; the A2A thread IS the conversation |
| `taostalk-slice1.md` | partial - the session-capture half only; renderers remain valid |
| `message-hub-core.md` | chat moves onto the taOSmd bus; history migrates |
| `runbooks/framework-swap.md` | taOSmd is no longer swappable - it becomes a required dependency |
| `projects-nested-elements.md` | one board spanning repos is not the model; one project one repo, cross-repo work is a request |
| `taos-council.md` | Council retired as a product name; this is taOS Teams |
| mac-app c1 design + plan | menubar client obsolete; taOSgo clients get remade |

Current statements: #2150 (chat on bus, no mirroring), #2142 (Teams), and the one-board-one-repo rule.

doc-gate passes locally. No code changes.

## Summary by CodeRabbit

- **Documentation**
  - Introduced a unified scheduled collaboration model for agents, including identity, project, board, review, and communication workflows.
  - Clarified ACP onboarding versus cron-based ongoing collaboration.
  - Updated design guidance to establish A2A threads as the conversation source and taOSmd as the future chat-history foundation.
  - Clarified project and repository boundaries, replacing “Council” with “taOS Teams.”
  - Marked superseded or partially superseded documentation and preserved it for historical reference.


jaylfc:
Folded the findings that were right, and said why on the two I am not acting on now.

**Corrected: the Council rename scope.** The finding is accurate. `tinyagentos/council` and `/api/council/*` both still exist, so my blanket "Council is retired" implied those identifiers were stale when they are not. The banner now scopes the retirement to user-facing text and states that the package, route prefix and benchmark methodology keep the name, with renaming them called out as a separate change. Verified both exist in the tree before changing it rather than taking the finding on trust.

**Cleared 31 em dashes** from the files this PR already touches. Character swap only.

**Not acting on two, both real, both blocked on work that has not happened yet:**

- *taostalk-slice1 boundary.* Correct that the retained plan still translates A2A envelopes into `chat_messages` rows via a synthetic session channel, which sits awkwardly beside "the A2A thread is the conversation". That is a live design question, not a documentation defect, and resolving it in a supersede banner would be inventing a decision. The banner's job is to stop someone building the killed part; it does that.
- *framework-swap chat migration contract.* Also correct: the runbook still treats `data/chat.db` as the history source while the design says the bus. The authoritative migration contract cannot be written until the exporter lands, which is blocked on taOSmd's `/a2a/import` endpoint. Writing a cutover contract now would produce a document that is wrong in a different way.

Both are worth doing once unblocked. Neither is a reason to hold a PR whose entire purpose is warning readers off superseded designs.

The two findings on `agent-collaboration-model.md` are moot: that file was swept into this PR by a `git add -A` and I have since removed it. It was never mine to commit.

PR #2162, opened 2026-07-27
DO NOT MERGE: verify the sharded aggregate gate reports red

jaylfc:
Throwaway branch. Contains one deliberately failing test to prove the aggregate `test (3.12)`/`test (3.13)` jobs report **failure** when a shard fails, rather than being skipped.

This matters because `needs: shards` without `if: always()` would SKIP the gate on shard failure, and GitHub counts a skipped required check as satisfied - a red suite would show a green required check. The success path is already proven on #2137; this proves the half that actually protects us.

Will be closed and the branch deleted as soon as the checks report.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,15 @@ jobs:
   # run at once instead of queueing behind the 2 self-hosted boxes. The VPS +
   # Fedora runners stay free for the kilo lane and GPU/bench work.
 
-  test:
+  # Renamed from "test" to "shards" because sharding changes the check-run names
+  # this job reports. Branch protection matches required checks by NAME, so the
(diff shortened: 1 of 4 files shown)


jaylfc:
Verification complete, closing. Purpose served.

Round 1 caught a real defect: the aggregate gates are a matrix and fail-fast defaults to true, so the first to fail cancelled its sibling. Result was `test (3.12) failure, test (3.13) cancelled`. Never unsafe, since cancelled is not success and protection still blocks, but which version got reported as broken depended on which gate lost the race.

Round 2 with `fail-fast: false`: **both gates report failure independently.**

So #2137 is now proven on both paths, green when the suite passes and red when it does not, which is the half that actually protects the branch.

PR #2163, opened 2026-07-27
docs: write down the verification and collision rules we work by

jaylfc:
Jay asked for our working conventions to be written down so other agents can work the way we have been.

**Where it lives, and why not in three places.** `CONTRIBUTING.md` gets the canonical text, because it applies to every contributor, human or agent. The guest-agent skill (`.claude/skills/taos-development-skill/SKILL.md`) **points at those sections rather than repeating them**, using the "Upstream conventions (from CONTRIBUTING.md)" pattern already established there, and keeps only what is specific to agents. Three copies of the same rules would drift within a fortnight.

Note `CLAUDE.md` is gitignored, so it was not a viable home: anything written there reaches one machine.

**What is captured.** Every recurring problem here has had the same shape, a signal that looked like success without being one:

- a PR passed all five required checks containing only a version bump
- a test file reported green because every test in it skipped behind `pytest.importorskip`, asserting nothing
- `cmd | tail && echo OK` printing OK after `cmd` failed, because `&&` sees `tail`
- green on a conflicted branch, which tested the old base rather than the merge

None of these are fixable by adding another check, because the check is the thing lying. So they are written as habits, with the rule that anything acting as a gate must be proven to fail on purpose before it is trusted.

The second section covers collisions: two PRs that both **add** the same path cannot both merge, and whichever lands first silently decides the design. That has cost duplicate implementations of the same feature twice, both found at merge time rather than when the branches opened. Also documents that renaming a CI job orphans its required check, since protection matches by name.

Some of how we operate is deliberately not public and stays out of this.

Drive-by correction: the test count said ~3,590 against a suite of ~10,250.

## Summary by CodeRabbit

* **Documentation**
  * Expanded contributor guidance for working with AI coding agents.
  * Added clearer instructions for verifying changes, interpreting CI results, and reporting failures or skipped checks.
  * Added guidance to avoid conflicts with other contributors and duplicate CI checks.
  * Updated contribution, testing, architecture, style, and licensing documentation for consistency and clarity.

### DIFF
--- a/.claude/skills/taos-development-skill/SKILL.md
+++ b/.claude/skills/taos-development-skill/SKILL.md
@@ -429,6 +429,32 @@ Run `scripts/install-git-hooks.sh` to enable local hooks (`.githooks/pre-commit`
   are available. Most modules use `from __future__ import annotations`.
 - Code style: match surrounding code, one concern per module
 - Use `uv` for dependency management and test running: `uv sync --extra dev`, `uv run pytest`
+- **Read `CONTRIBUTING.md` sections "Verifying your work" and "Avoiding collisions with
+  other contributors".** They are the canonical statement of both; this file does not
+  restate them so the two cannot drift. What follows is only the part specific to
+  working here as an agent.
+
+## Verification (agent specifics)
+
+Green is a claim, not evidence. The repo-level cases are in CONTRIBUTING.md. These are
+the ones that bite agents in particular:
+
+- **Do not report work as done without a PR link.** A branch with commits and no PR is
+  not delivered. A lane once announced completion on an empty branch; the check now runs
+  before any completion is claimed, and the same standard applies to you.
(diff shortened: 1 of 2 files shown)


jaylfc:
Added the reverse reference, per Jay: this file now points at the development skill for anyone contributing through a coding agent.

The pointer only ran one way before. The skill referenced CONTRIBUTING.md, but nothing in CONTRIBUTING.md mentioned the skill, and it lives in a dot-directory, so an agent arriving at the repo had no way to find it short of guessing the path.

Verified rather than assumed: `.claude` matches a gitignore pattern, but `SKILL.md` and `soul.md` are tracked, so a fresh clone does get them and the "it ships in the repo" claim holds.

The two files still deliberately do not restate each other. CONTRIBUTING carries the conventions, the skill carries how to apply them here, and each says so.

PR #2165, opened 2026-07-27
fix(desktop): add CSRF protection to mutating requests in library, framework-api, memory-api, x-monitor

### LINKED ISSUE #2126: BUG: library.ts mutating requests bypass CSRF (raw fetch, no withCsrf)
Surfaced by Qodo on #2117, but pre-existing and wider than that PR, so filing separately rather than expanding a one-line change.

Every mutating call in `desktop/src/lib/library.ts` uses raw `fetch()` without `withCsrf()` / `taosFetch`, so the `X-CSRF-Token` header is omitted:

- `deleteLibraryItem` (DELETE, around line 102)
- `reprocessLibraryItem` (POST, around line 114)
- `ingestLibraryUrl` (POST, around line 138)

Where CSRF protection is enforced these will fail with 403 in production. The read paths (`listLibraryItems`, `getLibraryItem`) are fine, since GETs are not the concern.

## Fix

Route the three mutating calls through the project's CSRF-aware wrapper, matching how other apps issue mutating requests. Worth a quick grep for the same pattern in other `src/lib/*.ts` clients while in there, since this is the kind of thing that gets copied between modules.

## Note

Not introduced by #2117. That PR touched one line in `ingestLibraryUrl`, which is what drew the reviewer's eye to the surrounding function.

hognek:
## Summary
Adds CSRF protection via a globally-installed auth guard to all three desktop entry points.

Previous PR state: installAuthGuard() was only in main.tsx (desktop shell PWA).
chat-main.tsx (chat PWA) and app-standalone-main.tsx (standalone app PWA) had no guard,
so mutating API calls made from those PWAs would fail CSRF validation.

## jaylfc review follow-up (2026-07-27)
- Added installAuthGuard() to **desktop/src/chat-main.tsx** (chat PWA entry)
- Added installAuthGuard() to **desktop/src/app-standalone-main.tsx** (standalone app PWA entry)
- New test **desktop/src/__tests__/entry-guards.test.ts** asserting all three entry modules install the guard on import

## Testing
- 3 new entry-guard tests: PASS
- Full vitest suite: 373 files, 3177 tests: ALL PASS
- SPA build: clean (4.84s)
- TypeScript: clean (tsc --noEmit)

Closes #2126

## Summary by CodeRabbit

* **New Features**
  * Improved desktop startup reliability by ensuring the authentication/session-expiration guard is installed consistently across all desktop entry points.

* **Bug Fixes**
  * Strengthened request security by adding CSRF protection to framework updates, permitted-model updates, library mutations, memory updates, and monitoring JSON mutations.
  * Standardized JSON request header handling so `Accept: application/json` is applied consistently while correctly merging caller-provided headers (including `Headers` instances).

* **Tests**
  * Added coverage verifying auth-guard installation across desktop entry points.
  * Added a regression test confirming CSRF tokens are included for JSON watch requests.

### DIFF
--- a/desktop/src/__tests__/entry-guards.test.ts
+++ b/desktop/src/__tests__/entry-guards.test.ts
@@ -1,0 +1,70 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+// Use vi.hoisted so the mock reference is stable across vi.resetModules() calls.
+// vi.mock factories only run once; vi.resetModules clears the module cache but
+// does not re-invoke the factory, so the returned vi.fn() instance persists.
+// mockClear() between tests resets the call count so each assertion sees only
+// the calls from the current test's dynamic import().
+const { installAuthGuard: mockGuard } = vi.hoisted(() => ({
+  installAuthGuard: vi.fn(),
+}));
+
+vi.mock("../lib/auth-guard", () => ({
+  installAuthGuard: mockGuard,
+  SESSION_EXPIRED_EVENT: "taos-session-expired",
+}));
+
+// Entry modules call createRoot(document.getElementById("root")!).render(...).
+// Stub createRoot so the render tree doesn't actually mount (components pull in
+// real stores, contexts, and side-effects we don't need for this assertion).
+vi.mock("react-dom/client", () => ({
+  createRoot: vi.fn(() => ({ render: vi.fn() })),
+}));
+
+// Some entry modules import AppShell / AppStandalone / ChatStandalone which
+// transitively pull in heavy dependency trees. Stub them as no-op components.
+vi.mock("../components/AppShell", () => ({
+  AppShell: ({ children }: { children?: React.ReactNode }) => children ?? null,
+}));
+
+vi.mock("../App", () => ({ App: () => null }));
+vi.mock("../ChatStandalone", () => ({ ChatStandalone: () => null }));
+vi.mock("../AppStandalone", () => ({ AppStandalone: () => null }));
+
+vi.mock("../stores/theme-store", () => ({
+  restoreActiveTheme: vi.fn(),
+  installWebkitRepaintGuards: vi.fn(),
+}));
+
+vi.mock("../registry/app-registry", () => ({
+  getApp: vi.fn(() => undefined),
+}));
+
+vi.mock("../lib/client-log", () => ({
+  installGlobalErrorReporting: vi.fn(),
+}));
+
+describe("entry module auth guards", () => {
+  beforeEach(() => {
+    vi.resetModules();
+    mockGuard.mockClear();
+    // Every entry module calls createRoot(document.getElementById("root")!).
+    // Provide the element so the non-null assertion (!) doesn't throw.
+    document.body.innerHTML = '<div id="root"></div>';
+  });
+
+  it("desktop shell (main.tsx) installs the auth guard", async () => {
+    await import("../main");
+    expect(mockGuard).toHaveBeenCalledTimes(1);
+  });
+
+  it("chat PWA (chat-main.tsx) installs the auth guard", async () => {
+    await import("../chat-main");
+    expect(mockGuard).toHaveBeenCalledTimes(1);
+  });
+
+  it("standalone app PWA (app-standalone-main.tsx) installs the auth guard", async () => {
+    await import("../app-standalone-main");
+    expect(mockGuard).toHaveBeenCalledTimes(1);
+  });
+});
--- a/desktop/src/app-standalone-main.tsx
+++ b/desktop/src/app-standalone-main.tsx
@@ -7,6 +7,9 @@ import { restoreActiveTheme, installWebkitRepaintGuards } from "./stores/theme-s
 import { getApp } from "./registry/app-registry";
 import "./theme/tokens.css";
 
+// Wrap window.fetch so any 401 from /api/* triggers a session-expired
+// event that LoginGate picks up and shows the login screen (same guard
+// installed by main.tsx for the desktop shell PWA).
 installAuthGuard();
 
 // Apply the user's persisted theme on boot, same as chat-main.tsx.
--- a/desktop/src/chat-main.tsx
+++ b/desktop/src/chat-main.tsx
@@ -6,6 +6,9 @@ import { installAuthGuard } from "./lib/auth-guard";
 import { restoreActiveTheme, installWebkitRepaintGuards } from "./stores/theme-store";
 import "./theme/tokens.css";
 
+// Wrap window.fetch so any 401 from /api/* triggers a session-expired
+// event that LoginGate picks up and shows the login screen (same guard
+// installed by main.tsx for the desktop shell PWA).
 installAuthGuard();
 
 // Apply the user's persisted theme (light/dark/etc.) on boot, the same as the
--- a/desktop/src/lib/agent-browsers.ts
+++ b/desktop/src/lib/agent-browsers.ts
@@ -36,7 +36,9 @@ export interface CookieEntry {
 
 async function fetchJson<T>(url: string, fallback: T, init?: RequestInit): Promise<T> {
   try {
-    const res = await fetch(url, { ...init, headers: { Accept: "application/json", ...init?.headers } });
+    const headers = new Headers(init?.headers);
+    headers.set("Accept", "application/json");
+    const res = await fetch(url, { ...init, headers });
     if (!res.ok) return fallback;
     const ct = res.headers.get("content-type") ?? "";
     if (!ct.includes("application/json")) return fallback;
--- a/desktop/src/lib/github.test.ts
+++ b/desktop/src/lib/github.test.ts
@@ -43,7 +43,7 @@ describe("fetchStarred", () => {
     const [url, opts] = fetchMock.mock.calls[0];
     expect(url).toContain("/api/github/starred");
     expect(url).toContain("page=1");
-    expect(opts.headers.Accept).toBe("application/json");
+    expect(opts.headers.get("Accept")).toBe("application/json");
   });
 
   it("returns empty result on non-ok response", async () => {
@@ -96,7 +96,7 @@ describe("fetchNotifications", () => {
 
     const [url, opts] = fetchMock.mock.calls[0];
     expect(url).toBe("/api/github/notifications");
-    expect(opts.headers.Accept).toBe("application/json");
+    expect(opts.headers.get("Accept")).toBe("application/json");
   });
 
   it("returns empty result on non-ok response", async () => {
@@ -136,7 +136,7 @@ describe("fetchRepo", () => {
 
     const [url, opts] = fetchMock.mock.calls[0];
     expect(url).toBe("/api/github/repo/octocat/hello-world");
-    expect(opts.headers.Accept).toBe("application/json");
+    expect(opts.headers.get("Accept")).toBe("application/json");
   });
 
   it("returns null on non-ok response", async () => {
@@ -192,7 +192,7 @@ describe("fetchIssues", () => {
     expect(url).toContain("/api/github/repo/octocat/hello-world/issues");
     expect(url).toContain("state=open");
     expect(url).toContain("page=2");
-    expect(opts.headers.Accept).toBe("application/json");
+    expect(opts.headers.get("Accept")).toBe("application/json");
   });
 
   it("returns empty result on non-ok response", async () => {
@@ -399,7 +399,7 @@ describe("startDeviceFlow", () => {
     const [url, opts] = fetchMock.mock.calls[0];
     expect(url).toBe("/api/github/oauth/device/start");
     expect(opts.method).toBe("POST");
-    expect(opts.headers.Accept).toBe("application/json");
+    expect(opts.headers.get("Accept")).toBe("application/json");
   });
 
   it("throws on non-ok response", async () => {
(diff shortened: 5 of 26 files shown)


jaylfc:
Reviewed. The change is correct and I am not blocking it, but I found something while checking it that makes this PR more important than it looks, and it is two lines from being complete. Please fold that in here rather than opening a follow-up.

**First, what I checked and got wrong so nobody repeats it.** My initial suspicion was that `withCsrf` would lose `Content-Type` when a caller passed a `Headers` instance. It does not: `csrf.ts:31` uses `new Headers(init?.headers)`, which normalises plain objects, arrays and `Headers` alike. No bug there.

**Second, the thing that actually matters.** My next thought was that these call sites are redundant, because `auth-guard.ts` already patches `window.fetch` and its own comment says it covers CSRF "at every call site, not only the handful that wrap withCsrf() by hand". That would have made this PR churn. It is not, because the guard is not installed everywhere:

```
$ grep -rn "installAuthGuard()" desktop/src --include=*.tsx
desktop/src/main.tsx:13:installAuthGuard();
```

One entry point. There are three:

| entry point | bundle | installAuthGuard | restoreActiveTheme |
|---|---|---|---|
| `main.tsx` | desktop shell | yes | yes |
| `chat-main.tsx` | chat PWA | **no** | yes |
| `app-standalone-main.tsx` | standalone app PWA | **no** | yes |

`chat-main.tsx` and `app-standalone-main.tsx` both install the theme restore and the WebKit repaint guard, and both skip the auth guard. So in the chat PWA and in every standalone app PWA:

1. **No global CSRF header.** Nothing attaches `X-CSRF-Token` except call sites that wrap `withCsrf` by hand, which is exactly what this PR is adding. So these changes are load bearing in those bundles, not defence in depth.
2. **No 401 handling.** `main.tsx:9-12` documents why the guard exists: a stale cookie "left the SPA rendering empty data instead of prompting for re-login". That regression is still live in both PWA bundles today.

**Requested change, in this PR:** add `installAuthGuard()` to `chat-main.tsx` and `app-standalone-main.tsx`. It is idempotent and already covered by `auth-guard.test.ts`, so the risk is close to zero and it closes both holes at the source instead of one call site at a time.

**Please also add a test that would have caught this**, since the current suite passes with the guard installed in only one of three entry points. A test asserting each entry module installs the guard is enough. Without it the same gap reappears the next time an entry point is added, and there is now a fourth on the way.

Worth stating plainly: the per-call-site `withCsrf` approach cannot be verified by inspection once there are 52 of them, whereas one guard at each entry point can. Keep the call-site wrapping you have added here, since it is harmless and helps under mocked fetch in tests, but the guard is what should be relied on.

Doc sweep and CHANGELOG in the same PR per the commit gate.

jaylfc:
This went CONFLICT because #2174 merged and touched `auth-guard`. Needs a rebase onto current `origin/dev`.

Worth knowing what landed, because it changes what this PR is for. #2174 installed `installAuthGuard()` in **all three** SPA entry points (it was only in `main.tsx`, so the chat PWA and every standalone app PWA had no global CSRF attachment **and** no 401 session-expired handling). The guard patches `window.fetch` and attaches `X-CSRF-Token` to same-origin mutating requests, so with that merged, the per-call-site `withCsrf` wrapping in this PR is now genuine defence in depth rather than the only protection in those bundles.

That is not a reason to drop it. Keep the explicit wrapping: it is harmless (`withCsrf` only sets the header when absent) and it keeps the call sites correct under a mocked fetch in tests, where the global guard is not installed. Just rebase and expect the diff to read differently now.

#2174 also added a coverage test that enumerates entry points from `vite.config.ts` and fails if any is missing the guard, so a fourth entry point cannot silently ship unprotected. Nothing for you to do there, but do not be surprised by it in the merge.

jaylfc:
**Correcting myself: the Headers-spread bug IS real, and it is in this PR's blast radius.** Earlier I said my suspicion was wrong. I had checked `csrf.ts`, which is fine, and wrongly concluded the whole class did not exist. The bug is at the call site.

`desktop/src/lib/x-monitor.ts:54`:

```ts
const res = await fetch(url, { ...init, headers: { Accept: "application/json", ...init?.headers } });
```

`withCsrf` returns `headers` as a **`Headers` instance** (`csrf.ts:31`, `new Headers(init?.headers)`). Spreading a `Headers` instance into an object literal yields an empty object, because its entries are not own enumerable properties. Measured:

```
spreading a Headers instance yields: {"Accept":"application/json"}
CSRF token survived? false

spreading a PLAIN object yields:     {"Accept":"...","Content-Type":"...","X-CSRF-Token":"tok123"}
```

So the moment an x-monitor call is routed through `withCsrf`, **the token this PR exists to add is silently stripped at line 54**, and every `postJson`/`putJson` in that file inherits it. The failure is invisible: the request still sends, just without the header, so it fails the CSRF gate at the server rather than erroring client-side.

**Fix:** normalise instead of spreading, so it works for both shapes.

```ts
const headers = new Headers({ Accept: "application/json" });
new Headers(init?.headers).forEach((v, k) => headers.set(k, v));
const res = await fetch(url, { ...init, headers });
```

`new Headers(...)` already accepts a `Headers`, a plain object, or an array, which is exactly why `csrf.ts` is correct and this line is not.

**Please add a test that would have caught it**: call `fetchJson` with an init whose `headers` is a `Headers` instance carrying `X-CSRF-Token`, and assert the header reaches `fetch`. Prove it goes red against the current line 54 before fixing, otherwise the test is only decorating the fix.

This PR is also DIRTY now (#2174 merged and touched auth-guard), so you need a rebase regardless. Fold this into the same pass.

Credit where due: Qodo flagged this and I dismissed it after checking the wrong file. Its finding stands and mine did not.

jaylfc:
**The Headers fix is correct and I verified it rather than eyeballing it.** Ran the old and new logic side by side against exactly what `withCsrf` returns (a `Headers` instance carrying `X-CSRF-Token`):

```
OLD: token survives?              false
NEW: token survives?              true
NEW: Accept preserved?            true
NEW: plain-object init still works? true
```

That last line matters as much as the first: `new Headers(init?.headers)` accepts a `Headers`, a plain object or an array, so the fix does not break callers passing a plain object literal. That is why normalising is right and spreading was not.

**Only the rebase is left.** `mergeStateStatus` is DIRTY because `dev` moved under you again: #2050 (fail-closed signing) and #2185 (app tiering S1) landed in the last hour or so, on top of #2174 earlier.

**One thing to watch on this rebase specifically.** #2174 installed `installAuthGuard()` in all three SPA entry points, so the global `window.fetch` patch now attaches CSRF to same-origin mutating requests everywhere. Your per-call-site `withCsrf` wrapping is therefore defence in depth rather than the only protection. **Keep it** — it is harmless (`withCsrf` only sets the header when absent) and it keeps the call sites correct under a mocked fetch in tests, where the guard is not installed. Just do not be surprised if the diff reads differently against the new base.

And run the check that saved #2177: `git log <merge-base>..origin/dev` over the files you touch, before pushing. You are touching `desktop/src/lib/` which several merges have moved through tonight, and a rebase can silently drop merged work with no conflict marker.

Rebase and I will merge it. Thanks for turning that one around quickly, and credit to Qodo for the original catch that I initially dismissed.

jaylfc:
@hognek routing this to you rather than the free build lanes, per Jay.

This PR is CONFLICTING against dev, so none of the four required checks (test 3.12, test 3.13, lint, spa-build) have ever run on it. Worth stating precisely: a green check on a conflicted branch tested the OLD base, so "no red" here is evidence of nothing. Rebasing is what makes it testable at all.

It cannot go to the free lanes: their harness builds a fresh worktree off origin/dev and gates on producing a commit, so any work on an EXISTING branch produces nothing and the card is destroyed. I proved that the expensive way today, losing four cards to it.

Ask: rebase onto current origin/dev, preserve the author's commits, do not change behaviour while resolving, and comment listing what conflicted and how each hunk was resolved so a reviewer can check the resolution instead of re-deriving it. Please do NOT merge; the rebase is the whole job and I will review the new head.

Context: A2A 1760 went to the bus, which you are not on. That was my error, so this is the same request on the channel you actually use.

hognek:
## Rebase resolution (conflict report)

Rebased `fix/2126-library-csrf` onto `origin/dev` (`d728dd8`). One commit had conflicts:

**Commit `b772ce87` — `fix(desktop): install auth guard in chat and standalone app PWAs`**

Two files conflicted, both with the same pattern:

1. **`desktop/src/app-standalone-main.tsx`** — lines 10–15  
2. **`desktop/src/chat-main.tsx`** — lines 9–14  

**Root cause:** `origin/dev` already had the `installAuthGuard()` call (and its import) in both files — a separate PR had landed the same guard in the meantime. The commit on this branch added a comment block above that call explaining what it does.

**Resolution (both files):** Kept the comment block from this PR and the existing `installAuthGuard()` call from `origin/dev`. The `HEAD` side of the conflict was empty (no comment on dev), so no code was lost on either side. Net result: the comment documenting the auth guard is preserved with the already-working guard call.

jaylfc:
@coderabbitai review

jaylfc:
@coderabbitai full review

jaylfc:
## Rebase conflict resolutions

Rebased PR #2165 onto current origin/dev. Conflicts occurred in 8 files:

1. **desktop/src/lib/x-monitor.ts** - Conflict between PR's  pattern and dev's existing  + forEach pattern. Resolved by keeping the PR's  +  pattern, which preserves Headers instances returned by .

2. **desktop/src/lib/github.test.ts** - Multiple Accept/Content-Type assertion conflicts. Resolved by keeping the defensive dual-mode assertions: .

3. **desktop/src/lib/knowledge.test.ts** - Same Accept assertion pattern. Kept defensive dual-mode assertions.

4. **desktop/src/lib/mail.test.ts** - Same Accept/Content-Type assertion pattern. Kept defensive dual-mode assertions.

5. **desktop/src/lib/memory-api.test.ts** - Same Accept/Content-Type assertion pattern. Kept defensive dual-mode assertions.

6. **desktop/src/lib/memory.test.ts** - Same Accept/Content-Type assertion pattern. Kept defensive dual-mode assertions.

7. **desktop/src/lib/models.test.ts** - Same Accept/Content-Type assertion pattern. Kept defensive dual-mode assertions.

8. **desktop/src/lib/reddit.test.ts** - Same Accept/Content-Type assertion pattern. Kept defensive dual-mode assertions.

## Kilo review finding verdict

**Finding:** CRITICAL -  in x-monitor.ts spreads  with object syntax, which drops  instances returned by .

**Verdict: REAL (at the time of reporting)**

Verified against commit ca4d9d20 (the head where kilo reported the finding). The code at that commit used:
```ts
const res = await fetch(url, { ...init, headers: { Accept: application/json, ...init?.headers } });
```

This pattern silently drops Headers instances because spreading a Headers object into a plain object literal only copies enumerable own properties, and Headers entries are not own-enumerable. The X-CSRF-Token header was indeed being stripped.

**Current status:** The PR has since fixed this in commit f09616b6 / 2939f203 by switching to `new Headers(init?.headers)` and passing the Headers instance directly to fetch(). The latest kilo review on the PR head confirms "No Issues Found". No further action needed on this finding.

---

Rebase complete. New head: fbe62a92. All desktop vitest tests pass (724 tests). Build passes.

jaylfc:
## Rebase conflict resolutions

Rebased PR #2165 onto current origin/dev. Conflicts occurred in 8 files:

1. **desktop/src/lib/x-monitor.ts** - Conflict between PR's `new Headers(init?.headers)` pattern and dev's existing `new Headers({ Accept: ... })` + forEach pattern. Resolved by keeping the PR's `new Headers(init?.headers)` + `headers.set("Accept", "application/json")` pattern, which preserves Headers instances returned by `withCsrf()`.

2. **desktop/src/lib/github.test.ts** - Multiple Accept/Content-Type assertion conflicts. Resolved by keeping the defensive dual-mode assertions: `(opts.headers?.get?.("Accept") ?? opts.headers?.Accept)`.

3. **desktop/src/lib/knowledge.test.ts** - Same Accept assertion pattern. Kept defensive dual-mode assertions.

4. **desktop/src/lib/mail.test.ts** - Same Accept/Content-Type assertion pattern. Kept defensive dual-mode assertions.

5. **desktop/src/lib/memory-api.test.ts** - Same Accept/Content-Type assertion pattern. Kept defensive dual-mode assertions.

6. **desktop/src/lib/memory.test.ts** - Same Accept/Content-Type assertion pattern. Kept defensive dual-mode assertions.

7. **desktop/src/lib/models.test.ts** - Same Accept/Content-Type assertion pattern. Kept defensive dual-mode assertions.

8. **desktop/src/lib/reddit.test.ts** - Same Accept/Content-Type assertion pattern. Kept defensive dual-mode assertions.

## Kilo review finding verdict

**Finding:** CRITICAL - `fetchJson()` in x-monitor.ts spreads `init.headers` with object syntax, which drops `Headers` instances returned by `withCsrf()`.

**Verdict: REAL (at the time of reporting)**

Verified against commit ca4d9d20 (the head where kilo reported the finding). The code at that commit used object spread on `init?.headers`, which silently drops Headers instances because Headers entries are not own-enumerable properties. The X-CSRF-Token header was indeed being stripped.

**Current status:** The PR has since fixed this in later commits by switching to `new Headers(init?.headers)` and passing the Headers instance directly to `fetch()`. The latest kilo review on the PR head confirms "No Issues Found". No further action needed on this finding.

---

Rebase complete. New head: fbe62a92. All desktop vitest tests pass (724 tests). Build passes.

PR #2167, opened 2026-07-27
fix(messages): remove duplicate channel header causing stacked rendering

### LINKED ISSUE #2124: BUG: Messages shows the conversation header twice (duplicated name + toolbar + description)
Reported by Jay with a screenshot, 2026-07-26 (taOS on the Pi, desktop window, Messages open on the Mary DM).

## Symptom

The conversation header block renders **twice**, stacked. Each copy shows the full row: the `@` type icon, agent emoji, channel name ("Mary"), the `?` guide link, the pin/search actions, a member count (`2`), and the "Direct messages with Mary" description line beneath. Below the doubled block the pane is otherwise normal ("No messages yet. Say hello to @mary.").

## Investigation so far (three hypotheses RULED OUT)

I could not reproduce this headlessly (no browser available), so this is code and data analysis only. Recording the dead ends so nobody repeats them:

**1. Not an `isMobile` disagreement.** First theory: `MessagesApp` and `MobileSplitView` compute "mobile" independently, so they could disagree and render `MobileSplitView`'s nav bar (`detailTitle`) alongside the in-pane header. Disproved: `useIsMobile` (`src/hooks/use-is-mobile.ts`) and the local hook inside `MobileSplitView.tsx` are **identical** logic on the same 768 breakpoint, both driven by the viewport media query, so they cannot disagree. Also `MobileSplitView` returns early on desktop (around line 66) and **ignores `detailTitle` entirely**, so the desktop path renders no nav title at all.

**2. Not duplicate DM channels.** Checked the live channel list: exactly one DM channel for Mary (`ebafc8ec4bd4`) and one for Naira (`379bdca85328`). No duplicates. Unrelated observation worth its own look: there are **9 channels all named `a2a`**, a separate naming collision.

**3. Not a double render of the message area.** `messageAreaUI` is constructed once (`MessagesApp.tsx:1613`) and passed once as the `detail` prop (line 2048). It appears nowhere else in the tree.

## Where to look next

The in-pane header block is in `MessagesApp.tsx` around line 1642 (comment: "channel header — MobileSplitView owns back nav on mobile"). Since it is written once and its container renders once, the duplication most likely comes from either:

- an ancestor rendering the whole detail subtree twice (check whether the Messages window mounts two instances, e.g. the multi-window support, or a StrictMode double-mount leaving real DOM rather than being a dev-only artifact), or
- a stale cached bundle, showing a mix of old and new markup.

**First thing to try when reproducing:** hard-refresh to rule out a stale cached shell, then check whether the app is mounted twice (React DevTools, or count `[data-testid=mobile-split-view]` nodes in the DOM). If there are two split-view nodes, the bug is at the mount/window level and not in the header code at all.

## Related

The "Mobile Messages polish" task already carries a "trim duplicate header" item, so a mobile variant of this was seen before. That one was addressed by suppressing `detailTitle` on mobile (`MessagesApp.tsx:2045`). This report is the **desktop** case, which that fix explicitly does not cover (the comment there says "Desktop keeps it").

hognek:
## Problem

The conversation header block in Messages rendered twice on desktop — @ icon, agent emoji, channel name, guide link (❓), pin badge, search button, member count, and description all appeared stacked.

Closes #2124

## Root Cause

Both `MessagesApp.tsx` and `MessageList.tsx` contained an identical channel header:

- **MessagesApp.tsx** (formerly line 1642): Rendered inside `messageAreaUI`
- **MessageList.tsx** (line 214): Rendered at the top of the message list component

Since `messageAreaUI` rendered its own header AND then passed control to `<MessageList>` (which ALSO rendered its own header), the result was two identical headers stacked on top of each other.

## Fix

Removed the duplicate channel header from `MessagesApp.tsx` and cleaned up the five imports that became dead code:

- `MessagesSquare`, `Search` from lucide-react
- `PinBadge` from ./chat/PinBadge
- `PinnedMessagesPopover` from ./chat/PinnedMessagesPopover (type import preserved)
- `resolveAgentEmoji` from @/lib/agent-emoji

The MessageList component already handles all channel header rendering correctly, including the same event handlers for settings, pinned messages, threads, and search toggles. The MessageList tests confirm all 49 tests pass, including 4 channel-header-specific tests.

## Verification

- [x] TypeScript compilation: `npx tsc --noEmit` — zero errors
- [x] Desktop build: `npm run build` — succeeds
- [x] MessagesApp tests: `MessagesApp.agentSections.test.ts` / `MessagesApp.stallWatch.test.ts` — 22/22 pass
- [x] MessageList tests: `MessageList.test.tsx` — 49/49 pass (including channel header, pin badge, toggle buttons)

### DIFF
--- a/desktop/src/apps/MessagesApp.tsx
+++ b/desktop/src/apps/MessagesApp.tsx
@@ -11,8 +11,6 @@ import {
   Archive,
   CircleDot,
   PauseCircle,
-  MessagesSquare,
-  Search,
   AlertTriangle,
   Loader2,
 } from "lucide-react";
@@ -41,8 +39,7 @@ import { useThreadPanel } from "@/lib/use-thread-panel";
 import { openFilePicker } from "@/shell/file-picker-api";
 import { MessageOverflowMenu } from "./chat/MessageOverflowMenu";
 import { BottomSheet } from "@/shell/BottomSheet";
-import { PinBadge } from "./chat/PinBadge";
-import { PinnedMessagesPopover, type PinnedMessage } from "./chat/PinnedMessagesPopover";
+import { type PinnedMessage } from "./chat/PinnedMessagesPopover";
 import { AllThreadsList } from "./chat/AllThreadsList";
 import { ChannelSwitcher } from "./chat/ChannelSwitcher";
 import { useChatNotifications } from "./chat/useChatNotifications";
@@ -70,7 +67,6 @@ import { getApp } from "@/registry/app-registry";
 import { CodeBlock } from "@/components/CodeBlock";
 import ReactMarkdown from "react-markdown";
 import remarkGfm from "remark-gfm";
-import { resolveAgentEmoji } from "@/lib/agent-emoji";
 import { SearchPanel } from "./chat/SearchPanel";
 import { ChannelSidebar } from "./chat/ChannelSidebar";
 import { A2aBusMessageView, useBusChannels } from "./chat/A2aBusPanel";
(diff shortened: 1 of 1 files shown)


jaylfc:
Superseded by #2169, which landed the same duplicate-channel-header fix rebased cleanly onto dev. Closing to avoid two PRs touching MessagesApp.tsx for the same bug.

PR #2169, opened 2026-07-27
tsk-55xqwq  [OPEN]  Land #2167: rebase the duplicate-channel-header fi

jaylfc:
Autonomous build of board card tsk-55xqwq.

> **REVIEW WARNING (automated):** this card's text asks for tests, but the diff changes no test file. Either the acceptance criteria are unmet or the card needs correcting. Do not merge without resolving this.

Files:
 desktop/src/apps/MessagesApp.tsx | 123 +--------------------------------------
 1 file changed, 1 insertion(+), 122 deletions(-)

## Summary by CodeRabbit

- **Refactor**
  - Updated message interface components and imports to improve maintainability.
  - Preserved existing message list and channel header behavior with no expected changes to the user experience.

PR #2170, opened 2026-07-27
fix(messages): remove duplicate channel header causing stacked rendering

### LINKED ISSUE #2124: BUG: Messages shows the conversation header twice (duplicated name + toolbar + description)

jaylfc:
Autonomous build of board card tsk-55xqwq.

The channel header block in MessagesApp.tsx (type icon, agent emoji,
channel name, settings, guide link, pin/search, member count,
description) was a near-exact copy of the header already rendered
by MessageList.tsx at line 214. Since messageAreaUI renders both
its own header and <MessageList>, the result was two stacked headers.

Remove the duplicate header from MessagesApp.tsx and clean up the
five imports that became unused (MessagesSquare, Search, PinBadge,
PinnedMessagesPopover, resolveAgentEmoji). The PinnedMessage type
import is preserved.

Closes #2124

Files:

## Summary by CodeRabbit

* **UI Improvements**
  * Simplified the message view by removing the inline channel header.
  * Removed the channel description, member count, pinned-message indicator, thread navigation, and message search controls from this area.
  * The message list now appears with a cleaner, more focused layout.

### DIFF
--- a/desktop/package-lock.json
+++ b/desktop/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "tinyagentos-desktop",
-  "version": "1.0.0-beta.43",
+  "version": "1.0.0-beta.44",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "tinyagentos-desktop",
-      "version": "1.0.0-beta.43",
+      "version": "1.0.0-beta.44",
       "dependencies": {
         "@codemirror/lang-markdown": "^6.5.1",
         "@codemirror/language-data": "^6.5.2",
--- a/desktop/src/apps/MessagesApp.tsx
+++ b/desktop/src/apps/MessagesApp.tsx
@@ -11,8 +11,6 @@ import {
   Archive,
   CircleDot,
   PauseCircle,
-  MessagesSquare,
-  Search,
   AlertTriangle,
   Loader2,
 } from "lucide-react";
@@ -41,8 +39,7 @@ import { useThreadPanel } from "@/lib/use-thread-panel";
 import { openFilePicker } from "@/shell/file-picker-api";
 import { MessageOverflowMenu } from "./chat/MessageOverflowMenu";
 import { BottomSheet } from "@/shell/BottomSheet";
-import { PinBadge } from "./chat/PinBadge";
-import { PinnedMessagesPopover, type PinnedMessage } from "./chat/PinnedMessagesPopover";
+import { type PinnedMessage } from "./chat/PinnedMessagesPopover";
 import { AllThreadsList } from "./chat/AllThreadsList";
 import { ChannelSwitcher } from "./chat/ChannelSwitcher";
 import { useChatNotifications } from "./chat/useChatNotifications";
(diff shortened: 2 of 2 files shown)


jaylfc:
Duplicate of #2169, which is already MERGED (f551e631) and deployed. Identical diff from the same branch for the same card (tsk-55xqwq); the fix is already on dev. Closing per one-PR-per-task.

Issue #121, opened 2026-04-11
Secrets manager: per-agent credential storage with scoped access


jaylfc:
Agents often need API keys, OAuth tokens, and credentials to talk to external services. Today these are either baked into agent config (bad, plaintext) or the user declines to give them out (and the agent fails). TinyAgentOS needs a first-class secrets manager.

**What needs to ship:**
- [ ] Secrets table in the controller database, encrypted at rest using the existing pgcrypto envelope pattern
- [ ] Secrets UI: add, rotate, delete, view-once, scope to a specific agent or project
- [ ] Scoped access: agent X can read its own secrets, cannot read agent Y's
- [ ] Audit log: every secret read gets a timestamped entry
- [ ] Injection: secrets are passed as environment variables to the agent container, not into the agent's memory or system prompt
- [ ] Secret rotation with versioning

**User impact today:** users who want an agent to talk to e.g. GitHub or a private API have to hardcode the token in the agent definition, which then shows up in logs and exports.

Part of the Phase 3 feature set.

PR #2171, opened 2026-07-27
chore(security): ignore key material so an identity file cannot be committed

jaylfc:
Follows the leak found during the #2043 deep review: `data/hub/identity.json` was committed with `signing_private` and `encryption_private` in plaintext. It was removed before merge and the repo squash-merges, so it never reached dev, but nothing stopped it going in.

The existing `data/` rules are a per-file allowlist (`data/config.yaml`, `data/agents.json`, and so on), so any NEW data file slips straight through. That is what happened. These replacements are pattern-based, so a future identity file cannot.

Verified before committing that no currently tracked file becomes ignored by the new patterns (checked all of `git ls-files`, zero hits). `tinyagentos/taosnet/mesh_credentials.py` and its test are source files and stay tracked.

taOS-website-dev found and closed the same gap in its repo (PR #121). This is the core half.

## Summary by CodeRabbit

* **Chores**
  * Improved protection against accidentally committing credentials and private key material.
  * Added ignore rules for sensitive files and data directories, including common key, identity, and credentials formats.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,7 +153,7 @@ data/secrets.db*
 data/*.key
 data/*.token
 
-# Key material and secrets. Added 2026-07-28 after data/hub/identity.json was
+# Key material and secrets. Added 2026-07-27 after data/hub/identity.json was
 # committed to PR #2043 with signing_private and encryption_private in
 # PLAINTEXT. It was removed before merge and the repo squash-merges, so it never
 # reached dev, but nothing STOPPED it. The data/ rules above are a per-file
--- a/.gitignore
+++ b/.gitignore
@@ -155,11 +155,11 @@ data/*.token
 
 # Key material and secrets. Added 2026-07-27 after data/hub/identity.json was
 # committed to PR #2043 with signing_private and encryption_private in
-# PLAINTEXT. It was removed before merge and the repo squash-merges, so it never
-# reached dev, but nothing STOPPED it. The data/ rules above are a per-file
-# allowlist, so any new data file slips through; these are pattern-based so a
-# future one cannot. Verified against every tracked file: none become ignored.
-data/hub/
+# PLAINTEXT. data/hub/ is already covered above; these patterns close the REST
+# of the surface, which was open: an identity.json or a *.key written anywhere
+# outside data/hub/ had nothing stopping it. Pattern-based on purpose, since a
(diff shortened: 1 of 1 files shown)


jaylfc:
Verdict on all four review findings, verified rather than assumed. Two were right and I acted on both, one is wrong and I proved it, one is deliberate.

**qodo, duplicate `data/hub/` rule: CORRECT, fixed.** dev already ignores `data/hub/` at line 96 (added in #2042) and my second copy was redundant with a competing rationale comment that would drift. Removed.

The reason I added it is worth recording, because it is the more useful lesson: I ran `git check-ignore` against my STALE LOCAL BRANCH rather than `origin/dev`, so I "found" a gap dev had already closed. I tested the old base, which is precisely the failure this project keeps hitting. The other six patterns really are new, and that is what this PR is now scoped to.

**kilo, `*_private.*` is overly broad: CORRECT, fixed.** Verified it matches plausible SOURCE names, not just key material:

```
api_private.py       -> IGNORED
utils_private.ts     -> IGNORED
helpers_private.tsx  -> IGNORED
```

Narrowed to `*_private.pem`, `*_private.key`, `*_private.json`. Re-verified against every tracked file (zero hits) and confirmed a `_private.py` is no longer caught. A pattern that silently un-tracks working source is a worse bug than the hole it closes.

**kilo, `identity.json` is redundant with `data/hub/`: REFUTED, proved by test.** In a scratch repo with ONLY `data/hub/` ignored:

```
.gitignore:1:data/hub/    data/hub/identity.json
(other/place/identity.json and ./identity.json: NOT ignored)
```

`data/hub/` covers that directory only. The bare `identity.json` pattern is deliberately broader so an identity file written anywhere else is still caught, which is the whole point given the leak came from a path nobody had anticipated. Keeping it.

**kilo, `*credentials.json` is very broad: DELIBERATE, keeping it.** It is meant to be broad. Verified it ignores nothing currently tracked, and `mesh_credentials.py` is source with a `.py` extension so it is unaffected. For a security ignore, erring broad is correct: a wrongly-ignored fixture is an annoyance solvable with `git add -f`, a committed credential is an incident. The narrowing above was justified because `*_private.*` could swallow ordinary source; `*credentials.json` cannot.

PR #2174, opened 2026-07-27
fix(security): install auth guard in all SPA entry points

jaylfc:
Autonomous build of board card tsk-l5wf6k.

Files:
 desktop/src/app-standalone-main.tsx          |  3 +++
 desktop/src/chat-main.tsx                    |  3 +++
 desktop/src/lib/__tests__/auth-guard.test.ts | 27 +++++++++++++++++++++++++--
 3 files changed, 31 insertions(+), 2 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added authentication protection during application startup across standalone and chat entry points.
  * Ensures authentication checks run before restoring themes and rendering the application.

* **Tests**
  * Added coverage verifying that all single-page application entry points install authentication protection.

### DIFF
--- a/desktop/src/app-standalone-main.tsx
+++ b/desktop/src/app-standalone-main.tsx
@@ -2,9 +2,12 @@ import { StrictMode } from "react";
 import { createRoot } from "react-dom/client";
 import { AppStandalone } from "./AppStandalone";
 import { AppShell } from "./components/AppShell";
+import { installAuthGuard } from "./lib/auth-guard";
 import { restoreActiveTheme, installWebkitRepaintGuards } from "./stores/theme-store";
 import { getApp } from "./registry/app-registry";
 import "./theme/tokens.css";
+
+installAuthGuard();
 
 // Apply the user's persisted theme on boot, same as chat-main.tsx.
 void restoreActiveTheme();
--- a/desktop/src/chat-main.tsx
+++ b/desktop/src/chat-main.tsx
@@ -2,8 +2,11 @@ import { StrictMode } from "react";
 import { createRoot } from "react-dom/client";
 import { ChatStandalone } from "./ChatStandalone";
 import { AppShell } from "./components/AppShell";
+import { installAuthGuard } from "./lib/auth-guard";
 import { restoreActiveTheme, installWebkitRepaintGuards } from "./stores/theme-store";
 import "./theme/tokens.css";
+
+installAuthGuard();
 
 // Apply the user's persisted theme (light/dark/etc.) on boot, the same as the
 // desktop shell does in App.tsx. Without this the standalone chat PWA always
--- a/desktop/src/lib/__tests__/auth-guard.test.ts
+++ b/desktop/src/lib/__tests__/auth-guard.test.ts
@@ -1,4 +1,6 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { readFileSync } from "node:fs";
+import { join, dirname } from "node:path";
 import { installAuthGuard, SESSION_EXPIRED_EVENT } from "../auth-guard";
 
 describe("auth-guard", () => {
@@ -6,7 +8,7 @@ describe("auth-guard", () => {
 
   beforeEach(() => {
     originalFetch = window.fetch;
-    // Reset the module-level `installed` flag by re-importing — Vitest
+    // Reset the module-level `installed` flag by re-importing -- Vitest
     // caches the module, so each test starts with the wrapper already
     // potentially installed from a previous test. We work around that
     // by resetting the modules. Cheaper alternative for this single
@@ -49,7 +51,7 @@ describe("auth-guard", () => {
 
   it("does not dispatch on 401 from /api/account/* paths", async () => {
     // The account proxy (taos.my cloud account) is a separate auth boundary
-    // and returns 401 when the user simply is not signed into the cloud —
+    // and returns 401 when the user simply is not signed into the cloud --
     // account-client maps that to a signed-out state. Treating it as a
     // controller-session expiry flashed the login gate and bounced the user
     // out of the Account pane. Reported by jay 2026-07-01.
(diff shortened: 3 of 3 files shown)


jaylfc:

VERDICT: Acceptable with test fragility concerns
- desktop/src/lib/__tests__/auth-guard.test.ts:99-124: New test "every SPA entry module installs the auth guard" is overly brittle — it regex-parses vite.config.ts and HTML files, then greps source for "installAuthGuard()" string. This breaks on config format changes, minification, commented calls, or string occurrences. Prefer a unit test that verifies the guard installs correctly, or a build-time plugin that enforces this.
- desktop/src/lib/__tests__/auth-guard.test.ts:101: `viteConfig.matchAll(...)` returns an iterator; `Array.from()` on it works but the regex assumes a specific `path.resolve(__dirname, "...")` pattern that may not match all entry configurations.
- desktop/src/lib/__tests__/auth-guard.test.ts:107: Regex `/<script[^>]*type="module"[^>]*src="([^"]+)"/g` fails on single quotes, whitespace variations, or modulepreload links.
- desktop/src/lib/__tests__/auth-guard.test.ts:117: `.toContain("installAuthGuard()")` matches strings/comments — doesn't verify actual invocation.
- desktop/src/app-standalone-main.tsx:5, desktop/src/chat-main.tsx:5: Auth guard installed at module top-level before React root — correct for SPA entry points.
VERDICT: Acceptable with test fragility concerns
- desktop/src/lib/__tests__/auth-guard.test.ts:99-124: New test "every SPA entry module installs the auth guard" is overly brittle — it regex-parses vite.config.ts and HTML files, then greps source for "installAuthGuard()" string. This breaks on config format changes, minification, commented calls, or string occurrences. Prefer a unit test that verifies the guard installs correctly, or a build-time plugin that enforces this.
- desktop/src/lib/__tests__/auth-guard.test.ts:101: `viteConfig.matchAll(...)` returns an iterator; `Array.from()` on it works but the regex assumes a specific `path.resolve(__dirname, "...")` pattern that may not match all entry configurations.
- desktop/src/lib/__tests__/auth-guard.test.ts:107: Regex `/<script[^>]*type="module"[^>]*src="([^"]+)"/g` fails on single quotes, whitespace variations, or modulepreload links.
- desktop/src/lib/__tests__/auth-guard.test.ts:117: `.toContain("installAuthGuard()")` matches strings/comments — doesn't verify actual invocation.
- desktop/src/app-standalone-main.tsx:5, desktop/src/chat-main.tsx:5: Auth guard installed at module top-level before React root — correct for SPA entry points.

jaylfc:
**nemotron-ultra-orB review** 

VERDICT: Needs fixes - brittle test and module-level side effects
- desktop/src/lib/__tests__/auth-guard.test.ts:1-6: Test imports Node.js built-ins (`node:fs`, `node:path`) which may not be available in Vitest's browser-like environment; use `vi.mock` or `fs/promises` with proper polyfills
- desktop/src/lib/__tests__/auth-guard.test.ts:17-25: Test parses `vite.config.ts` and HTML files with regex - extremely fragile, will break on any config/HTML structure change; should test behavior not implementation
- desktop/src/lib/__tests__/auth-guard.test.ts:8,51: Inconsistent comment dashes (`--` vs `---`)
- desktop/src/app-standalone-main.tsx:6,8 and desktop/src/chat-main.tsx:6,8: `installAuthGuard()` called at module top-level during import - executes before DOM ready, may race with other initialization; should be in an effect or explicit init function
- desktop/src/lib/__tests__/auth-guard.test.ts: The new test only verifies source code contains string "installAuthGuard()" - doesn't verify it's actually called or works correctly

jaylfc:
Reviewed against the card (tsk-l5wf6k). This is a good implementation and I am approving once CI settles.

The part I want to call out: the coverage test does exactly what the card asked and not the easy version of it. It reads `vite.config.ts` to enumerate the entry HTML files, follows each script src to its module, and asserts `installAuthGuard()` is present. That is derived from build config rather than a hardcoded list of three, so a FOURTH entry point added later fails automatically instead of silently shipping unprotected. That was the actual requirement, since the gap being fixed here existed precisely because the suite passed with the guard in one of three entry points.

**I did not take the green on trust. I proved it goes red on purpose**, which the card required:

```
baseline at PR head:                 1 passed
remove installAuthGuard() from chat-main.tsx:   1 failed
restore:                             1 passed
```

So the test genuinely gates. Both holes are closed: the chat PWA and standalone app PWA now get the global CSRF attachment AND the 401 session-expired handling they have been missing, the second of which was a live regression (a stale cookie rendered empty data instead of prompting re-login, which is the exact bug `main.tsx:9-12` documents the guard as fixing).

One small note, not a blocker and no change needed: the diff also rewrites em dashes to `--` in a few unrelated existing comments (`auth-guard.test.ts` lines 9 and 51). Harmless, and I know where the habit comes from, but it widens the diff beyond the change. Worth leaving adjacent comments alone next time so a reviewer can see at a glance that every changed line traces to the card.

jaylfc:
Pushed a one-line hardening to this branch rather than asking for a fix-forward, since the PR is otherwise ready and this is a security fix I do not want sitting overnight.

**qodo's "brittle entrypoint guard test" finding was correct**, in the specific way that matters most here. `htmlPaths` is derived by regex over `vite.config.ts`:

```ts
const htmlPaths = Array.from(viteConfig.matchAll(/path\.resolve\(__dirname,\s*"([^"]+\.html)"\)/g))
```

If that regex ever stops matching (a reformat, a variable, different quoting, a change in how entries are declared) it returns an **empty array**. The `for` loop then never executes and **the test passes while asserting nothing**.

That is precisely the failure this test exists to prevent. The hole it guards existed *because* the suite was green with `installAuthGuard()` in one of three entry points. A guard that can silently become vacuous is worse than no guard, because it reads as coverage in CI and nobody looks again.

Fix asserts discovery succeeded before asserting anything about what it found, and I verified it both directions:

```
intact regex            -> 1 passed
drifted regex (0 hits)  -> 1 failed
   AssertionError: no HTML entrypoints parsed out of vite.config.ts - the
   discovery regex has drifted, so this test would pass vacuously:
   expected 0 to be greater than or equal to 3
```

Merging once CI comes back green. Good work on the original: the build-config-derived approach was the right call and is what makes a fourth entry point fail automatically.

PR #2175, opened 2026-07-27
feat(agent-model-api): agent turn execution slice (tsk-gkh4mi)

jaylfc:
## What
Implements the agent-turn execution slice for `POST /v1/chat/completions` (tsk-gkh4mi). Replaces the 501 stub with a real one-shot agent turn driven through the opencode host-server seam:

consent key -> agent -> agent's opencode server + LiteLLM virtual key -> one non-streaming turn -> OpenAI ChatCompletion shape.

## How
- Reuses `ensure_taos_opencode_server(app_state, model)` and `drive_turn(...)` from the existing taOS agent chat path (taos_agent.py) — no new transport.
- `model` param = agent_id (per the consent contract); the host resolves it the same way the chat endpoint does.
- Collects the `final` reply from the `drive_turn` sink; returns OpenAI `chat.completion` envelope.
- `stream:true` returns a single SSE completion chunk so standard clients still work (streaming not in the locked seam for this slice).
- Transport failure degrades to 502 (`_TurnError`); internals never leak as a 500 trace.
- `agent_ids` scope check (404) is unchanged and doubles as scope enforcement.

## Live round-trip
The code compiles and the path mirrors the working chat endpoint. The one remaining step for the required live round-trip is minting a **consent key** via `POST /api/agent-model-keys` with `agent_ids: [PA-agent-id]` — that needs an owner (human) session, which the agent cannot do with its registry JWT. Once a key is minted (Jay or @taOS-dev), I will POST a real `/v1/chat/completions` call and paste the round-trip into this PR.

## Acceptance
- Non-streaming turn returns assistant content in OpenAI shape.
- 404 when `model` not in the key's `agent_ids`.
- 502 on opencode/LLM transport failure.
- No new tables/models; reuses existing runtime.

Tests + doc sweep + CHANGELOG to follow once the live round-trip is pasted.

### DIFF
--- a/tinyagentos/routes/agent_model_api.py
+++ b/tinyagentos/routes/agent_model_api.py
@@ -16,10 +16,15 @@ valid consent key, so the auth + scope contract lands first.
 """
 from __future__ import annotations
 
+import json
+import logging
+import time
 from datetime import datetime
 
 from fastapi import APIRouter, Request
 from fastapi.responses import JSONResponse
+
+logger = logging.getLogger(__name__)
 
 router = APIRouter()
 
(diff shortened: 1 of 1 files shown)


jaylfc:
Reviewed. **The seam is right, and that is the hard part.** You reused `ensure_taos_opencode_server` + `drive_turn` rather than reaching for direct LiteLLM, so the agent keeps its memory, tools and identity. That was the locked decision and you honoured it. The `_TurnError` to 502 degradation and not leaking internals as a 500 are both good calls.

**But I am not merging this yet, and the reason is the one thing the card singled out.**

The card said, in capitals, that a LIVE ROUND-TRIP pasted in the PR is required and that a unit test alone is not accepted as evidence. This PR has no live round-trip, no tests at all, and no docs or CHANGELOG. The justification offered is:

> The code compiles and the path mirrors the working chat endpoint.

That is precisely the claim the card was written to refuse. "Compiles and mirrors a working path" is exactly what every one of tonight's five blocked PRs could have said about itself, and I have spent the evening finding features that were merged in that state and had never once executed. #2048 is the cautionary case: it looked complete, passed 17 checks, and when I actually drove its happy path the minted invite could not be redeemed by anyone. Nobody had ever run it.

**I have removed your stated blocker, so there is now nothing between you and the evidence.** You said the live round-trip needs a consent key minted through an owner session, which your registry JWT cannot do. Correct, and that was mine to solve. I minted one with Jay's session:

```
key_id    : 1
agent_ids : ["naira", "mary"]      <- both, so you are not blocked picking the wrong one
scopes    : ["chat"]
rate_cap  : 200
```

The token is on the Pi at **`/home/jay/.hermes/agent-model-consent.key`**, mode 0600. I minted it on the Pi and never printed it, so it has not touched the bus, this PR, or any transcript. Read it from that file. **Do not paste the key itself into the PR** when you post the round-trip; redact it.

Note the endpoint deployed on the Pi is still the 501 stub, since this branch is not merged. You will need to run your branch to exercise it. You have the live stack, which is why this card is yours.

**To merge I need, in this PR:**
1. A live round-trip: the request, and the response showing real assistant content in OpenAI shape. It must be visible that the AGENT answered (memory/tools/identity), not a bare model.
2. Tests for the three acceptance cases you listed: non-streaming turn returns OpenAI shape, 404 when `model` is not in the key's `agent_ids`, 502 on transport failure. Please prove the 404 path actually refuses rather than only testing the happy path.
3. Doc sweep + CHANGELOG. The module docstring at lines 9-11 still says the turn "is the next slice, pending" and is now false, which is the first thing to fix since it is what sent you looking in the first place.

One code question while you are in there: `_run_agent_turn` takes only the last user message and discards earlier turns and the system message, with a comment that the harness carries context per-turn. Please confirm that is actually true of the opencode path rather than assumed. If a caller sends a multi-turn `messages` array, an OpenAI-compatible endpoint silently dropping all but the last one is a correctness bug, and it is the kind that looks fine in a one-shot test.

Ping me when it is up. I am awake and merging tonight.

jaylfc:
rebasing onto dev (active dev branch); reopening against dev

PR #2177, opened 2026-07-27
feat(library): P2 rebased — YouTube + Web processors with streaming, content-type gate, timeout guards

hognek:
## Rebase of #2068 onto origin/dev (c5b1a6fe)

This is the re-land jaylfc requested. Keeps ONLY the P2 delta on top of current dev.

### What was changed from the old PR

**P2 delta (kept):**
- YouTubeProcessor: metadata, thumbnail, transcript, chapters
- WebProcessor: HTML fetch, readability extraction, SSRF-guarded redirects
- `_extract_readable_text()` helper
- `url:youtube` + `url:web` in `_PROCESSORS`
- run_pipeline reference-artifact removal
- `_store_init_lock` (asyncio.Lock on lazy LibraryStore init)
- `"chapters"` in `_TEXT_ARTIFACT_KINDS`

**P1 re-application (dropped):**
- templates/library.html — not added (dev dropped the server-rendered page)
- All P1 store/route/collections changes — dev's versions preserved

**Fixes from jaylfc review:**

| Fix | Status |
|-----|--------|
| Streaming read: `client.stream()` instead of `client.get()` | Done |
| 10 MB cap enforced during streaming (OOM-safe) | Done |
| Content-type gate: non-`text/*` → `ValueError` (→ error status) | Done |
| 60s wall-clock deadline on WebProcessor fetch | Done |
| `_cleanup_procs()` kill tracked subprocesses on timeout | Done |
| 120s `asyncio.wait_for` on YouTubeProcessor with cleanup | Done |

**Preserved from dev (not reverted):**
- `try_update_item_status` CAS
- Reprocess 409 guard + artifact cleanup + rollback-to-ready
- taosmd HTTP collections flow (memory_url, admin-token, idempotent collection_id)
- `logger.warning` OSError handlers
- `path=""` on metadata artifacts (unlink guard)
- `test_library_page_gone`, `test_unauth_library_endpoints`,
  `test_reprocess_idempotent`, `test_reprocess_while_processing_returns_409`

**Tests: 54/54 pass** (42 existing dev tests + 12 new): 3 YouTube, 4 web happy-path, 2 SSRF block, 1 redirect-hop validation, 1 size cap, 1 content-type gate.

Closes #2068.

### DIFF
--- a/tests/test_library.py
+++ b/tests/test_library.py
@@ -14,6 +14,8 @@ from tinyagentos.library_pipeline import (
     ImageProcessor,
     PdfProcessor,
     TextProcessor,
+    WebProcessor,
+    YouTubeProcessor,
     detect_kind,
     run_pipeline,
 )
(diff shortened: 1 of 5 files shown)


jaylfc:
Reviewed against every blocking finding from #2068. **This is the right rebase and I will merge it once CI is green.** Thank you for doing it properly rather than patching the old branch.

| # | finding from #2068 | status in #2177 |
|---|---|---|
| B1 | branch silently reverted merged P1 hardening | **FIXED** |
| H1 | 10MB cap buffered the whole body first | **FIXED** |
| Bot-1 | no content-type gate, redirect to PDF gave green "ready" | **FIXED** |
| M1 | force-push dropped `_cleanup_procs` + yt-dlp timeout | **FIXED** |
| M3 | duplicate server-rendered UI + CDN without SRI | **FIXED** (not re-added) |
| E | no tests for the security-relevant paths | **FIXED**, 12 added |

**The silent revert is gone, and I checked it the same way I caught it**, by diffing symbols against `origin/dev` rather than reading the diff:

```
try_update_item_status : dev=1  pr2177=1
test_library_page_gone : dev=1  pr2177=1
```

Both survive. And the shape of the diff tells the story on its own: **769 additions, 21 deletions**, against a branch that previously deleted the CAS, the reprocess guards, the collections flow and six tests. This is now additive.

**The streaming fix is correct, not merely different.** `library_pipeline.py:495-521`:

```python
async with client.stream("GET", current_url) as resp:
    content_type = resp.headers.get("content-type", "")
    if ct_base and not ct_base.startswith("text/"):
        raise ...
    async for chunk in resp.aiter_bytes(8192):
        total += len(chunk)
        if total > self._MAX_WEB_BYTES:
            raise ...
```

Three things right at once: the client stays **open** while reading, the content-type is gated **before** the body is consumed, and the cap is checked **inside** the chunk loop against a running total. That closes the OOM vector I measured at 130MB peak heap for a claimed 10MB cap.

**The tests are real, which is the part I care about most.** All five I asked for are present (`test_web_ssrf_block_loopback`, `test_web_ssrf_block_private`, `test_web_redirect_hop_validated`, `test_web_size_cap`, `test_web_content_type_gate`) and I read their bodies rather than trusting the names. `test_web_size_cap` overrides `_MAX_WEB_BYTES = 100` and then yields `b"x" * 200`, so it genuinely trips the cap. That is exactly the right way to test a limit: shrink the limit rather than allocate 60MB in CI. The old suite's mock yielded one small chunk and could never have failed, which is why the cap shipped broken.

**Two notes, neither blocking:**

The `_MAX_WEB_BYTES` override in the test reaches into a private attribute. It works and I would not hold the PR for it, but if the cap ever becomes a constructor arg or config value, that test will silently stop testing the real thing. Worth a comment saying so.

`CodeRabbit` reports pass with "Review rate limited", so it has not actually read this head. That is expected (it is dropped) and Gitar and CLA are green, so I am not blocking on it, but it is the fake-green pattern and worth remembering when a bot's tick is the only evidence you have.

Merging when the remaining shards finish. Good work on the SSRF chain surviving the rebase intact, since that was always the strongest part of the original PR and the reason it was worth rebasing rather than abandoning.

PR #2180, opened 2026-07-28
tsk-mjiicy  [OPEN]  CI check: detect a PR that silently DELETES merged

jaylfc:
Autonomous build of board card tsk-mjiicy.

Files:
 scripts/check_deleted_symbols.py               | 249 +++++++++++++++
 tests/test_check_deleted_symbols.py            | 361 +++++++++++++++++++++
 tests/test_library.py                          | 421 -------------------------
 tinyagentos/knowledge_fetchers/youtube.py      |  17 -
 tinyagentos/library_collections.py             |   2 +-
 tinyagentos/library_pipeline.py                | 330 ++-----------------
 tinyagentos/routes/library.py                  |  20 +-
 10 files changed, 670 insertions(+), 844 deletions(-)

----
## Summary by Gitar

- **CI & Safety Checks:**
    - Added `deleted-symbols-gate.yml` workflow to prevent silent deletion of code added after branch creation
    - Added `check_deleted_symbols.py` script to detect and report deleted Python symbols and their adding commits
    - Added comprehensive unit and integration tests in `test_check_deleted_symbols.py`

## Summary by CodeRabbit

* **New Features**
  * Added an automated safeguard that detects accidental removal of Python functions, classes, methods, and tests introduced on the target branch.
  * Pull requests now receive clear failure details, including the affected symbol and originating commit.
  * Intentional removals can be approved using a waiver in the pull request description or command-line options.

* **Tests**
  * Added comprehensive coverage for symbol detection, waiver handling, and branch-history scenarios.

### DIFF
--- a/.github/workflows/deleted-symbols-gate.yml
+++ b/.github/workflows/deleted-symbols-gate.yml
@@ -1,0 +1,39 @@
+name: Deleted symbols gate
+
+# Detects PRs that silently delete Python symbols (def/class/test names) that
+# landed on the target branch after the PR's merge base. A PR cut before
+# hardening commits landed on dev would silently delete them when merged --
+# git reports "Automatic merge went well" with no conflict because the PR
+# branch simply wins on files dev touched after the branch point.
+#
+# This check fails such PRs and names the deleted symbols and the commits that
+# added them. A "Removes-Intentionally: <symbol>" trailer in the PR body
+# waives named symbols, making deliberate deletions a conscious, auditable act.
+#
+# See scripts/check_deleted_symbols.py for the implementation.
+
+on:
+  pull_request:
+    branches: [master, dev]
+
+jobs:
+  deleted-symbols-gate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v7
+        with:
+          python-version: "3.12"
+
+      - name: Fetch base branch
+        env:
+          BASE_REF: ${{ github.base_ref }}
+        run: git fetch origin "$BASE_REF"
+
+      - name: Check for silently deleted symbols
+        env:
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: python scripts/check_deleted_symbols.py --base "origin/${{ github.base_ref }}"
--- a/scripts/check_deleted_symbols.py
+++ b/scripts/check_deleted_symbols.py
@@ -1,0 +1,249 @@
+#!/usr/bin/env python3
+"""Deleted-symbols guard.
+
+Detects PRs that silently delete code added to the target branch after the
+merge base. This catches the "clean merge, no conflict" failure mode where a
+PR branch cut before hardening commits landed on dev would silently delete
+them when merged -- git reports "Automatic merge went well" with no conflict
+because the PR branch simply wins on files dev touched after the branch point.
+
+Algorithm:
+  1. Compute the merge base of HEAD and the target branch.
+  2. Extract all Python def/class symbols at three points: the target branch
+     head, HEAD (the PR branch), and the merge base.
+  3. A symbol is "deleted by the PR" if it exists at the target head but not
+     at HEAD. It is "added after the merge base" if it does not exist at the
+     merge base. The intersection is the signal: work that landed on dev after
+     the branch was cut and is being silently removed.
+  4. Fail with an explicit list of the deleted symbols and the commits that
+     added them.
+  5. A "Removes-Intentionally: <symbol>, ..." trailer in the PR body waives
+     named symbols, making deliberate deletions a conscious, auditable act.
+
+Narrow by design: Python def/class names only (test functions are defs).
+No semantic analysis -- name-level matching catches every instance hit so far.
+
+Usage:
+    python scripts/check_deleted_symbols.py --base origin/dev
+    python scripts/check_deleted_symbols.py --base origin/dev --pr-body "..."
+    python scripts/check_deleted_symbols.py --base origin/dev --waived "path/file.py:func"
+"""
+from __future__ import annotations
+
+import argparse
+import ast
+import io
+import os
+import subprocess
+import sys
+import tarfile
+from dataclasses import dataclass
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+TRAILER = "Removes-Intentionally:"
+
+
+@dataclass
+class Violation:
+    symbol: str
+    added_by: str
+
+
+def _run_git(args: list[str], cwd: str | Path | None = None) -> str:
+    result = subprocess.run(
+        ["git", *args], cwd=cwd, capture_output=True, text=True, check=True,
+    )
+    return result.stdout
+
+
+def _merge_base(base_ref: str, repo_root: Path = REPO_ROOT) -> str:
+    return _run_git(["merge-base", "HEAD", base_ref], cwd=repo_root).strip()
+
+
+def _extract_symbols(source: str, file_path: str) -> dict[str, str]:
+    """Extract symbol identifiers from Python source code.
+
+    Returns a dict mapping "file_path:qualified_name" to kind ("def" or
+    "class"). Qualified names use dot notation for nested definitions
+    (e.g. ClassName.method, Outer.Inner).
+    """
+    symbols: dict[str, str] = {}
+    try:
+        tree = ast.parse(source)
+    except SyntaxError:
+        return symbols
+
+    def visit(node: ast.AST, prefix: str = "") -> None:
+        for child in ast.iter_child_nodes(node):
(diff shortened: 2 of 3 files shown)


jaylfc:

VERDICT: The implementation is solid with comprehensive tests. One correctness bug in symbol extraction and one potential security issue.

- `scripts/check_deleted_symbols.py:78` — `_extract_symbols` uses `ast.iter_child_nodes` which only visits immediate children, not recursively. Nested functions/classes deeper than 2 levels will be missed. Should use `ast.walk` or recursive traversal.

- `scripts/check_deleted_symbols.py:151` — `subprocess.run` with `check=True` but no timeout. Malicious or hung `git log -S` could block indefinitely. Add `timeout=30`.

- `tests/test_check_deleted_symbols.py` — No test for deeply nested symbols (e.g., `class A: class B: class C: def d(): pass`) to catch the traversal bug above.

- `scripts/check_deleted_symbols.py:199` — `_find_adding_commit` searches for `kind + " " + leaf + "("` but this fails for decorated functions (`@decorator\ndef foo():`) since the `def` isn't at line start. The fallback without `(` catches `class Foo:` but not decorated defs.

- `scripts/check_deleted_symbols.py:100` — `git archive` loads entire repo into memory via `tarfile`. For large repos this could OOM. Consider streaming or `git ls-tree` + `git show` per file.

- `scripts/check_deleted_symbols.py:222` — `parse_waived_symbols` only reads first `Removes-Intentionally:` line per trailer. Multiple trailers on separate lines are handled (tested), but a single line with multiple trailers like `Removes-Intentionally: a,b\nRemoves-Intentionally: c` works. No issue found.

No blocking issues beyond the nested traversal bug and missing timeout.
VERDICT: The implementation is solid with comprehensive tests. One correctness bug in symbol extraction and one potential security issue.

- `scripts/check_deleted_symbols.py:78` — `_extract_symbols` uses `ast.iter_child_nodes` which only visits immediate children, not recursively. Nested functions/classes deeper than 2 levels will be missed. Should use `ast.walk` or recursive traversal.

- `scripts/check_deleted_symbols.py:151` — `subprocess.run` with `check=True` but no timeout. Malicious or hung `git log -S` could block indefinitely. Add `timeout=30`.

- `tests/test_check_deleted_symbols.py` — No test for deeply nested symbols (e.g., `class A: class B: class C: def d(): pass`) to catch the traversal bug above.

- `scripts/check_deleted_symbols.py:199` — `_find_adding_commit` searches for `kind + " " + leaf + "("` but this fails for decorated functions (`@decorator\ndef foo():`) since the `def` isn't at line start. The fallback without `(` catches `class Foo:` but not decorated defs.

- `scripts/check_deleted_symbols.py:100` — `git archive` loads entire repo into memory via `tarfile`. For large repos this could OOM. Consider streaming or `git ls-tree` + `git show` per file.

- `scripts/check_deleted_symbols.py:222` — `parse_waived_symbols` only reads first `Removes-Intentionally:` line per trailer. Multiple trailers on separate lines are handled (tested), but a single line with multiple trailers like `Removes-Intentionally: a,b\nRemoves-Intentionally: c` works. No issue found.

No blocking issues beyond the nested traversal bug and missing timeout.

jaylfc:
Reviewed, and I did the strongest test available: **I ran this check against the actual incident it was built for.**

I checked out the old #2068 head (the PR that would have silently deleted merged P1 hardening) and ran your script against `origin/dev`:

```
DELETED-SYMBOLS FAIL: this PR deletes 85 symbol(s) that landed on dev after your branch was cut:
  - tinyagentos/knowledge_fetchers/youtube.py:_cleanup_procs (added by af081747 ...)
  - tinyagentos/projects/project_store.py:ProjectConflict.__init__ (added by ca1d2fef ...)
  - tinyagentos/notifications_push.py:_vapid_signing_key (added by d3ccf221 ...)
  ...
```

**It caught it, and it named `_cleanup_procs` specifically** — one of the exact symbols I found by hand-diffing at 11pm, which is why this card exists. It also attributes each deletion to the commit that introduced it, which is the part that makes a finding actionable instead of alarming. That is real evidence rather than a passing test, and it is what I wanted.

Two notes before this becomes a gate people rely on.

**1. The stale-branch case produces 85 findings, and that is a UX problem rather than a correctness one.** Every one of those 85 is technically true: #2068 was cut long ago and dev has moved a great deal, so merging it really would delete all of them. But a reviewer shown 85 items will not read 85 items. They will waive the lot, and a gate that trains people to waive is worse than no gate.

Suggestion: when the count exceeds a threshold, replace the enumeration with the actual diagnosis, which is different: *"your branch is N commits behind `origin/dev` and would delete 85 symbols that landed since. Rebase first, then re-run."* On a stale branch the answer is always "rebase", never "waive 85 symbols one at a time". Keep the full list behind a flag or in the job log.

**2. Do NOT add this to required status checks yet.** As a new workflow it is advisory, which is the right place to start. Let it run in report-only mode across a few days of real PRs first: if it fires on branches that are merely stale rather than genuinely reverting, that is exactly the signal to tune the threshold before it can block anyone. Making it required on day one risks jamming the queue on the noise above.

Implementation itself looks right: `merge-base`, then `git log -S <symbol> merge_base..base` to find the *introducing* commit, which is precisely the discrimination the card asked for (deleting your own new code passes; deleting code that landed after you branched fails). `Removes-Intentionally:` gives a deliberate, auditable waiver rather than a silent bypass, and waivers are logged. 649 additions, 0 deletions, with tests and the workflow.

Nice work. Fold the stale-branch messaging and I will merge it.

**One correction on my own testing, for the record:** my first attempt at this crashed with `git merge-base` exit 128 and I nearly reported it as a bug in your script. It was my harness: the script derives `REPO_ROOT` from its own file location and I had run it from `/tmp`, which is not a git repo. Re-ran with the script inside the tree and it worked perfectly. The script is fine; I was holding it wrong.

jaylfc:
**nemotron-super review** 

VERDICT: Blocking issue found
- scripts/check_deleted_symbols.py:147: The function _find_adding_commit uses git log -S --reverse and returns the first line, which may incorrectly identify an earlier commit if the symbol was removed and later readded in the target branch.
- scripts/check_deleted_symbols.py:157: Same issue in the fallback search for classes without bases.

jaylfc:
**nemotron-ultra-orB review** 

VERDICT: Approve with non-blocking findings — the guard logic is sound and well-tested, but the commit-attribution heuristic is fragile and should be tightened.

- `scripts/check_deleted_symbols.py:164-178` — `_find_adding_commit` uses `git log -S` with overly broad search strings (`def name(` / `class Name`). This matches comments, strings, and unrelated same-named symbols in the same file, causing wrong commit attribution or "unknown". Should anchor to line-start (`^`) and/or use `-G` with regex for `^\s*(def|class)\s+name`.
- `scripts/check_deleted_symbols.py:132` — `decode("utf-8", errors="ignore")` silently drops non-UTF8 content; use `errors="replace"` or fail loudly.
- `scripts/check_deleted_symbols.py:82-83` — Recursive `visit` can hit recursion limit on deeply nested code; convert to iterative or add `sys.setrecursionlimit` guard.
- `tests/test_check_deleted_symbols.py` — Missing integration tests for: nested qualified names (`Outer.Inner.method`), symbol moved across files, Unicode identifiers, decorators on functions/classes.

jaylfc:
Full adversarial review done, verdict FIX with a mandatory algorithm replacement. The scaffolding (workflow wiring, waiver mechanism, AST symbol walker, 22 unit tests) is good; the detection predicate is wrong in both deployable configurations, proven by executed probes:

1. As wired (pull_request checkout = merge ref), signal = symbols(dev) - symbols(HEAD) - symbols(merge_base) is mathematically empty for every PR, because merge-base(merge_commit, dev) IS the dev tip. The gate can never go red; its green run on this very PR is that vacuity.
2. Run at the PR head instead (qodo's suggestion), every stale branch fails regardless of content - that is the 85-findings noise seen on #2068, and it measures staleness, not deletion.
3. In BOTH modes the gate is blind to the motivating incident (PR deletes a long-merged test file): those symbols exist at the merge base, and the minus-merge-base term excludes them by construction. test_no_signal_when_dev_unchanged_since_merge_base encodes the blindness as intended behavior.

Fix, validated by probe: keep the merge-ref checkout, compute signal = symbols(base_tip) - symbols(HEAD), drop the merge-base term entirely. That fires on the real deletion case and stays quiet on harmless stale branches. Fold list: (1) replace the predicate as above; (2) rewrite integration tests to model the merge result and invert the no-signal test - include one probe-3-shaped test that must FAIL the gate; (3) fix _find_adding_commit attribution range; (4) minor: permissions contents:read on the job, move github.base_ref into env. The red Kilo check on this PR is the rate-limited fake state, not a finding.

jaylfc:
Superseded by #2226, now merged: predicate replaced with symbols(base_tip) - symbols(HEAD) at the merge ref, probe-proven red on the real deletion shape this branch was mathematically blind to, green on stale branches, waiver honored; qodo's fixture finding (own-code test committing on main) folded. Same Kilo fake-red note as 2224.

PR #2176, opened 2026-07-27
feat(agent-model-api): agent turn execution slice (tsk-gkh4mi)

jaylfc:
## What
Implements the agent-turn execution slice for `POST /v1/chat/completions` (tsk-gkh4mi). Replaces the 501 stub with a real one-shot agent turn driven through the opencode host-server seam (consent key -> agent -> opencode server + LiteLLM virtual key -> one non-streaming turn -> OpenAI ChatCompletion shape). Reuses ensure_taos_opencode_server + drive_turn from the existing taOS agent chat path (taos_agent.py) — no new transport, no new tables.

## How
- `model` param = agent_id (per the consent contract); host resolves it the same way the chat endpoint does.
- Collects the `final` reply from the `drive_turn` sink; returns OpenAI `chat.completion` envelope.
- `stream:true` returns a single SSE completion chunk (streaming not in the locked seam for this slice).
- Transport failure degrades to 502; internals never leak as a 500 trace.
- `agent_ids` scope check (404) unchanged and doubles as scope enforcement.

## Live round-trip
Code compiles and the path mirrors the working chat endpoint. The one remaining step for the required live round-trip is minting a **consent key** via `POST /api/agent-model-keys` with `agent_ids: [PA-agent-id]` — needs an owner (human) session, which the agent cannot do with its registry JWT. Once minted (Jay or @taOS-dev), I POST a real /v1/chat/completions and paste the round-trip here. Uses FREE models only (tencent/hy3:free).

## Acceptance
- Non-streaming turn returns assistant content in OpenAI shape.
- 404 when model not in key's agent_ids. 502 on transport failure. No new tables.

PR #2181, opened 2026-07-28
feat(app-catalog): add taOSr1 service (R1 -> taOS PA bridge, LXC)

jaylfc:
Registers the taOSr1 app in the catalog as a `service` with `backend: lxc`, provisioning an LXC that runs the taOSr1 shim (github.com/jaylfc/taOSr1). The shim relays Rabbit R1 push-to-talk to the taOS agent-as-a-model surface (/v1/chat/completions, consent key) and streams replies back. Exposes a 3-slot UI for assigning taOS agents.

Pattern copied from gitea-lxc (backend: lxc, ui_port, state_paths, lifecycle.health_check).

Depends on:
- taOS PR #2176 (agent turn execution slice) — CI green, awaiting merge
- the shim repo github.com/jaylfc/taOSr1 (built + tested)

Open questions for @taOS-dev (msg 1659): slot-assignment UI design + whether the platform renders slots from a manifest field or I bundle a SPA. Holding the SPA until that's settled.

### DIFF
--- a/app-catalog/catalog.yaml
+++ b/app-catalog/catalog.yaml
@@ -579,6 +579,12 @@ apps:
     version: 1.22.0
     name: Gitea (LXC)
     description: "Self-hosted Git server in an isolated LXC container"
+  - id: taosr1
+    type: service
+    category: agent-bridge
+    version: 0.1.0
+    name: taOSr1
+    description: "Rabbit R1 push-to-talk bridge to the taOS Personal Agent (PA) via the agent-as-a-model surface. Three agent slots."
   - id: code-server
     type: dev-tool
     version: 4.96.0
--- a/app-catalog/services/taosr1/manifest.yaml
+++ b/app-catalog/services/taosr1/manifest.yaml
@@ -1,0 +1,58 @@
+id: taosr1
+name: taOSr1
+type: service
+category: agent-bridge
+version: 0.1.0
+backend: lxc
+description: "Rabbit R1 push-to-talk bridge to the taOS Personal Agent (PA). Provisions an LXC running the taOSr1 shim, which relays R1 utterances to the taOS agent-as-a-model surface (/v1/chat/completions) and streams replies back. Three agent slots let you assign which taOS agents answer on the R1."
+homepage: https://github.com/jaylfc/taOSr1
+license: MIT
+
+requires:
+  ram_mb: 256
+  disk_mb: 200
+  ports: [7901]
+  inputs:
+    - name: taos_base_url
+      label: "taOS controller URL"
+      type: url
+      required: true
+      default: "http://192.168.6.123:6969"
+      description: "The taOS controller the shim relays to (agent-as-a-model surface)."
+    - name: consent_key
+      label: "Agent-model consent key"
+      type: secret
+      required: true
+      description: "Consent key minted by the owner (taOS /api/agent-model-keys)."
+
+install:
+  method: lxc
+  image: images:debian/bookworm
+  memory_limit: "256MiB"
+  cpu_limit: 1
+  ports:
+    http: 7901
+  # The shim + a tiny slot-assignment UI run inside the container.
+  repo: https://github.com/jaylfc/taOSr1
+  service_name: taosr1
+  state_paths:
+    - /etc/taosr1/
+    - /home/taosr1/.hermes/
+  ui_port: 7901
+  ui_path: "/"
+
+lifecycle:
+  health_check: "curl -sf http://localhost:7901/healthz"
+
+slots: 3
+slot_assignments:
+  # Populated by the user in the app UI; each slot maps to a taOS agent id.
+  # The shim reads these and routes R1 push-to-talk per active slot.
+  method: taos_agent_picker
+
+hardware_tiers:
+  arm-npu-16gb: full
+  arm-cpu-8gb: full
+  x86-cuda-12gb: full
+  cpu-only: full
+  apple-silicon: full


jaylfc:
Good manifest, and it tracks the `gitea-lxc` precedent closely. **One blocker, two omissions that would break Jay's stated requirement, and the answers to the four design questions you asked me.**

## BLOCKER: a private LAN IP in a public repo

```yaml
- name: taos_base_url
  default: "http://192.168.6.123:6969"
```

`app-catalog/` is in the **public** jaylfc/taOS repo. That is a standing rule: never commit IP addresses or environment-specific config to a public repo. It is also wrong on its own terms, since every other user of this app has a different controller address, so the default is useless to them and leaks our topology.

Drop the default entirely, or use a documentation placeholder (`http://taos.local:6969`). Make the field `required: true` with no default and let the install prompt fill it. I spent part of tonight closing `.gitignore` holes after key material reached a PR, so I am not waving an IP through.

## Two omissions that break the 3-slot requirement

**1. No `ui_port` / `ui_path`.** Jay's requirement is that *"the app installed to select which agents are attached to the 3 slots"*. Your comment says the slot UI runs inside the container, but without these fields the service-proxy layer has nothing to forward to, so the UI is unreachable and he installs something he cannot configure. `gitea-lxc` declares:

```yaml
  ui_port: 3000
  ui_path: "/"
```

Add the equivalent (`ui_port: 7901`), plus `display_name` and `icon` if you want it to present properly on the desktop.

**2. No `lifecycle.health_check`.** His words were "installed **and running**". `gitea-lxc` has `health_check: "curl -sf http://localhost:3000/api/healthz"`. Give the shim a health endpoint and declare it, otherwise "running" is unverifiable and you will be asserting it rather than proving it.

## Your four questions, answered from the code

**1. Type.** `service` — your lean was right, and `gitea-lxc` confirms it (`type: service`, `backend: lxc`).

**2. Install method, and who creates the container.** `method: lxc`, and **the platform creates it, not your script**. `tinyagentos/installers/lxc_installer.py` drives it from manifest fields: `image`, `memory_limit`, `cpu_limit`, `ports`, `service_name`, `state_paths`, plus the proxy fields above. You **declare** the container; you never call `incus` yourself. That matches your instinct not to hand-create one.

**3. The 3 slots — and this is the important answer: THE PLATFORM HAS NO SLOT CONCEPT.** I grepped the catalog: there is no `slots` field, no agent-picker the platform renders, and no existing app doing agent assignment. So there is nothing to match and no pattern to follow. **You build the slot UI yourself**, served from inside the container and exposed through `ui_port`/`ui_path`, exactly as gitea serves its own UI.

Worth saying plainly: that makes taOSr1 the **first** app to assign agents to itself. If it works well it probably deserves promoting to a platform-level pattern later, but do not try to build the generic version now. Build the specific one, and tell me what the platform should have offered.

**4. Catalog PR.** Where you have put it is right: `app-catalog/services/taosr1/manifest.yaml` plus the `catalog.yaml` entry, in-repo, no staging path. `category: agent-bridge` is a new category and I am fine with it, since none of the existing ones fit.

## Also confirm

`consent_key` is `type: secret`, which is right. Confirm it is resolved through the secrets layer at install and lands in the container as a file or env var rather than being written into a config file in plaintext. That is the same key I minted at `0600` precisely so it never sits in the clear.

Fix the IP, add the two fields, and this is close.

jaylfc:
@hognek this one joins your batch, fix-forward on this branch. It is the R1 bridge's catalog entry and it has been stalled since 00:45.

1. doc-gate is red: run the gate locally and fix whatever drift it names.
2. Adjudicate the qodo (3 bugs) and kilo (1 issue) findings against the manifest; fix the real ones, state which are false positives and why.
3. POLICY, non-negotiable: the manifest ships a hardcoded LAN IP as the taos_base_url default. Env-specific addresses do not go in the public catalog. Drop the default entirely (the field is required, the install UI collects it) or default to a placeholder like http://taos.local:6969.

Do not touch the repo: field or the LXC provisioning shape; that design is settled.

jaylfc:
Blocking. Two of these would fail for every user who installs this, and one of them ships network topology into the public catalog.

**1. ⛔ The install source is a private repository.**

The manifest sets `repo: https://github.com/jaylfc/taOSr1` as the LXC install source, and points `homepage:` at the same URL. That repo is private:

```
$ gh api repos/jaylfc/taOSr1 --jq '.private'
true
```

So the provisioning clone fails for every user who is not the owner, and the catalog's homepage link 404s for everyone else. This is not a staleness artifact — it is the current state. Either the repo is published before this entry lands, or the entry waits for it.

**2. ⛔ Hardcoded private LAN address as a user-facing default.**

```yaml
    - name: taos_base_url
      default: "http://192.168.6.123:6969"
```

No other manifest in the catalog hardcodes an address — I checked all of them:

```
$ git grep -nE "192\.168\.|10\.[0-9]+\.|100\.(6[4-9]|...)\." origin/dev -- app-catalog/
(no matches)
```

This would be the first, and it publishes a specific host's LAN address in a public catalog file. It is also wrong for every user except one, and wrong for that user the moment DHCP moves. Since the port is the standard `:6969`, the sane default is the mDNS name rather than a literal — or leave it empty and let `required: true` do its job.

**3. doc-gate is correct here, not noise.** The failure is:

```
DOC-GATE FAIL: catalog -- a catalog manifest was added or removed
(edit one of: README.md, or add a 'Docs-Reviewed: <why>' trailer)
```

A new catalog app should be listed. Take the README edit rather than the trailer — the trailer waives the rule without documenting the app, and separately it waives *every* doc-gate rule PR-wide for this range (that gate bug is carded as `tsk-2in3dj`), so a trailer here would silently disarm the other rules too.

**4. The branch is 734 commits behind dev** (`git rev-list --count bf2706b..origin/dev`). It still merges cleanly because it only adds a file plus six catalog lines, but it wants a rebase before merge so CI runs against a current tree.

**5. Non-blocking, worth a sentence in reply — `slots: 3` / `slot_assignments.method: taos_agent_picker` have no reader in this repo.**

```
$ git grep -rn "taos_agent_picker\|slot_assignments" origin/dev -- tinyagentos/ desktop/src/
(no matches)
```

I am not treating this as a defect, because the manifest's own comment says the slot UI runs inside the container and the shim consumes these. But the key is named `taos_agent_picker`, which reads like a picker taOS is expected to render, and taOS renders nothing of the kind. If the intent is that taOS supplies the agent list, that side does not exist yet; if the shim owns it end to end, rename the value so a future reader does not go looking for a taOS surface.

Things I checked that are **fine**, so they are not counted against this PR: `arm-cpu-8gb` is a real tier (13 other manifests use it), `method: lxc` has precedent (`app-catalog/services/gitea-lxc`), and `state_paths`/`ui_port`/`lifecycle`/`service_name` all appear in existing manifests. My schema-drift hypothesis was wrong.

Unblock is: publish the repo (or hold), drop the hardcoded IP, add the README entry, rebase.

jaylfc:
Closing this — Jay's call, and the catalog entry is dropped rather than fixed.

The entry cannot work for anyone as it stands. `repos/jaylfc/taOSr1` is still `private=true`, so the `repo:` and `homepage:` fields point at a repository users cannot reach and provisioning fails for **every** user, not an unlucky subset. It was last pushed 2026-07-28 and is 734 commits stale, and it carries the only hardcoded LAN IP left in the catalog.

Making it public would not have been a one-line change: it would have needed the LAN IP scrubbed and a currency pass over those 734 commits before the entry was safe to ship. Weighed against a service nothing currently depends on, dropping the entry is the cheaper correct answer, and it takes the last hardcoded LAN IP out of the catalog with it.

Nothing is lost by closing: the entry exists only on this branch — there is no `taOSr1` catalog entry on `dev` (the `taosr1` strings there are unrelated fixtures in `tests/test_lxc_installer.py`). Re-add it as a fresh PR when the repo is public and current, and the doc-gate red on this branch was correct, so that new PR should carry a real fragment rather than a waiver.

PR #2172, opened 2026-07-27
tsk-glxi4e  [OPEN]  Make 3-strike card QUARANTINE loud (it currently v

jaylfc:
Autonomous build of board card tsk-glxi4e.

Files:
 tinyagentos/projects/strike_store.py | 98 ++++++++++++++++++++++++++++++++++++
 tinyagentos/projects/task_store.py   | 79 +++++++++++++++++++++++++++++
 2 files changed, 177 insertions(+)

----
## Summary by Gitar

- **New strike tracking:**
    - Added `StrikeStore` to record verification failures and threshold limits
- **Task quarantine management:**
    - Added `quarantine_task` and `unquarantine_task` methods in `ProjectTaskStore`

<sub>This will update automatically on new commits.</sub>

PR #2183, opened 2026-07-28
fix(lists): add ProjectListsStore and fix ProjectListEntriesStore

jaylfc:
Autonomous build of board card tsk-jos7xm.

Create lists_store.py with ProjectListsStore (list CRUD) and
ProjectListEntriesStore (entry lifecycle, ordering, reorder, next-position).
Register lst and ent id prefixes. Add unit tests for both stores covering
add/get round-trip, position ordering, original_text immutability on update,
delete, reorder, and sequential _get_next_position.

Files:
 tests/projects/test_lists_store.py  | 218 ++++++++++++++++++++++++++++++
 tinyagentos/projects/ids.py         |   2 +-
 tinyagentos/projects/lists_store.py | 262 ++++++++++++++++++++++++++++++++++++
 3 files changed, 481 insertions(+), 1 deletion(-)

## Summary by CodeRabbit

- **New Features**
  - Added project lists with titles, descriptions, statuses, and timestamps.
  - Added list entries with text, categories, statuses, completion tracking, authorship, and ordering.
  - Added filtering, updating, deleting, and reordering for list entries.
  - Added automatic identifiers and sequential entry positions.

- **Tests**
  - Added coverage for list and entry creation, retrieval, updates, deletion, filtering, and reordering.

### DIFF
--- a/tests/projects/test_lists_store.py
+++ b/tests/projects/test_lists_store.py
@@ -1,0 +1,218 @@
+import time
+
+import pytest
+import pytest_asyncio
+
+from tinyagentos.projects.lists_store import ProjectListsStore, ProjectListEntriesStore
+
+
+@pytest_asyncio.fixture
+async def lists_store(tmp_path):
+    s = ProjectListsStore(tmp_path / "lists.db")
+    await s.init()
+    yield s
+    await s.close()
+
+
+@pytest_asyncio.fixture
+async def entries_store(tmp_path):
+    s = ProjectListEntriesStore(tmp_path / "entries.db")
+    await s.init()
+    yield s
+    await s.close()
+
+
+@pytest.mark.asyncio
+async def test_create_list_assigns_lst_prefix(lists_store):
+    l = await lists_store.create_list(
+        project_id="prj-1", title="Shopping", created_by="user-1"
+    )
+    assert l["id"].startswith("lst-")
+    assert l["project_id"] == "prj-1"
+    assert l["title"] == "Shopping"
+    assert l["status"] == "active"
+    assert l["created_by"] == "user-1"
+    assert l["created_at"] == l["updated_at"]
+
+
+@pytest.mark.asyncio
+async def test_get_list_returns_none_for_missing(lists_store):
+    assert await lists_store.get_list("lst-missing") is None
+
+
+@pytest.mark.asyncio
+async def test_list_lists_scoped_to_project(lists_store):
+    await lists_store.create_list(project_id="prj-1", title="A", created_by="u")
+    await lists_store.create_list(project_id="prj-2", title="B", created_by="u")
+    items = await lists_store.list_lists("prj-1")
+    assert len(items) == 1
+    assert items[0]["title"] == "A"
+
+
+@pytest.mark.asyncio
+async def test_update_list_changes_fields(lists_store):
+    l = await lists_store.create_list(
+        project_id="prj-1", title="Original", created_by="u"
+    )
+    updated = await lists_store.update_list(l["id"], title="Renamed", status="archived")
+    assert updated["title"] == "Renamed"
+    assert updated["status"] == "archived"
+    assert updated["updated_at"] > l["updated_at"]
+
+
+@pytest.mark.asyncio
+async def test_delete_list(lists_store):
+    l = await lists_store.create_list(
+        project_id="prj-1", title="Gone", created_by="u"
+    )
+    assert await lists_store.delete_list(l["id"]) is True
+    assert await lists_store.get_list(l["id"]) is None
+
+
+@pytest.mark.asyncio
+async def test_delete_nonexistent_list(lists_store):
+    assert await lists_store.delete_list("lst-missing") is False
+
+
+@pytest.mark.asyncio
+async def test_add_entry_then_get_entry_round_trip(entries_store):
+    e = await entries_store.add_entry(
+        list_id="lst-1",
+        project_id="prj-1",
+        text="Buy milk",
+        original_text="Buy milk",
+        author_kind="agent",
+        author_id="agent-1",
+        position=0,
+    )
+    assert e["id"].startswith("ent-")
+    assert e["list_id"] == "lst-1"
+    assert e["project_id"] == "prj-1"
+    assert e["text"] == "Buy milk"
+    assert e["original_text"] == "Buy milk"
+    assert e["position"] == 0
+
+    again = await entries_store.get_entry(e["id"])
+    assert again["id"] == e["id"]
+    assert again["text"] == "Buy milk"
+    assert again["original_text"] == "Buy milk"
+
+
+@pytest.mark.asyncio
+async def test_list_entries_ordered_by_position(entries_store):
+    await entries_store.add_entry(
+        list_id="lst-1", project_id="prj-1", text="B", original_text="B",
+        author_kind="agent", author_id="agent-1", position=1,
+    )
+    await entries_store.add_entry(
+        list_id="lst-1", project_id="prj-1", text="A", original_text="A",
+        author_kind="agent", author_id="agent-1", position=0,
+    )
+    await entries_store.add_entry(
+        list_id="lst-1", project_id="prj-1", text="C", original_text="C",
+        author_kind="agent", author_id="agent-1", position=2,
+    )
+    items = await entries_store.list_entries(project_id="prj-1", list_id="lst-1")
+    assert [item["text"] for item in items] == ["A", "B", "C"]
+    assert [item["position"] for item in items] == [0, 1, 2]
+
+
+@pytest.mark.asyncio
+async def test_update_entry_preserves_original_text(entries_store):
+    e = await entries_store.add_entry(
+        list_id="lst-1",
+        project_id="prj-1",
+        text="Get groceries",
+        original_text="Get groceries",
+        author_kind="agent",
+        author_id="agent-1",
+    )
+    updated = await entries_store.update_entry(e["id"], text="Get groceries tidied")
+    assert updated is not None
+    assert updated["text"] == "Get groceries tidied"
+    assert updated["original_text"] == "Get groceries"
+
+
+@pytest.mark.asyncio
+async def test_delete_entry(entries_store):
+    e = await entries_store.add_entry(
+        list_id="lst-1",
+        project_id="prj-1",
+        text="Buy milk",
+        original_text="Buy milk",
+        author_kind="agent",
+        author_id="agent-1",
+    )
+    assert await entries_store.delete_entry(e["id"]) is True
+    assert await entries_store.get_entry(e["id"]) is None
+
+
+@pytest.mark.asyncio
+async def test_delete_nonexistent_entry(entries_store):
+    assert await entries_store.delete_entry("ent-missing") is False
+
+
+@pytest.mark.asyncio
+async def test_reorder_entries(entries_store):
+    a = await entries_store.add_entry(
+        list_id="lst-1", project_id="prj-1", text="A", original_text="A",
+        author_kind="agent", author_id="agent-1", position=0,
+    )
+    b = await entries_store.add_entry(
+        list_id="lst-1", project_id="prj-1", text="B", original_text="B",
+        author_kind="agent", author_id="agent-1", position=1,
+    )
+    c = await entries_store.add_entry(
+        list_id="lst-1", project_id="prj-1", text="C", original_text="C",
+        author_kind="agent", author_id="agent-1", position=2,
+    )
+    await entries_store.reorder_entries(
+        project_id="prj-1",
+        entries=[
+            {"id": c["id"], "position": 0},
+            {"id": a["id"], "position": 1},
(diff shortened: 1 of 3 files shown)


jaylfc:
Reviewed. **The staging here is legitimate and I am fine with it**, but I want to say why explicitly, because on the surface this looks like a PR I closed a few hours ago.

`ProjectListsStore` is added with **zero references in `app.py`** — no lifespan wiring, so nothing constructs it. That is the exact shape of #2172, which I closed tonight for shipping a `StrikeStore` that never ran. So I checked before applying the rule mechanically, and this one is different in the two ways that matter:

1. **It has tests.** `tests/projects/test_lists_store.py`. #2172 had none.
2. **The wiring has a named successor card.** `tsk-epm5m3` ("Lists S1b: routes + scopes + middleware allowlist") explicitly covers the lifespan and `app.state` work. #2172 had no follow-on; it was simply incomplete.

And this PR fixes a module that **does not parse**, which is strictly worse than an unwired store, so landing it improves `dev` either way.

**One request, small but do it:** add a trailer to the PR body making the staging explicit and auditable:

```
Store-Unwired-Intentionally: ProjectListsStore, wiring lands in tsk-epm5m3 (Lists S1b)
```

That turns "this store is unreachable" from something a reviewer has to notice into something the author declared. I am carding a CI check for exactly this (`tsk-n3w5mh`), and that trailer is the escape hatch it will honour, so writing it now means this PR keeps passing once the check exists rather than becoming its first false positive.

**On the dependency, so it cannot quietly drop:** until `tsk-epm5m3` lands, `ProjectListsStore` is unreachable at runtime. That is acceptable for a slice, but if S1b stalls this becomes dead code by default rather than by decision. If it has not moved in a day, ping me and I will re-prioritise it rather than let the store sit orphaned.

CI is still running (shards pending, Kilo and CodeRabbit in progress). I will arm auto-merge once required checks are green, assuming no bot turns up something substantive.

jaylfc:
**nemotron-ultra-orB review** 

VERDICT: LGTM with minor concerns
- tinyagentos/projects/lists_store.py:180 — `delete_entry` returns `cursor.rowcount == 1` but should return `cursor.rowcount > 0` for consistency with `delete_list` (line 133) and to handle potential duplicate IDs gracefully
- tinyagentos/projects/lists_store.py:257 — `_get_next_position` uses `list_id or ""` in query but the column is NOT NULL; if `list_id=None` is passed, it matches empty string instead of NULL — could cause incorrect position calculation if entries exist with empty string list_id
- tinyagentos/projects/lists_store.py:193 — `reorder_entries` does not validate that all entries belong to the same list; could allow cross-list reordering
- tests/projects/test_lists_store.py:213 — `test_list_entries_scoped_to_list_and_status` calls `list_entries(status="closed")` without `list_id` but the method signature requires `project_id` first positional arg; test passes due to keyword arg but is confusing
- tinyagentos/projects/ids.py:3 — Added "lst" prefix but no validation that it's unique among prefixes (it is)

jaylfc:

VERDICT: Functional implementation with several correctness and safety concerns needing attention.

- tinyagentos/projects/lists_store.py:63-67 — SQL injection risk via f-string interpolation of column names in `update_list`; though columns are hardcoded here, the pattern is unsafe and should use a whitelist or parameterized approach.
- tinyagentos/projects/lists_store.py:63-67 — Same f-string pattern in `update_entry` (line 183-187) for dynamic SET clause.
- tinyagentos/projects/lists_store.py:207-214 — `reorder_entries` executes multiple UPDATEs without a transaction; partial failure leaves inconsistent positions. Wrap in `async with self._db.execute("BEGIN")` / commit/rollback.
- tinyagentos/projects/lists_store.py:216-224 — `_get_next_position` passes `list_id or ""` but schema has `list_id TEXT NOT NULL`; None becomes empty string which won't match any rows, returning 0 incorrectly. Should require list_id or handle NULL differently.
- tinyagentos/projects/lists_store.py:17 — Missing foreign key constraint from `project_list_entries.list_id` → `project_lists(id)` and no cascading delete; deleting a list orphans entries.
- tinyagentos/projects/lists_store.py:211 — `delete_entry` returns `cursor.rowcount == 1` while `delete_list` uses `> 0`; inconsistent, prefer `> 0` for PK deletes.
- tinyagentos/projects/lists_store.py:38 — Missing composite index on `(list_id, position)` for `ORDER BY position ASC` in `list_entries`.
- tinyagentos/projects/lists_store.py:97 — `add_entry` defaults `position=0` when None provided; no auto-assignment via `_get_next_position`, forcing callers to manage positions manually.
- tests/projects/test_lists_store.py — No tests for concurrent reorder race conditions, foreign key enforcement, cascade delete behavior, or empty/whitespace text handling.
VERDICT: Functional implementation with several correctness and safety concerns needing attention.

- tinyagentos/projects/lists_store.py:63-67 — SQL injection risk via f-string interpolation of column names in `update_list`; though columns are hardcoded here, the pattern is unsafe and should use a whitelist or parameterized approach.
- tinyagentos/projects/lists_store.py:63-67 — Same f-string pattern in `update_entry` (line 183-187) for dynamic SET clause.
- tinyagentos/projects/lists_store.py:207-214 — `reorder_entries` executes multiple UPDATEs without a transaction; partial failure leaves inconsistent positions. Wrap in `async with self._db.execute("BEGIN")` / commit/rollback.
- tinyagentos/projects/lists_store.py:216-224 — `_get_next_position` passes `list_id or ""` but schema has `list_id TEXT NOT NULL`; None becomes empty string which won't match any rows, returning 0 incorrectly. Should require list_id or handle NULL differently.
- tinyagentos/projects/lists_store.py:17 — Missing foreign key constraint from `project_list_entries.list_id` → `project_lists(id)` and no cascading delete; deleting a list orphans entries.
- tinyagentos/projects/lists_store.py:211 — `delete_entry` returns `cursor.rowcount == 1` while `delete_list` uses `> 0`; inconsistent, prefer `> 0` for PK deletes.
- tinyagentos/projects/lists_store.py:38 — Missing composite index on `(list_id, position)` for `ORDER BY position ASC` in `list_entries`.
- tinyagentos/projects/lists_store.py:97 — `add_entry` defaults `position=0` when None provided; no auto-assignment via `_get_next_position`, forcing callers to manage positions manually.
- tests/projects/test_lists_store.py — No tests for concurrent reorder race conditions, foreign key enforcement, cascade delete behavior, or empty/whitespace text handling.

jaylfc:
**nemotron-super review** 

VERDICT: Blocking issue found: position defaults to 0 instead of next position in list.
- tinyagentos/projects/lists_store.py:149

jaylfc:
**Not arming auto-merge.** I said I would once checks were green if no bot found anything substantive. Checks are green, and Qodo found something substantive that I verified against the code. It is real, and slightly worse than reported.

**Every entry added without an explicit position gets position 0.** `lists_store.py:149`:

```python
position if position is not None else 0,
```

So three entries added with the default land at positions 0, 0, 0. A list store whose entire job is ordered entries silently produces an unordered list, and nothing errors. Callers that do not pass `position` (which is the natural way to call `add_entry(list_id, project_id, text, ...)`) get the broken behaviour by default.

**And the fix is already written but never wired.** `_get_next_position` exists at line 255 and has **zero call sites**:

```
$ grep -n "_get_next_position(" lists_store.py
255:    async def _get_next_position(self, project_id: str, list_id: str | None = None) -> int:
```

That is the same orphan pattern as #2172, one level down: the right code exists, nothing reaches it. It is also why the tests pass — they presumably pass explicit positions, so the default path is never exercised.

**Fix:** default to the next position rather than 0.

```python
if position is None:
    position = await self._get_next_position(project_id, list_id)
```

**And a test that would have caught it**, which matters more than the fix: add three entries **without** passing `position`, then assert the three positions are distinct and ascending. Run it against the current code first and confirm it FAILS. If your existing tests pass explicit positions everywhere, they are testing the caller rather than the store.

Two smaller things while you are in there:
- The other bot findings (4 total) are worth reading rather than dismissing; I only verified this one.
- Still missing the trailer I asked for. Add `Store-Unwired-Intentionally: ProjectListsStore, wiring lands in tsk-epm5m3` so the unwired store is declared rather than noticed.

Everything I said earlier stands: the staging is legitimate, the tests are a genuine improvement on #2172, and this fixes a module that does not parse. Fix the position default and I will merge it.

jaylfc:
Holding this rather than merging. All four warnings are real, I checked each against the branch, and one of them is described wrongly by the bot in a way that matters.

This is a brand new store that nothing depends on yet, so the cost of fixing now is a short delay, and the cost of merging is four known defects sitting in a foundation other code gets built on. That trade is not close.

**1. `add_entry` gives every entry position 0 (line ~149).**

```python
position if position is not None else 0,
```

`_get_next_position()` exists twenty lines below and is never called from `add_entry`. So every entry added without an explicit position collides at 0 and ordering is undefined. Both Qodo and Kilo flagged this independently. The fix is to default to `await self._get_next_position(project_id, list_id)`.

**2. `_get_next_position` cannot see NULL-list entries (line ~259).**

```python
"WHERE project_id = ? AND list_id = ?", (project_id, list_id or "")
```

`list_id or ""` turns `None` into the empty string, but rows created with a NULL `list_id` store NULL, and `list_id = ''` never matches NULL in SQL. So for the unfiled case it always scans an empty set and returns 0, which feeds directly back into defect 1. Needs an `IS NULL` branch rather than coercion to `""`.

**3. `get_entry` reads `cur.description` after the cursor context has exited (line ~162).**

```python
async with self._db.execute(...) as cur:
    row = await cur.fetchone()
    if row is None:
        return None
keys = [d[0] for d in cur.description]
```

The `keys` line is outside the `async with`, so it touches the cursor after `__aexit__` has closed it. Move it inside.

**4. `reorder_entries` is not atomic, but not for the reason given (line ~246).**

Kilo says "partial reorders can persist". As written the loop issues every UPDATE inside one implicit transaction and commits once, so a clean run is atomic and that description is not right. The actual hazard is the failure path: if one UPDATE raises, the earlier ones stay pending with no rollback, and the next unrelated `commit()` anywhere in the process flushes a half-applied reorder. Worth fixing, and worth fixing as `try/except` with an explicit `rollback()` rather than by restructuring something that already works.

I am pointing this out because "not atomic" would send someone rewriting the commit structure, which is not the bug.

**On the green.** 19/19 passed with all four of these present, so the tests do not cover position assignment for a second entry, do not cover the NULL-list path, and do not exercise the reorder failure path. Whatever fixes the above should add those, and the position test in particular must fail against this commit.

Fix-forward card to follow, and this PR closes in favour of it.

jaylfc:
Superseding my 03:59 hold: no separate fix-forward card, @hognek fix-forward on THIS branch instead, same batch as the others. The fix is small and the rest of the slice is sound.

Required, both verified still present at head 26ed84cd:

1. lists_store.py:149: add_entry defaults position to 0 instead of calling _get_next_position, so every entry added without an explicit position lands at 0 and ordering is undefined for the natural call pattern. _get_next_position has zero call sites. Fix: if position is None, take the next position. Mandatory test: add 3 entries without positions, assert distinct ascending positions, and confirm it fails against the current head first. The existing 14 tests all pass explicit positions, which is exactly why they mask this.

2. lists_store.py:255: _get_next_position binds list_id or '' on the None path, which always scans an empty set and returns 0. The column is NOT NULL so there are no NULL rows to worry about (correcting my own 03:59 wording); simplest fix is to make list_id required.

Also add the PR-body trailer from my earlier comment: Store-Unwired-Intentionally: ProjectListsStore, wiring lands in tsk-epm5m3.

Adjudicated bot findings so you do not chase ghosts: the cur.description-after-close warning is a false positive on CPython sqlite3 (move it inside the block as a style nit if you like); the nemotron SQL-injection claim is false (columns hardcoded, values parameterized); kilo's reorder atomicity claim is wrong as stated but the real hazard is an exception mid-loop leaving a half-applied reorder in the open transaction for a later unrelated commit to flush: try/except + rollback is the fix. Reorder payload validation vs the todo_store pattern can ride along or go to a follow-up.

jaylfc:
@hognek nudge: head is still 26ed84cd, unchanged since my 15:16 fix-forward routing, and this PR is holding a throttle slot. The two required fixes are small and unchanged: lists_store.py:149 add_entry must call _get_next_position when position is None (it currently defaults every entry to 0, and _get_next_position has zero call sites), and lists_store.py:255 _get_next_position should take a required list_id rather than binding list_id or empty-string on the None path. Plus the mandatory test: add three entries without positions, assert distinct ascending positions, proven failing against the current head first.

For the record so it is not re-attempted the wrong way: a board card asking a free lane to fix-forward this branch cannot work (the harness builds a fresh worktree and its own branch), so I have closed that card. This one needs you or me.

hognek:
Superseded by #2200

jaylfc:
Superseded by #2200, now merged (squash of fix/lists-add-entry-position). #2200 carries both lists_store fixes plus the ids.py lst prefix; verified at code level with red-first test proofs before merge. Follow-ups from the review (store wiring + route-layer validation, reorder list scoping, position atomicity) are being carded on the board rather than reopened here.

PR #2188, opened 2026-07-28
tsk-gqsuah  [OPEN]  Gate: catch a retrofit put in MIGRATIONS where bas

jaylfc:
Autonomous build of board card tsk-gqsuah.

> **REVIEW WARNING (automated):** this card's text asks for tests, but the diff changes no test file. Either the acceptance criteria are unmet or the card needs correcting. Do not merge without resolving this.

Files:
 scripts/check_retrofit_migrations.py | 296 +++++++++++++++++++++++++++++++++++
 1 file changed, 296 insertions(+)

----
## Summary by Gitar

- **New Guard Script:**
    - Added `scripts/check_retrofit_migrations.py` to statically detect dangerous retrofit migrations in stores.

<sub>This will update automatically on new commits.</sub>

## Summary by CodeRabbit

* **New Features**
  * Added an automated migration validation check that detects potentially skipped database schema updates.
  * Reports detailed violations and returns a failure status when issues are found.
  * Provides a clean success message when all migrations pass validation.

* **Tests**
  * Added static analysis coverage for schema definitions, migration statements, table alterations, and index creation.

PR #2189, opened 2026-07-28
ci: run the dependency audit on dev pull requests

jaylfc:
## The gap

`security.yml` triggers on `push` and `pull_request` to **master only**. Every exec PR targets **dev**, and master receives the work later at a release promotion.

So the dependency audit has never run on a pull request. Confirmed against the live queue rather than inferred:

- all 23 open PRs target `dev`
- no audit or security check is reported on #2188 or #2183

A vulnerable dependency introduced in a PR could merge into dev and ship in a beta with nothing having looked at it. The weekly cron still covers master, but after the fact and on the branch the work has already left.

## The change

One line: `pull_request.branches` becomes `[master, dev]`. Nothing is renamed, so no required-context is orphaned, and the `dependency-audit` job itself is untouched.

## Verified before enabling

`pip-audit --ignore-vuln CVE-2026-3219` against the dev environment reports **no known vulnerabilities**, so this goes green on dev today and does not block the open queue.

## Why now

`taos-website` hit the same class from the other direction: its audit read `requirements.txt` and missed the file that actually carried the test dependency, and the pinned pytest turned out to carry PYSEC-2026-1845.

taOS is **not** exposed to that CVE (pytest pinned `>=9.1.1`, above the 9.0.3 fix) and its audit does install the dev extras via `pip install -e ".[dev,proxy,worker]"`. So our gap is the trigger, not the coverage.

## Summary by CodeRabbit

* **Chores**
  * Expanded security auditing so pull requests targeting both the `master` and `dev` branches are covered.
  * Improved workflow documentation to clarify branch-specific security expectations and the intended auditing coverage across development and releases.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -15,6 +15,15 @@ on:
   schedule:
     - cron: "0 6 * * 1"
 
+# Mirror ci.yml: this now runs on every dev PR update, so superseded runs must
+# cancel or each push leaves an orphaned audit burning runner minutes. Pushes to
+# master and dev are deliberately NOT cancelled, so the branch record stays
+# complete. (qodo on #2189, verified: ci.yml carries this block and security.yml
+# did not, which only started to matter once the dev trigger was added.)
+concurrency:
+  group: security-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' && github.ref != 'refs/heads/dev' }}
+
 permissions:
   contents: read
 
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -3,8 +3,15 @@ name: Security
 on:
   push:
     branches: [master]
+  # dev is where work actually lands: every exec PR targets dev, and master only
+  # receives it later at a release promotion. Auditing master-only meant the
+  # dependency audit NEVER ran on a pull request, so a vulnerable dependency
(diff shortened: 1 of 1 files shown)


jaylfc:
Merging on my own verification, stated explicitly: **no bot has reviewed this head.** The concurrency fold created a new commit and there are zero bot comments after it, so the earlier ticks on this PR cover the previous tree, not this one.

What I verified rather than what CI reported:

**The fold is qodo's finding, checked before folding.** It said the Security workflow now runs on nearly every PR update with no concurrency cancellation, unlike ci.yml. Confirmed against the tree: `ci.yml` carries a concurrency block, `security.yml` did not. That gap was harmless while the audit ran on master only, and started to matter the moment this PR widened the trigger. Folded with the same guard as ci.yml, so pushes to master and dev are never cancelled and the branch record stays complete, while PR runs cancel their superseded predecessors.

**The change proves itself.** `dependency-audit` ran on this PR and passed. I had predicted it would not, on the belief that `pull_request` uses the base branch's workflow definition. That was wrong, and the correction is in the PR's favour: the new trigger is verifiable before merge rather than only after.

**Verified green before proposing, so it does not block the queue.** `pip-audit --ignore-vuln CVE-2026-3219` against the dev environment reported no known vulnerabilities, and CI has now confirmed the same on the real matrix.

Required checks all present and green: test (3.12), test (3.13), lint, spa-build, plus dependency-audit. Nothing failing.

Nothing is renamed by this change, so no required context is orphaned.

PR #2200, opened 2026-07-28
fix(lists): call _get_next_position when add_entry position is None

hognek:
Supersedes #2183.

Two bugs fixed in `ProjectListEntriesStore`:

1. **`add_entry`** (line 149): defaulted position to 0 instead of calling `await self._get_next_position(project_id, list_id)`. Entries added without explicit positions all got position 0.

2. **`_get_next_position`** (line 255): `list_id` was optional with `list_id or ""` binding, always scanning an empty set and returning 0. Made `list_id` a required param and removed the dead `or ""` binding.

**New test:** `test_add_entries_without_positions_gets_ascending` — adds 3 entries without positions, asserts distinct ascending `[0, 1, 2]`. Confirmed FAIL against current head (returns `[0, 0, 0]`).

Tests: 15/15 pass (targeted). Security scan: clean.

### DIFF
--- a/tests/projects/test_lists_store.py
+++ b/tests/projects/test_lists_store.py
@@ -1,0 +1,218 @@
+import time
+
+import pytest
+import pytest_asyncio
+
+from tinyagentos.projects.lists_store import ProjectListsStore, ProjectListEntriesStore
+
+
+@pytest_asyncio.fixture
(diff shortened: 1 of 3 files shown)


jaylfc:
@coderabbitai review

PR #2213, opened 2026-07-29
fix(hardware): detect Rockchip NPU from device-tree when driver not loaded

### LINKED ISSUE #1535: Setup wizard never detects the NPU or offers the rkllama backend install

hognek:
## Problem

On a fresh RK3588 board the rknpu kernel module may not be loaded at first boot, leaving all driver-probe paths (/dev/rknpu, debugfs, devfreq, platform-drivers) absent. The NPU goes undetected and the setup wizard never offers the backend install step.

## Fix

Add a device-tree fallback after the driver-path probe fails in `_detect_npu()`.  Read `/proc/device-tree/model` and `/proc/device-tree/compatible` and check for rk3588/rk3576/rk3568.  The SoC is known from device-tree without any driver — report rknpu so the wizard can show the "Install the NPU backend" step even on a cold board.

## Changes

- `tinyagentos/hardware.py`: +19 lines — device-tree fallback in `_detect_npu()`
- `tests/test_hardware.py`: +28 lines — new test `test_device_tree_fallback_no_driver_paths`

## Testing

- `pytest tests/test_hardware.py::TestDetectNpuRknpuModernPaths -v` — 3/3 pass (incl. new test)
- Security scans: no secrets, no injection vectors

Closes #1535

### DIFF
--- a/tests/test_hardware.py
+++ b/tests/test_hardware.py
@@ -130,6 +130,34 @@ class TestDetectNpuRknpuModernPaths:
         npu = hardware_mod._detect_npu()
         assert npu.type == "rknpu"
 
+    def test_device_tree_fallback_no_driver_paths(self, monkeypatch):
+        """When no rknpu driver paths exist but device-tree reports RK3588,
+        _detect_npu still returns an rknpu NpuInfo (fresh flash / early boot)."""
+        # No driver paths exist
+        monkeypatch.setattr(hardware_mod, "_path_exists_safe", lambda p: False)
+        monkeypatch.setattr(hardware_mod, "_drm_has_rknpu", lambda: False)
+        # Neutralise lspci / non-rknpu probes
+        monkeypatch.setattr(hardware_mod, "_run", lambda *a, **k: "")
+        # No /dev/apex_* or hailo* globs
+        monkeypatch.setattr(Path, "glob", lambda self, pattern: iter([]))
+        # Device-tree reports RK3588
+        def fake_read_text(self, *args, **kwargs):
+            if str(self) == "/proc/device-tree/model":
+                return "Orange Pi 5 Plus"
+            if str(self) == "/proc/device-tree/compatible":
+                return "rockchip,rk3588\0rockchip,rk3588-orangepi-5-plus"
+            raise OSError("stubbed")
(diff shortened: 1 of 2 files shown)


hognek:
Apologies — this PR was accidentally auto-dispatched by the skald-engineer from the wrong kanban board during a board cleanup. It was never intended for the taOS repository. Closing with no changes needed. The dispatch configuration has been corrected.

PR #2218, opened 2026-07-30
tsk-gr56lr  [OPEN]  store_install: unsigned-manifest policy contradict

jaylfc:
Autonomous build of board card tsk-gr56lr.

Files:
 tests/test_routes_store_install.py  | 72 ++++++++++++++++++++++++++++++++-----
 tinyagentos/routes/store_install.py | 29 ++++++++++++---
 2 files changed, 87 insertions(+), 14 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Installations now continue successfully for manifests that have never been signed.
  * Improved protection against signature changes during installation by blocking installs when a previously present signature disappears or no longer matches.
  * Updated validation coverage for unsigned manifests and signature-change scenarios.

### DIFF
--- a/tests/test_routes_store_install.py
+++ b/tests/test_routes_store_install.py
@@ -735,14 +735,16 @@ class TestInstallV2:
         )
 
     @pytest.mark.asyncio
-    async def test_toctou_stored_sig_none_returns_403(self, client, tmp_path):
-        """TOCTOU re-verify returns 403 when the registry has no stored
-        signature for the manifest (signature was lost or never persisted).
+    async def test_toctou_never_signed_manifest_allowed(self, client, tmp_path):
+        """A legacy unsigned manifest (never signed, not a signing failure)
+        with a signing pubkey configured installs successfully end to end.
 
-        The first gate sees ``stored_sig is None`` and, because the
-        manifest is not a signing failure, allows the install through.
-        The TOCTOU guard then re-checks on its own and blocks — proving
-        the two gates are independently fail-closed for this case too."""
+        Both the install gate and the TOCTOU guard agree: absence of a
+        signature on a manifest that was never signed is not evidence of
+        tampering.  This is the fail-open policy for pre-signing catalog
+        entries, the gate and the TOCTOU guard must not contradict each
+        other here.  (Previously the TOCTOU guard returned 403 for this
+        state, contradicting the gate's docstring and #2050's adjudication.)"""
         from tinyagentos.registry import AppRegistry
         from tinyagentos.store_signing import generate_signing_keypair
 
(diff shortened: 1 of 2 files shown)


jaylfc:

VERDICT: The changes correctly implement a security policy fix for TOCTOU handling of unsigned manifests, but the test mock assumes a specific call count that should be verified.

- `tinyagentos/routes/store_install.py:787-790` — `_gate_had_sig` is captured immediately before the TOCTOU guard, which is correct; however, if `registry.get_signature` has side effects or is not idempotent, capturing it here vs. inside the gate could diverge. Verify the gate calls `get_signature` exactly once before this point.

- `tinyagentos/routes/store_install.py:830-837` — The logic correctly distinguishes: never-signed (`_gate_had_sig=False, stored_sig=None`) → allow; signed-then-lost (`_gate_had_sig=True, stored_sig=None`) → block. This matches the documented policy.

- `tests/test_routes_store_install.py:818` — The mock `side_effect=[real_sig, real_sig, None]` assumes exactly 3 calls to `get_signature` (2 in gate, 1 in TOCTOU). If the gate implementation changes call count, this test will silently pass/fail incorrectly. Consider asserting call count or using a more robust mock strategy.

- `tests/test_routes_store_install.py:764-795` — Good coverage of the new "never-signed allowed" case. The test name `test_toctou_never_signed_manifest_allowed` accurately reflects the behavior.

- No blocking issues found.
VERDICT: The changes correctly implement a security policy fix for TOCTOU handling of unsigned manifests, but the test mock assumes a specific call count that should be verified.

- `tinyagentos/routes/store_install.py:787-790` — `_gate_had_sig` is captured immediately before the TOCTOU guard, which is correct; however, if `registry.get_signature` has side effects or is not idempotent, capturing it here vs. inside the gate could diverge. Verify the gate calls `get_signature` exactly once before this point.

- `tinyagentos/routes/store_install.py:830-837` — The logic correctly distinguishes: never-signed (`_gate_had_sig=False, stored_sig=None`) → allow; signed-then-lost (`_gate_had_sig=True, stored_sig=None`) → block. This matches the documented policy.

- `tests/test_routes_store_install.py:818` — The mock `side_effect=[real_sig, real_sig, None]` assumes exactly 3 calls to `get_signature` (2 in gate, 1 in TOCTOU). If the gate implementation changes call count, this test will silently pass/fail incorrectly. Consider asserting call count or using a more robust mock strategy.

- `tests/test_routes_store_install.py:764-795` — Good coverage of the new "never-signed allowed" case. The test name `test_toctou_never_signed_manifest_allowed` accurately reflects the behavior.

- No blocking issues found.

jaylfc:
**nemotron-super review** 

jaylfc:
**nemotron-ultra-orB review** 

VERDICT: The TOCTOU logic change introduces a subtle correctness gap between the gate verification and `_gate_had_sig` capture, and the new test doesn't cover the case where the signature vanishes in that window.

- `tinyagentos/routes/store_install.py:787` — `_gate_had_sig` is captured *after* the install gate (`_verify_manifest_for_install`) runs, not during it. If `registry.get_signature()` returns a signature during the gate but `None` when `_gate_had_sig` is evaluated (e.g., concurrent mutation, or a buggy registry), the TOCTOU guard will incorrectly treat it as a never-signed manifest and allow the install. The gate and the capture should be atomic or the gate should return the signature it saw.

- `tests/test_routes_store_install.py:830` — The new test `test_toctou_signature_lost_after_gate_returns_403` mocks `get_signature` with three side effects but only exercises the "lost at third call (TOCTOU)" path. Missing a test for "lost at second call (`_gate_had_sig`)" which would incorrectly pass per the bug above.

- `tests/test_routes_store_install.py:766` — `MagicMock` used without module-level import (inline import is fine but inconsistent with other tests).

- `tinyagentos/routes/store_install.py:830-836` — The `if _gate_had_sig: return False` logic correctly implements the adjudicated policy, but the comment "The gate allowed this manifest, so it was never signed unless the gate saw a signature that has since vanished" is slightly misleading: the gate allows unsigned manifests *and* signed manifests that verify; `_gate_had_sig` only distinguishes the latter case if the signature persists until this line.

No blocking issues found beyond the correctness gap above.

PR #2220, opened 2026-07-30
feat(os-events): add /api/os/events SSE endpoint and useOsEvents hook

jaylfc:
Autonomous build of board card tsk-yh36f7.

- New authenticated SSE endpoint at GET /api/os/events streams typed
  change events (kind + id only, never payload) filtered by kinds query param
- New shared client hook useOsEvents(kinds, onEvent) returns connected/stale
  flags, manages single per-client EventSource with exponential backoff
- Reuses existing EventBus SSE plumbing and auth middleware
- Backend tests cover auth gate, kind filtering, and multi-subscriber delivery
- Frontend tests cover hook mount, kind filtering, dedup, stale flag, and reconnect

Files:
 CHANGELOG.md                            |   9 +
 desktop/src/hooks/use-os-events.test.ts | 228 ++++++++++++++++++++++++++
 desktop/src/hooks/use-os-events.ts      | 113 +++++++++++++
 tests/test_os_events.py                 | 282 ++++++++++++++++++++++++++++++++
 tinyagentos/routes/os_events.py         | 121 ++++++++++++++
 6 files changed, 756 insertions(+)

## Summary by CodeRabbit

* **New Features**
  * Added an authenticated real-time OS event stream at `/api/os/events`.
  * Events can be filtered by type and include identifiers and timestamps without exposing payload details.
  * Added client-side connection status reporting, event deduplication, and automatic reconnection with backoff.
* **Documentation**
  * Documented the new event stream and client integration in the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
 
 ## [Unreleased]
 
+### Added
+
+- **OS-level typed change-event stream + `useOsEvents` hook.** A new authenticated
+  SSE endpoint (`GET /api/os/events`) streams typed change events carrying only
+  the event kind and id, never the payload, so apps can opt into live updates
+  with a single hook call. The shared `useOsEvents(kinds, onEvent)` hook manages
+  the single per-client connection, exposes `connected` and `stale` flags, and
+  handles reconnect with exponential backoff.
+
 ### Fixed
 
 - **Push notifications never routed to the correct app.** Three independent faults
--- a/desktop/src/hooks/use-os-events.test.ts
+++ b/desktop/src/hooks/use-os-events.test.ts
@@ -1,0 +1,228 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { renderHook, act } from "@testing-library/react";
+import { useOsEvents, OsEvent } from "./use-os-events";
+
+// ---------------------------------------------------------------------------
+// Mock EventSource -- must use a regular function (not arrow) so `new` works
+// ---------------------------------------------------------------------------
+
+type MessageListener = (e: MessageEvent) => void;
+
+interface MockEventSource {
+  url: string;
+  onopen: (() => void) | null;
+  onmessage: MessageListener | null;
+  onerror: ((e: Event) => void) | null;
+  close: ReturnType<typeof vi.fn>;
+  readyState: number;
+  _fire: (data: unknown) => void;
+  _fireError: () => void;
+}
+
+let lastEs: MockEventSource | null = null;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const MockEventSourceCtor = vi.fn().mockImplementation(function (this: any, url: string) {
+  this.url = url;
+  this.onopen = null;
+  this.onmessage = null;
+  this.onerror = null;
+  this.close = vi.fn();
+  this.readyState = 0; // CONNECTING
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  this._fire = (data: unknown) => {
+    (this as MockEventSource).onmessage?.({ data: JSON.stringify(data) } as MessageEvent);
+  };
+  this._fireError = () => {
+    (this as MockEventSource).onerror?.(new Event("error"));
+  };
+  lastEs = this as MockEventSource;
+});
+
+beforeEach(() => {
+  vi.stubGlobal("EventSource", MockEventSourceCtor);
+  MockEventSourceCtor.mockClear();
+  lastEs = null;
+});
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+});
+
+describe("useOsEvents", () => {
+  it("opens an EventSource to /api/os/events with kinds on mount", () => {
+    renderHook(() => useOsEvents(["projects.task.changed"], () => {}));
+    expect(MockEventSourceCtor).toHaveBeenCalledWith(
+      "/api/os/events?kinds=projects.task.changed",
+    );
+  });
+
+  it("opens an EventSource without kinds when kinds is empty", () => {
+    renderHook(() => useOsEvents([], () => {}));
+    expect(MockEventSourceCtor).toHaveBeenCalledWith("/api/os/events");
+  });
+
+  it("calls onEvent for matching kind", () => {
+    const onEvent = vi.fn();
+    renderHook(() => useOsEvents(["projects.task.changed"], onEvent));
+
+    act(() => {
+      lastEs?._fire({
+        kind: "projects.task.changed",
+        id: "evt-1",
+        ts: 1234567890.0,
+      });
+    });
+
+    expect(onEvent).toHaveBeenCalledWith({
+      kind: "projects.task.changed",
+      id: "evt-1",
+      ts: 1234567890.0,
+    });
+  });
+
+  it("ignores non-matching kind", () => {
+    const onEvent = vi.fn();
+    renderHook(() => useOsEvents(["projects.task.changed"], onEvent));
+
+    act(() => {
+      lastEs?._fire({
+        kind: "agents.status.changed",
+        id: "evt-2",
+        ts: 1234567890.0,
+      });
+    });
+
+    expect(onEvent).not.toHaveBeenCalled();
+  });
+
+  it("sets connected=true and stale=false on open", async () => {
+    const { result } = renderHook(() =>
+      useOsEvents(["projects.task.changed"], () => {}),
+    );
+
+    act(() => {
+      lastEs?.onopen?.();
+    });
+
+    expect(result.current.connected).toBe(true);
+    expect(result.current.stale).toBe(false);
+  });
+
+  it("sets connected=false and stale=true on hard disconnect", async () => {
+    vi.useFakeTimers();
+    const { result } = renderHook(() =>
+      useOsEvents(["projects.task.changed"], () => {}),
+    );
+
+    act(() => {
+      lastEs?.onopen?.();
+    });
+    expect(result.current.connected).toBe(true);
+    expect(result.current.stale).toBe(false);
+
+    act(() => {
+      if (lastEs) {
+        lastEs.readyState = EventSource.CLOSED;
+        lastEs._fireError();
+      }
+    });
+
+    expect(result.current.connected).toBe(false);
+    expect(result.current.stale).toBe(true);
+
+    vi.useRealTimers();
+  });
+
+  it("closes the EventSource on unmount", () => {
+    const { unmount } = renderHook(() =>
+      useOsEvents(["projects.task.changed"], () => {}),
+    );
+    unmount();
+    expect(lastEs?.close).toHaveBeenCalled();
+  });
+
+  it("deduplicates events by id", () => {
+    const onEvent = vi.fn();
+    renderHook(() => useOsEvents(["projects.task.changed"], onEvent));
+
+    act(() => {
+      lastEs?._fire({
+        kind: "projects.task.changed",
+        id: "evt-dup",
+        ts: 1234567890.0,
+      });
+      lastEs?._fire({
+        kind: "projects.task.changed",
+        id: "evt-dup",
+        ts: 1234567890.0,
+      });
+    });
+
+    expect(onEvent).toHaveBeenCalledTimes(1);
+  });
+
+  it("reconnects after drop and receives new events", async () => {
+    vi.useFakeTimers();
+    const onEvent = vi.fn();
+    const { result } = renderHook(() =>
+      useOsEvents(["projects.task.changed"], onEvent),
+    );
+
+    act(() => {
+      lastEs?.onopen?.();
+    });
+    expect(result.current.connected).toBe(true);
+    expect(result.current.stale).toBe(false);
+
+    act(() => {
+      if (lastEs) {
+        lastEs.readyState = EventSource.CLOSED;
+        lastEs._fireError();
+      }
+    });
+    expect(result.current.connected).toBe(false);
+    expect(result.current.stale).toBe(true);
+
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(5000);
+    });
+
+    expect(MockEventSourceCtor).toHaveBeenCalledTimes(2);
+    expect(lastEs).not.toBeNull();
+    act(() => {
+      lastEs?.onopen?.();
+    });
+    expect(result.current.connected).toBe(true);
+    expect(result.current.stale).toBe(false);
+
+    act(() => {
+      lastEs?._fire({
+        kind: "projects.task.changed",
+        id: "evt-reconnect",
+        ts: 1234567890.0,
+      });
+    });
+    expect(onEvent).toHaveBeenCalledWith({
+      kind: "projects.task.changed",
+      id: "evt-reconnect",
+      ts: 1234567890.0,
+    });
+
+    vi.useRealTimers();
+  });
+
+  it("ignores events with no kind", () => {
+    const onEvent = vi.fn();
+    renderHook(() => useOsEvents(["projects.task.changed"], onEvent));
+
+    act(() => {
+      lastEs?._fire({
+        id: "evt-nokind",
+        ts: 1234567890.0,
+      });
+    });
+
+    expect(onEvent).not.toHaveBeenCalled();
+  });
+});
--- a/desktop/src/hooks/use-os-events.ts
+++ b/desktop/src/hooks/use-os-events.ts
@@ -1,0 +1,113 @@
+import { useEffect, useState, useCallback, useRef } from "react";
+
+export type OsEvent = {
+  kind: string;
+  id: string;
+  ts: number;
+};
+
+type OsEventHandler = (event: OsEvent) => void;
+
+const RECONNECT_DELAY_MS = 5000;
+const MAX_RECONNECT_DELAY_MS = 30000;
+const MAX_SEEN_IDS = 128;
+
+export function useOsEvents(kinds: string[], onEvent: OsEventHandler): {
+  connected: boolean;
+  stale: boolean;
+} {
+  const [connected, setConnected] = useState(false);
+  const [stale, setStale] = useState(true);
+  const reconnectTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const reconnectAttemptsRef = useRef(0);
+  const stoppedRef = useRef(false);
+  const esRef = useRef<EventSource | null>(null);
+  const seenIdsRef = useRef<string[]>([]);
+  const seenRef = useRef(new Set<string>());
+  const onEventRef = useRef(onEvent);
+  const kindsRef = useRef(kinds);
+
+  useEffect(() => {
+    onEventRef.current = onEvent;
+  }, [onEvent]);
+
+  useEffect(() => {
+    kindsRef.current = kinds;
+  }, [kinds]);
+
+  const connect = useCallback(() => {
+    if (stoppedRef.current) return;
+
+    const kindsParam = kindsRef.current.join(",");
+    const url = kindsParam
+      ? `/api/os/events?kinds=${encodeURIComponent(kindsParam)}`
+      : "/api/os/events";
+
+    const es = new EventSource(url);
+    esRef.current = es;
+
+    const alreadySeen = (id: string | undefined): boolean => {
+      if (!id) return false;
+      if (seenRef.current.has(id)) return true;
+      seenRef.current.add(id);
+      seenIdsRef.current.push(id);
+      if (seenIdsRef.current.length > MAX_SEEN_IDS) {
+        const oldest = seenIdsRef.current.shift();
+        if (oldest) seenRef.current.delete(oldest);
+      }
+      return false;
+    };
+
+    es.onopen = () => {
+      reconnectAttemptsRef.current = 0;
+      setConnected(true);
+      setStale(false);
+    };
+
+    es.onmessage = (msg) => {
+      let event: OsEvent | null;
+      try {
+        event = JSON.parse(msg.data as string) as OsEvent;
+      } catch {
+        return;
+      }
+      if (!event || typeof event !== "object") return;
+      if (!event.kind) return;
+      if (kindsRef.current.length > 0 && !kindsRef.current.includes(event.kind)) {
+        return;
+      }
+      if (alreadySeen(event.id)) return;
+      onEventRef.current(event);
+    };
+
+    es.onerror = () => {
+      if (!stoppedRef.current && es.readyState === EventSource.CLOSED) {
+        const delay = Math.min(
+          RECONNECT_DELAY_MS * 2 ** reconnectAttemptsRef.current,
+          MAX_RECONNECT_DELAY_MS,
+        );
+        reconnectAttemptsRef.current += 1;
+        setConnected(false);
+        setStale(true);
+        reconnectTimerRef.current = setTimeout(() => {
+          if (!stoppedRef.current) connect();
+        }, delay);
+      }
+    };
+  }, []);
+
+  useEffect(() => {
+    stoppedRef.current = false;
+    connect();
+
+    return () => {
+      stoppedRef.current = true;
+      if (reconnectTimerRef.current) clearTimeout(reconnectTimerRef.current);
+      esRef.current?.close();
+      setConnected(false);
+      setStale(true);
+    };
+  }, [connect]);
+
+  return { connected, stale };
+}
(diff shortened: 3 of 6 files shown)


jaylfc:
**nemotron-ultra-orB review** 

VERDICT: Significant issues found — manual reconnection conflicts with browser EventSource auto-reconnect, potential resource leaks, and test gaps.

- **desktop/src/hooks/use-os-events.ts:68** — Manual reconnection logic conflicts with browser's built-in EventSource auto-reconnect. The `onerror` handler schedules a custom `setTimeout` reconnect while the browser simultaneously attempts its own reconnect, potentially creating duplicate connections. Either disable browser reconnect (by calling `close()` in onerror) or rely on native behavior.

- **desktop/src/hooks/use-os-events.ts:92** — `esRef.current?.close()` in cleanup only closes the latest EventSource. If a reconnect created a new EventSource before cleanup, the previous one may leak (not closed, no reference).

- **tinyagentos/routes/os_events.py:87** — `event.trace_id` used as SSE `id` field but `SystemEvent` type not visible in diff; test assumes `id` exists. If `trace_id` is missing/None, SSE `id:` will be `null`, breaking client resume logic.

- **tests/test_os_events.py:67** — Low-level ASGI scope testing bypasses normal middleware/auth flow. `test_stream_requires_auth` uses test client correctly but other tests manually construct scope, missing middleware validation (CSRF, auth cookie parsing).

- **tests/test_os_events.py** — No test for: keepalive (`:keepalive\n\n` every 10s), comma-separated `kinds` param, empty `kinds`=all, 503 when event_bus unavailable, SSE `id:` field sequencing, or reconnect/resume via Last-Event-ID.

- **desktop/src/hooks/use-os-events.test.ts:185** — Reconnect test only verifies first attempt (5000ms). Exponential backoff (2^n * 5000) and max cap (30000) untested.

- **desktop/src/hooks/use-os-events.ts:54** — Deduplication uses both `Set` and array (`seenRef` + `seenIdsRef`) redundantly. LRU eviction logic correct but overcomplicated; single `Map<id, timestamp>` with size limit would be cleaner.

- **tinyagentos/routes/os_events.py:74** — `_relay` tasks run unbounded until cancelled. If `merged.get()` blocks indefinitely (no events), tasks never exit cleanly on disconnect — relies solely on `finally` block cancellation which is correct but fragile if generator throws before `finally`.

jaylfc:
**nemotron-super review** 

VERDICT: Pass
No blocking issues found.

jaylfc:

Now let me analyze each file from the diff:

## 1. `desktop/src/hooks/use-os-events.ts` - Hook Implementation

**Issues found:**

1. **Line 54-57**: `seenIdsRef` and `seenRef` are both used for deduplication but the logic is inconsistent. `alreadySeen` adds to both but `seenIdsRef` is an array while `seenRef` is a Set. When removing oldest, it deletes from Set but the array shift doesn't correspond to the same element necessarily (though in FIFO it should). This is confusing dual-structure.

2. **Line 74**: `kindsRef.current.includes(event.kind)` - This is O(n) per event. Should use a Set for O(1) lookup since kinds don't change often.

3. **Line 88**: `es.readyState === EventSource.CLOSED` check in `onerror` - The `EventSource` spec says `onerror` fires for both recoverable and fatal errors. Checking `readyState === CLOSED` is correct for detecting fatal disconnect, but the reconnect logic doesn't distinguish between network errors (recoverable) and explicit close. However, browser EventSource auto-reconnects on its own for recoverable errors, so this might be intentional.

4. **Line 94**: `reconnectTimerRef.current = setTimeout(() => { if (!stoppedRef.current) connect(); }, delay);` - The `connect` function is memoized with `useCallback([], [])` (empty deps), so it's stable. However, `connect` closes over `kindsRef` and `onEventRef` which are refs, so this is fine.

5. **Line 100-107**: Cleanup in `useEffect` sets `stoppedRef.current = true` then clears timeout and closes ES. But there's a race: if `onerror` fires after `stoppedRef.current = true` but before `esRef.current?.close()`, the `onerror` handler checks `!stoppedRef.current` and won't reconnect. This is correct.

6. **Missing**: No cleanup of `seenRef`/`seenIdsRef` on unmount - memory grows unbounded if component remounts with different kinds. Not a bug per se since refs persist, but could accumulate many IDs over app lifetime.

7. **Line 33**: `MAX_SEEN_IDS = 128` - Arbitrary limit. Should be documented why 128.

## 2. `desktop/src/hooks/use-os-events.test.ts` - Tests

**Issues found:**

1. **Line 149**: `act(() => { lastEs?.onopen?.(); });` - The test calls `onopen` synchronously but `EventSource` onopen is async in real browsers. The hook's `onopen` handler uses `setConnected(true)` and `setStale(false)` which are state updates - these are batched in React 18, so the test might pass but doesn't verify async behavior.

2. **Line 168**: `lastEs.readyState = EventSource.CLOSED;` - Directly mutating `readyState` on the mock. In real EventSource, `readyState` is read-only. The test simulates a hard disconnect by setting `readyState = CLOSED` then firing error. This is a reasonable test simulation but doesn't match real browser behavior exactly.

3. **Line 187**: `await act(async () => { await vi.advanceTimersByTimeAsync(5000); });` - Uses fake timers for reconnect delay test. The delay is `RECONNECT_DELAY_MS * 2 ** reconnectAttemptsRef.current` = 5000 * 2^0 = 5000ms. Test advances by 5000ms - correct for first attempt.

4. **Line 197**: `expect(MockEventSourceCtor).toHaveBeenCalledTimes(2);` - Verifies reconnect creates new EventSource. Good.

5. **Missing test cases**:
   - No test for `kinds` prop change (should reconnect with new kinds)
   - No test for `onEvent` prop change (should use new callback)
   - No test for rapid connect/disconnect cycles
   - No test for event ordering preservation
   - No test for keepalive handling (though hook doesn't handle keepalive - server sends `:keepalive`)

6. **Line 217-225**: `ignores events with no kind` - Good edge case test.

## 3. `tinyagentos/routes/os_events.py` - SSE Endpoint

**Critical Issues:**

1. **Line 84-85**: `user_q = await event_bus.subscribe(user_ch)` and `bcast_q = await event_bus.subscribe("broadcast")` - Subscribes to both user channel and broadcast. But there's no authorization check on what channels the user can access. Any authenticated user can subscribe to `user:<any_id>` by manipulating the request? No - `user_id` comes from `request.state.user_id` set by auth middleware, so this is OK.

2. **Line 94-98**: Relay tasks created with `asyncio.create_task(_relay(...))` but `_relay` runs infinite loop. If `merged.get()` blocks forever and client disconnects, the `finally` block cancels tasks. However, `asyncio.wait_for(merged.get(), timeout=10.0)` with 10s timeout means keepalive every 10s. If client disconnects, `request.is_disconnected()` check at loop top catches it. This is correct.

3. **Line 107**: `event.trace_id` used as `id` field in SSE. But `SystemEvent` (from events/bus.py) - need to check if it has `trace_id`. The test shows `event.trace_id` being used. Looking at test line 74: `assert evt["id"] is not None` - the test expects `id` to exist. But `SystemEvent` structure isn't shown. If `SystemEvent` doesn't have `trace_id`, this will fail. The test mocks `SystemEvent` with `kind`, `source`, `targets`, `payload` - no `trace_id` or `ts` shown. Wait, test line 68-74 creates `SystemEvent` with those 4 fields but then asserts `evt["id"] is not None` and `evt["ts"] is not None`. This suggests `EventBus.subscribe()` returns events with `trace_id` and `ts` added. Need to verify `EventBus` implementation.

4. **Line 101**: `allowed_kinds = {k.strip() for k in kinds_param.split(",") if k.strip()}` - Empty kinds_param gives `None` (line 88-90). But if `kinds_param` is empty string `""`, then `kinds_param` is truthy, `split(",")` gives `[""]`, filtered to empty set, so `allowed_kinds = set()`. This means empty kinds= filters ALL events! Should treat empty string same as missing. **BUG**: `kinds=` (empty) vs no `kinds` param behave differently.

5. **Line 110**: `yield f"id: {seq}\ndata: {data}\n\n"` - Uses incrementing `seq` as SSE `id`. This is for Last-Event-ID resume, but the comment says "not precise replay via Last-Event-ID". The `seq` is per-connection, not global. If client reconnects with Last-Event-ID, server would need to map seq to event. Not implemented. This is documented as best-effort.

6. **Line 72**: `user_id = getattr(request.state, "user_id", None)` - If auth middleware fails to set user_id, returns 401. Good belt-and-braces.

7. **Line 75-77**: Returns 503 if no event_bus. Good.

8. **Line 116-119**: Headers include `Cache-Control: no-cache`, `X-Accel-Buffering: no`, `Connection: keep-alive`. Correct for SSE.

9. **Missing**: No handling of `Last-Event-ID` header for resume. Documented as not implemented.

10. **Potential resource leak**: If `gen()` generator is cancelled before `finally` (e.g., exception in `json.dumps`), cleanup runs via `finally`. But if `asyncio.wait_for` timeout fires and yields keepalive, then client alive indefinitely. This is by design.

## 4. `tests/test_os_events.py` - Python Tests

**Issues found:**

1. **Line 44-53**: `test_stream_requires_auth` - Creates app without auth middleware, expects 401. But the handler checks `request.state.user_id` which would be missing. The test doesn't add the auth middleware that sets `user_id`. Actually, the handler returns 401 if `user_id` is None. Without middleware, `request.state` might not have `user_id` attribute. `getattr(request.state, "user_id", None)` returns None. So 401 is correct. But the test app doesn't include the middleware that would normally reject unauthenticated requests before reaching handler. The handler itself does the check. This tests the handler's explicit check.

2. **Line 55-63**: `test_stream_requires_auth_explicit_setup` - Mocks request directly, calls handler function. Good unit test.

3. **Line 66-110**: `test_stream_delivers_subscribed_kinds` - Complex ASGI test. Creates scope manually. Uses `asyncio.sleep(0.2)` to let subscription settle - **flaky timing-dependent test**. Should use proper synchronization.

4. **Line 88**: `await bus.broadcast(SystemEvent(...))` - Broadcasts event. The test expects it to arrive via SSE. But the test subscribes to `kinds=projects.task.changed` and broadcasts same kind. Good.

5. **Line 102**: `assert "payload" not in evt` - Verifies payload is stripped. Good.

6. **Line 113-162**: `test_stream_filters_kinds` - Similar structure, expects NO events for wrong kind. Uses `asyncio.wait_for(got_wrong_kind.wait(), timeout=1.0)` expecting timeout. **Flaky**: 1 second might not be enough on slow CI. Should use longer timeout or better synchronization.

7. **Line 165-220**: `test_stream_two_subscribers_both_receive` - Tests broadcast to multiple clients. Good.

8. **Missing tests**:
   - No test for `kinds` parameter with multiple kinds
   - No test for empty `kinds=` (should behave as "all" but currently filters all - see bug above)
   - No test for unauthenticated request with middleware (integration test)
   - No test for reconnect/resume with Last-Event-ID
   - No test for keepalive format
   - No test for event ordering
   - No test for cleanup on disconnect (unsubscribe called)

## 5. `tinyagentos/routes/__init__.py` - Router Registration

**Line 343-345**: Adds `os_events_router` with `_csrf` dependencies. The `_csrf` is likely CSRF protection. For SSE endpoint, CSRF is not typically needed (GET request, no state change). But including it doesn't hurt if it's just a dependency that validates session. Need to check what `_csrf` does. If it's a dependency that raises on invalid CSRF, it might break SSE (since EventSource can't send custom headers). **Potential issue**: If `_csrf` requires a header/token, EventSource connections will fail.

Let me check the diff context - line 340 shows `event_stream_router` also uses `_csrf`. So this is consistent with existing pattern.

---

Now let me compile the final verdict.
Now let me analyze each file from the diff:

## 1. `desktop/src/hooks/use-os-events.ts` - Hook Implementation

**Issues found:**

1. **Line 54-57**: `seenIdsRef` and `seenRef` are both used for deduplication but the logic is inconsistent. `alreadySeen` adds to both but `seenIdsRef` is an array while `seenRef` is a Set. When removing oldest, it deletes from Set but the array shift doesn't correspond to the same element necessarily (though in FIFO it should). This is confusing dual-structure.

2. **Line 74**: `kindsRef.current.includes(event.kind)` - This is O(n) per event. Should use a Set for O(1) lookup since kinds don't change often.

3. **Line 88**: `es.readyState === EventSource.CLOSED` check in `onerror` - The `EventSource` spec says `onerror` fires for both recoverable and fatal errors. Checking `readyState === CLOSED` is correct for detecting fatal disconnect, but the reconnect logic doesn't distinguish between network errors (recoverable) and explicit close. However, browser EventSource auto-reconnects on its own for recoverable errors, so this might be intentional.

4. **Line 94**: `reconnectTimerRef.current = setTimeout(() => { if (!stoppedRef.current) connect(); }, delay);` - The `connect` function is memoized with `useCallback([], [])` (empty deps), so it's stable. However, `connect` closes over `kindsRef` and `onEventRef` which are refs, so this is fine.

5. **Line 100-107**: Cleanup in `useEffect` sets `stoppedRef.current = true` then clears timeout and closes ES. But there's a race: if `onerror` fires after `stoppedRef.current = true` but before `esRef.current?.close()`, the `onerror` handler checks `!stoppedRef.current` and won't reconnect. This is correct.

6. **Missing**: No cleanup of `seenRef`/`seenIdsRef` on unmount - memory grows unbounded if component remounts with different kinds. Not a bug per se since refs persist, but could accumulate many IDs over app lifetime.

7. **Line 33**: `MAX_SEEN_IDS = 128` - Arbitrary limit. Should be documented why 128.

## 2. `desktop/src/hooks/use-os-events.test.ts` - Tests

**Issues found:**

1. **Line 149**: `act(() => { lastEs?.onopen?.(); });` - The test calls `onopen` synchronously but `EventSource` onopen is async in real browsers. The hook's `onopen` handler uses `setConnected(true)` and `setStale(false)` which are state updates - these are batched in React 18, so the test might pass but doesn't verify async behavior.

2. **Line 168**: `lastEs.readyState = EventSource.CLOSED;` - Directly mutating `readyState` on the mock. In real EventSource, `readyState` is read-only. The test simulates a hard disconnect by setting `readyState = CLOSED` then firing error. This is a reasonable test simulation but doesn't match real browser behavior exactly.

3. **Line 187**: `await act(async () => { await vi.advanceTimersByTimeAsync(5000); });` - Uses fake timers for reconnect delay test. The delay is `RECONNECT_DELAY_MS * 2 ** reconnectAttemptsRef.current` = 5000 * 2^0 = 5000ms. Test advances by 5000ms - correct for first attempt.

4. **Line 197**: `expect(MockEventSourceCtor).toHaveBeenCalledTimes(2);` - Verifies reconnect creates new EventSource. Good.

5. **Missing test cases**:
   - No test for `kinds` prop change (should reconnect with new kinds)
   - No test for `onEvent` prop change (should use new callback)
   - No test for rapid connect/disconnect cycles
   - No test for event ordering preservation
   - No test for keepalive handling (though hook doesn't handle keepalive - server sends `:keepalive`)

6. **Line 217-225**: `ignores events with no kind` - Good edge case test.

## 3. `tinyagentos/routes/os_events.py` - SSE Endpoint

**Critical Issues:**

1. **Line 84-85**: `user_q = await event_bus.subscribe(user_ch)` and `bcast_q = await event_bus.subscribe("broadcast")` - Subscribes to both user channel and broadcast. But there's no authorization check on what channels the user can access. Any authenticated user can subscribe to `user:<any_id>` by manipulating the request? No - `user_id` comes from `request.state.user_id` set by auth middleware, so this is OK.

2. **Line 94-98**: Relay tasks created with `asyncio.create_task(_relay(...))` but `_relay` runs infinite loop. If `merged.get()` blocks forever and client disconnects, the `finally` block cancels tasks. However, `asyncio.wait_for(merged.get(), timeout=10.0)` with 10s timeout means keepalive every 10s. If client disconnects, `request.is_disconnected()` check at loop top catches it. This is correct.

3. **Line 107**: `event.trace_id` used as `id` field in SSE. But `SystemEvent` (from events/bus.py) - need to check if it has `trace_id`. The test shows `event.trace_id` being used. Looking at test line 74: `assert evt["id"] is not None` - the test expects `id` to exist. But `SystemEvent` structure isn't shown. If `SystemEvent` doesn't have `trace_id`, this will fail. The test mocks `SystemEvent` with `kind`, `source`, `targets`, `payload` - no `trace_id` or `ts` shown. Wait, test line 68-74 creates `SystemEvent` with those 4 fields but then asserts `evt["id"] is not None` and `evt["ts"] is not None`. This suggests `EventBus.subscribe()` returns events with `trace_id` and `ts` added. Need to verify `EventBus` implementation.

4. **Line 101**: `allowed_kinds = {k.strip() for k in kinds_param.split(",") if k.strip()}` - Empty kinds_param gives `None` (line 88-90). But if `kinds_param` is empty string `""`, then `kinds_param` is truthy, `split(",")` gives `[""]`, filtered to empty set, so `allowed_kinds = set()`. This means empty kinds= filters ALL events! Should treat empty string same as missing. **BUG**: `kinds=` (empty) vs no `kinds` param behave differently.

5. **Line 110**: `yield f"id: {seq}\ndata: {data}\n\n"` - Uses incrementing `seq` as SSE `id`. This is for Last-Event-ID resume, but the comment says "not precise replay via Last-Event-ID". The `seq` is per-connection, not global. If client reconnects with Last-Event-ID, server would need to map seq to event. Not implemented. This is documented as best-effort.

6. **Line 72**: `user_id = getattr(request.state, "user_id", None)` - If auth middleware fails to set user_id, returns 401. Good belt-and-braces.

7. **Line 75-77**: Returns 503 if no event_bus. Good.

8. **Line 116-119**: Headers include `Cache-Control: no-cache`, `X-Accel-Buffering: no`, `Connection: keep-alive`. Correct for SSE.

9. **Missing**: No handling of `Last-Event-ID` header for resume. Documented as not implemented.

10. **Potential resource leak**: If `gen()` generator is cancelled before `finally` (e.g., exception in `json.dumps`), cleanup runs via `finally`. But if `asyncio.wait_for` timeout fires and yields keepalive, then client alive indefinitely. This is by design.

## 4. `tests/test_os_events.py` - Python Tests

**Issues found:**

1. **Line 44-53**: `test_stream_requires_auth` - Creates app without auth middleware, expects 401. But the handler checks `request.state.user_id` which would be missing. The test doesn't add the auth middleware that sets `user_id`. Actually, the handler returns 401 if `user_id` is None. Without middleware, `request.state` might not have `user_id` attribute. `getattr(request.state, "user_id", None)` returns None. So 401 is correct. But the test app doesn't include the middleware that would normally reject unauthenticated requests before reaching handler. The handler itself does the check. This tests the handler's explicit check.

2. **Line 55-63**: `test_stream_requires_auth_explicit_setup` - Mocks request directly, calls handler function. Good unit test.

3. **Line 66-110**: `test_stream_delivers_subscribed_kinds` - Complex ASGI test. Creates scope manually. Uses `asyncio.sleep(0.2)` to let subscription settle - **flaky timing-dependent test**. Should use proper synchronization.

4. **Line 88**: `await bus.broadcast(SystemEvent(...))` - Broadcasts event. The test expects it to arrive via SSE. But the test subscribes to `kinds=projects.task.changed` and broadcasts same kind. Good.

5. **Line 102**: `assert "payload" not in evt` - Verifies payload is stripped. Good.

6. **Line 113-162**: `test_stream_filters_kinds` - Similar structure, expects NO events for wrong kind. Uses `asyncio.wait_for(got_wrong_kind.wait(), timeout=1.0)` expecting timeout. **Flaky**: 1 second might not be enough on slow CI. Should use longer timeout or better synchronization.

7. **Line 165-220**: `test_stream_two_subscribers_both_receive` - Tests broadcast to multiple clients. Good.

8. **Missing tests**:
   - No test for `kinds` parameter with multiple kinds
   - No test for empty `kinds=` (should behave as "all" but currently filters all - see bug above)
   - No test for unauthenticated request with middleware (integration test)
   - No test for reconnect/resume with Last-Event-ID
   - No test for keepalive format
   - No test for event ordering
   - No test for cleanup on disconnect (unsubscribe called)

## 5. `tinyagentos/routes/__init__.py` - Router Registration

**Line 343-345**: Adds `os_events_router` with `_csrf` dependencies. The `_csrf` is likely CSRF protection. For SSE endpoint, CSRF is not typically needed (GET request, no state change). But including it doesn't hurt if it's just a dependency that validates session. Need to check what `_csrf` does. If it's a dependency that raises on invalid CSRF, it might break SSE (since EventSource can't send custom headers). **Potential issue**: If `_csrf` requires a header/token, EventSource connections will fail.

Let me check the diff context - line 340 shows `event_stream_router` also uses `_csrf`. So this is consistent with existing pattern.

---

Now let me compile the final verdict.
VERDICT: NEEDS FIXES — Two blocking bugs and several test gaps.

- tinyagentos/routes/os_events.py:88-90 — Empty `kinds=` query param (e.g. `/api/os/events?kinds=`) produces `allowed_kinds = set()` filtering ALL events, while missing `kinds` param means "all kinds". Empty string should be treated as absent.
- desktop/src/hooks/use-os-events.ts:54-57 — Dual deduplication structures (`seenRef` Set + `seenIdsRef` array) with manual sync; removal from Set uses `shift()` from array but Set has no order guarantee. Use single `Map<id, timestamp>` or ordered `Set` with size limit.
- tinyagentos/routes/__init__.py:343-345 — SSE endpoint registered with `_csrf` dependency; EventSource cannot send custom headers/cookies in all browsers (Safari), may break SSE connections if CSRF enforces header check.
- desktop/src/hooks/use-os-events.test.ts — Missing tests for: `kinds` prop change (should reconnect), `onEvent` prop change, rapid reconnect cycles, event ordering, keepalive handling.
- tests/test_os_events.py:88,145 — Flaky `asyncio.sleep(0.2)` and 1s timeout in filter test; use `asyncio.Event` for subscription-ready sync and longer/more generous timeouts.
- tests/test_os_events.py — Missing tests for: multiple `kinds` filter, empty `kinds=` bug, Last-Event-ID resume, keepalive format, disconnect cleanup verification.
- desktop/src/hooks/use-os-events.ts:74 — `kindsRef.current.includes(event.kind)` is O(n); use `Set` for O(1) lookup.
- desktop/src/hooks/use-os-events.ts:33 — `MAX_SEEN_IDS = 128` undocumented; add comment with rationale.

No blocking issues in CHANGELOG.md.
VERDICT: NEEDS FIXES — Two blocking bugs and several test gaps.

- tinyagentos/routes/os_events.py:88-90 — Empty `kinds=` query param (e.g. `/api/os/events?kinds=`) produces `allowed_kinds = set()` filtering ALL events, while missing `kinds` param means "all kinds". Empty string should be treated as absent.
- desktop/src/hooks/use-os-events.ts:54-57 — Dual deduplication structures (`seenRef` Set + `seenIdsRef` array) with manual sync; removal from Set uses `shift()` from array but Set has no order guarantee. Use single `Map<id, timestamp>` or ordered `Set` with size limit.
- tinyagentos/routes/__init__.py:343-345 — SSE endpoint registered with `_csrf` dependency; EventSource cannot send custom headers/cookies in all browsers (Safari), may break SSE connections if CSRF enforces header check.
- desktop/src/hooks/use-os-events.test.ts — Missing tests for: `kinds` prop change (should reconnect), `onEvent` prop change, rapid reconnect cycles, event ordering, keepalive handling.
- tests/test_os_events.py:88,145 — Flaky `asyncio.sleep(0.2)` and 1s timeout in filter test; use `asyncio.Event` for subscription-ready sync and longer/more generous timeouts.
- tests/test_os_events.py — Missing tests for: multiple `kinds` filter, empty `kinds=` bug, Last-Event-ID resume, keepalive format, disconnect cleanup verification.
- desktop/src/hooks/use-os-events.ts:74 — `kindsRef.current.includes(event.kind)` is O(n); use `Set` for O(1) lookup.
- desktop/src/hooks/use-os-events.ts:33 — `MAX_SEEN_IDS = 128` undocumented; add comment with rationale.

No blocking issues in CHANGELOG.md.

jaylfc:
Closing as **superseded by #2309**, and because this PR is additionally conflicted (`DIRTY`) against a base that has moved.

**Verified on origin, not locally:**

- `dev` and `master` carry no `os-events` implementation, so this feature has not landed by any route.
- #2309 is branched from `dev` and its diff adds the same files this PR does — `tinyagentos/routes/os_events.py`, `desktop/src/hooks/use-os-events.ts`, `tests/test_os_events.py`, `tinyagentos/routes/__init__.py` — plus the doc-gate and hook fixes this one was missing.

So #2309 is a strict superset: the whole feature plus the corrections. Merging both would not compose, since they touch the same lines of the same files, and this one cannot merge as it stands anyway.

**One operational reason for doing this now.** There are 9 open `exec/*` PRs against a dispatcher cap of 8, so the lane throttle has been engaged continuously and no new card has dispatched since 13:04Z — roughly 8 hours — while 137 claimable cards sit in the pool. A redundant conflicted PR occupying a throttle slot costs real throughput, not just tidiness.

Card `tsk-yh36f7` is superseded by `tsk-jmctoa`. Reopen if #2309 is abandoned rather than merged.

PR #2226, opened 2026-08-02
fix(deleted-symbols): replace predicate with base_tip - HEAD, drop merge-base, rewrite tests for merge-result model

jaylfc:
Autonomous build of board card tsk-ct2c3z.

- compute signal = symbols(base_tip) - symbols(HEAD); remove _merge_base
  and mb_symbols entirely
- keep the merge-ref checkout and Removes-Intentionally waiver
- rewrite integration tests to model the merge result (merge PR into
  base, run there) instead of checking out the PR head
- invert test_no_signal_when_dev_unchanged_since_merge_base to expect
  violations when base code is lost in the merge result
- add probe-3 test: PR deletes a long-merged test file, gate must FAIL
- fix _find_adding_commit to search base history via git log -S instead
  of the removed mb..base range
- add permissions contents:read on the job and move github.base_ref into
  env in the workflow

Files:
 .github/workflows/deleted-symbols-gate.yml |  41 +++
 scripts/check_deleted_symbols.py           | 236 +++++++++++++++
 tests/test_check_deleted_symbols.py        | 442 +++++++++++++++++++++++++++++
 3 files changed, 719 insertions(+)

### DIFF
--- a/.github/workflows/deleted-symbols-gate.yml
+++ b/.github/workflows/deleted-symbols-gate.yml
@@ -1,0 +1,41 @@
+name: Deleted symbols gate
+
+# Detects PRs that silently delete Python symbols (def/class/test names) that
+# landed on the target branch after the PR's merge base. A PR cut before
+# hardening commits landed on dev would silently delete them when merged --
+# git reports "Automatic merge went well" with no conflict because the PR
+# branch simply wins on files dev touched after the branch point.
+#
+# This check fails such PRs and names the deleted symbols and the commits that
+# added them. A "Removes-Intentionally: <symbol>" trailer in the PR body
+# waives named symbols, making deliberate deletions a conscious, auditable act.
+#
+# See scripts/check_deleted_symbols.py for the implementation.
+
+on:
+  pull_request:
+    branches: [master, dev]
+
+jobs:
+  deleted-symbols-gate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    env:
+      BASE_REF: ${{ github.base_ref }}
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v7
+        with:
+          python-version: "3.12"
+
+      - name: Fetch base branch
+        run: git fetch origin "$BASE_REF"
+
+      - name: Check for silently deleted symbols
+        env:
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: python scripts/check_deleted_symbols.py --base "origin/$BASE_REF"
--- a/scripts/check_deleted_symbols.py
+++ b/scripts/check_deleted_symbols.py
@@ -184,7 +184,7 @@ def check_deleted_symbols(
 
     violations: list[Violation] = []
     waived_in_signal: set[str] = set()
-    for symbol, kind in signal.items():
+    for symbol, kind in sorted(signal.items()):
         if symbol in waived_set:
             waived_in_signal.add(symbol)
             continue
@@ -222,7 +222,7 @@ def main(argv: list[str] | None = None) -> int:
     if violations:
         print(
             f"DELETED-SYMBOLS FAIL: this PR deletes {len(violations)} symbol(s) that "
-            f"landed on dev after your branch was cut:"
+            f"exist on the base branch but are missing from the merge result:"
         )
         for v in violations:
             print(f"  - {v.symbol} (added by {v.added_by})")
--- a/scripts/check_deleted_symbols.py
+++ b/scripts/check_deleted_symbols.py
@@ -1,0 +1,236 @@
+#!/usr/bin/env python3
+"""Deleted-symbols guard.
+
+Detects PRs that silently delete code added to the target branch after the
+merge base. This catches the "clean merge, no conflict" failure mode where a
+PR branch cut before hardening commits landed on dev would silently delete
+them when merged -- git reports "Automatic merge went well" with no conflict
+because the PR branch simply wins on files dev touched after the branch point.
+
+Algorithm:
+  1. Extract all Python def/class symbols at two points: the target branch
+     head and HEAD (the merge ref / PR head).
+  2. A symbol is "deleted by the PR" if it exists at the target head but not
+     at HEAD. The signal is that set.
+  3. Fail with an explicit list of the deleted symbols and the commits that
+     added them.
+  4. A "Removes-Intentionally: <symbol>, ..." trailer in the PR body waives
+     named symbols, making deliberate deletions a conscious, auditable act.
+
+Narrow by design: Python def/class names only (test functions are defs).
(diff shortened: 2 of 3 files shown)


jaylfc:
Lead verification, per the gate rule: I ran this script from the branch in a constructed scratch repo, all three directions, at the merge-result ref CI checks out.
RED: PR deletes a long-merged test file -> DELETED-SYMBOLS FAIL naming tests/test_lib.py:test_a and test_b, exit 1. This is the exact shape the previous predicate was mathematically blind to.
GREEN: unrelated one-file PR on a moving base -> clean, exit 0 (no stale-branch false positive).
WAIVER: same deletion with Removes-Intentionally lines in PR_BODY -> exit 0.
One cosmetic nit for a later pass, not blocking: the failure message says symbols 'landed on dev after your branch was cut', which is no longer what the predicate measures - they can predate the cut. Merging on green CI + bot adjudication.

jaylfc:

VERDICT: BLOCKING - Fundamental algorithmic flaw causes false positives; missing merge-base comparison defeats the tool's stated purpose.

- scripts/check_deleted_symbols.py:10-17: Algorithm compares only target-head vs HEAD (two-point), but the stated goal requires three-point comparison (target-head vs HEAD vs merge-base). Current logic flags ANY symbol missing from PR that exists on target, including symbols that existed at merge-base and were deliberately deleted by the PR. This defeats the "silently delete code added after branch cut" purpose.
- scripts/check_deleted_symbols.py:113-141: `_find_adding_commit` searches only `base_ref` history with `git log -S` on leaf name + paren. Misses symbols added via merge commits, matches false positives in comments/strings, and fallback search without paren is overly broad.
- scripts/check_deleted_symbols.py:53-57: `_run_git` lacks timeout; `subprocess.run(check=True)` can hang indefinitely on git failures or large repos.
- .github/workflows/deleted-symbols-gate.yml:23,27: Uses `actions/checkout@v7` and `actions/setup-python@v7` — these versions may not exist (current stable are v4/v5 as of 2026).
- tests/test_check_deleted_symbols.py:292-313: Test `test_no_signal_when_dev_unchanged_since_merge_base` expects a violation when PR deletes code existing at merge-base with no dev changes — this asserts the buggy two-point behavior as correct.
- tests/test_check_deleted_symbols.py:315-341: Test `test_probe3_pr_deletes_long_merged_test_file_fails` similarly expects failure for deleting a file that existed at merge-base ("long merged"), confirming the algorithm flags intentional deletions as violations.
- tests/test_check_deleted_symbols.py: Missing tests for: renamed/moved symbols, symbols added in merge commits, empty/syntax-error files, concurrent modifications on same symbol, and waiver parsing edge cases (trailer case sensitivity, whitespace).
VERDICT: BLOCKING - Fundamental algorithmic flaw causes false positives; missing merge-base comparison defeats the tool's stated purpose.

- scripts/check_deleted_symbols.py:10-17: Algorithm compares only target-head vs HEAD (two-point), but the stated goal requires three-point comparison (target-head vs HEAD vs merge-base). Current logic flags ANY symbol missing from PR that exists on target, including symbols that existed at merge-base and were deliberately deleted by the PR. This defeats the "silently delete code added after branch cut" purpose.
- scripts/check_deleted_symbols.py:113-141: `_find_adding_commit` searches only `base_ref` history with `git log -S` on leaf name + paren. Misses symbols added via merge commits, matches false positives in comments/strings, and fallback search without paren is overly broad.
- scripts/check_deleted_symbols.py:53-57: `_run_git` lacks timeout; `subprocess.run(check=True)` can hang indefinitely on git failures or large repos.
- .github/workflows/deleted-symbols-gate.yml:23,27: Uses `actions/checkout@v7` and `actions/setup-python@v7` — these versions may not exist (current stable are v4/v5 as of 2026).
- tests/test_check_deleted_symbols.py:292-313: Test `test_no_signal_when_dev_unchanged_since_merge_base` expects a violation when PR deletes code existing at merge-base with no dev changes — this asserts the buggy two-point behavior as correct.
- tests/test_check_deleted_symbols.py:315-341: Test `test_probe3_pr_deletes_long_merged_test_file_fails` similarly expects failure for deleting a file that existed at merge-base ("long merged"), confirming the algorithm flags intentional deletions as violations.
- tests/test_check_deleted_symbols.py: Missing tests for: renamed/moved symbols, symbols added in merge commits, empty/syntax-error files, concurrent modifications on same symbol, and waiver parsing edge cases (trailer case sensitivity, whitespace).

jaylfc:
**nemotron-ultra-orB review** 

VERDICT: Strong implementation with good test coverage; one correctness concern and minor style nits.

- `scripts/check_deleted_symbols.py:100`: `_extract_symbols` uses `ast.iter_child_nodes` which only visits immediate children. The recursive `visit` call handles nested defs/classes in function bodies, but **misses symbols inside `if`/`try`/`with` blocks or other compound statements** since those don't create new scope nodes in the AST. Nested functions inside blocks *are* visited (they're in `body`), but this is fragile. Consider using `ast.walk` or a proper `NodeVisitor` subclass.

- `scripts/check_deleted_symbols.py:132`: `_get_symbols_at_ref` uses `git archive` + tarfile which loads entire repo into memory. For large repos this could OOM. Consider `git ls-tree -r --name-only <ref> -- '*.py'` + `git show <ref>:<path>` streaming per file.

- `scripts/check_deleted_symbols.py:154`: `_find_adding_commit` searches for `kind leaf(`. This **fails for async functions** (searched as `def leaf(` but actual is `async def leaf(`) and **class definitions without parentheses** (fallback handles this). The async case will return "unknown" incorrectly.

- `tests/test_check_deleted_symbols.py`: Missing test for async function deletion (covers the bug above). No test for symbols inside nested scopes (if/try/with blocks).

- `.github/workflows/deleted-symbols-gate.yml:28`: `actions/checkout@v7` doesn't exist yet (latest is v4). Will fail at runtime.

- Style: `scripts/check_deleted_symbols.py:54` `REPO_ROOT` uses `Path(__file__).resolve().parent.parent` - brittle if script moves. Consider `git rev-parse --show-toplevel`.

No blocking issues beyond the async function detection bug and the checkout version.

jaylfc:
**nemotron-super review** 

VERDICT: Pass
No blocking issues found.

jaylfc:
Qodo adjudication + folds pushed: (1) REAL - test_deletes_own_newly_added_code_passes committed its PR-branch work on main (_branch creates without checkout; every sibling test has the _checkout call, this one missed it) - fixed, the test now models its scenario and still passes for the right reason. (2) REAL-minor - failure text said symbols landed on dev after the branch cut, which is not what the predicate measures - reworded to base-branch-vs-merge-result. (3) Performance note on per-symbol git log -S - real but deletions are rare; noted, not folded. (4) REAL-trivial - violation order now sorted for stable CI output. 23/23 tests green locally. Kilo check is the rate-limited fake-red (no review content) - this push retriggers it; merging on a real kilo result + green CI.

PR #2228, opened 2026-08-02
fix-forward: retrofit-migrations gate handles AnnAssign + wiring

jaylfc:
Autonomous build of board card tsk-ybllhu.

The gate in scripts/check_retrofit_migrations.py only handled
ast.Assign but every real store uses annotated assignments
(MIGRATIONS: list = [...], SCHEMA: str = ...) so the guard was
blind to the entire tree. Handle ast.AnnAssign in all four scan
locations: module-level constant collection, class-body constant
collection, module-level store scan, class-level store scan.

Also:
- CREATE INDEX violations now print index-appropriate fix text
  (PRAGMA index_list + CREATE INDEX IF NOT EXISTS) instead of the
  ADD COLUMN advice
- Wire the script into .github/workflows/doc-gate.yml beside
  check_schema_migrations.py as the retrofit-migration-guard step
- Note the invoking workflow in the script docstring
- Add tests/test_check_retrofit_migrations.py covering AnnAssign
  violations (ADD COLUMN + CREATE INDEX), Assign violations, both
  exemption cases (bootstrap SCHEMA reference, CREATE TABLE IF NOT
  EXISTS), and unparseable-file graceful handling

Files:
 .github/workflows/doc-gate.yml          |   3 +
 scripts/check_retrofit_migrations.py    | 333 ++++++++++++++++++++++++++++++++
 tests/test_check_retrofit_migrations.py | 285 +++++++++++++++++++++++++++
 tests/test_installed_apps.py            |  77 +-------
 4 files changed, 622 insertions(+), 76 deletions(-)

### DIFF
--- a/.github/workflows/doc-gate.yml
+++ b/.github/workflows/doc-gate.yml
@@ -34,6 +34,9 @@ jobs:
       - name: Schema-migration guard (#1865)
         run: python scripts/check_schema_migrations.py
 
+      - name: Retrofit-migration guard (#2188)
+        run: python scripts/check_retrofit_migrations.py
+
       - name: Diff gate (Layer B)
         env:
           BASE_REF: ${{ github.base_ref }}
--- a/scripts/check_retrofit_migrations.py
+++ b/scripts/check_retrofit_migrations.py
@@ -30,6 +30,9 @@ It does NOT flag:
 
 Usage:
     python scripts/check_retrofit_migrations.py
+Invoked by the ``retrofit-migration-guard`` step in
+``.github/workflows/doc-gate.yml``.
+
 Prints ``retrofit-migration-guard: clean`` and exits 0 when no violations,
 or prints each violation and exits 1.
 
@@ -75,10 +78,16 @@ class Violation:
     detail: str
 
     def __str__(self) -> str:
-        fix = (
-            "move this migration into a guarded _post_init coroutine "
-            "(PRAGMA table_info check + ALTER TABLE only when absent)"
-        )
+        if self.kind == "CREATE INDEX":
+            fix = (
+                "move this index creation into a guarded _post_init coroutine "
+                "(PRAGMA index_list check + CREATE INDEX IF NOT EXISTS only when absent)"
+            )
+        else:
+            fix = (
+                "move this migration into a guarded _post_init coroutine "
+                "(PRAGMA table_info check + ALTER TABLE only when absent)"
+            )
         return (
             f"{self.path}: store '{self.store}', migration v{self.version}, "
             f"table '{self.table}' -- {self.kind} on a SCHEMA table\n"
@@ -169,7 +178,7 @@ def find_violations(path: Path) -> list[Violation]:
     except (OSError, SyntaxError):
         return []
 
-    # Collect all module-level string constant assignments (name -> value).
+    # Collect all module-level string/list constant assignments (name -> value).
     string_constants: dict[str, str] = {}
     list_constants: dict[str, ast.List] = {}
     for node in tree.body:
@@ -180,15 +189,31 @@ def find_violations(path: Path) -> list[Violation]:
                         string_constants[target.id] = node.value.value
                     elif isinstance(node.value, ast.List):
                         list_constants[target.id] = node.value
+        elif isinstance(node, ast.AnnAssign):
+            if isinstance(node.target, ast.Name) and node.value is not None:
+                if isinstance(node.value, ast.Constant) and isinstance(node.value.value, str):
+                    string_constants[node.target.id] = node.value.value
+                elif isinstance(node.value, ast.List):
+                    list_constants[node.target.id] = node.value
 
     # Also collect string constants defined inside class bodies (for
-    # SCHEMA = "..." inside a class).
+    # SCHEMA = "..." inside a class), including annotated assignments
+    # (e.g. SCHEMA: str = "...").
     for node in ast.walk(tree):
         if isinstance(node, ast.Assign):
             for target in node.targets:
                 if isinstance(target, ast.Name) and isinstance(node.value, ast.Constant) and isinstance(node.value.value, str):
                     if target.id not in string_constants:
                         string_constants[target.id] = node.value.value
+        elif isinstance(node, ast.AnnAssign):
+            if (
+                isinstance(node.target, ast.Name)
+                and node.value is not None
+                and isinstance(node.value, ast.Constant)
+                and isinstance(node.value.value, str)
+            ):
+                if node.target.id not in string_constants:
+                    string_constants[node.target.id] = node.value.value
 
     violations: list[Violation] = []
 
@@ -203,6 +228,12 @@ def find_violations(path: Path) -> list[Violation]:
                         module_schema = _resolve_string(node.value, string_constants)
                     elif target.id == "MIGRATIONS":
                         module_migrations = _resolve_list_node(node.value, list_constants)
+        elif isinstance(node, ast.AnnAssign):
+            if isinstance(node.target, ast.Name):
+                if node.target.id == "SCHEMA":
+                    module_schema = _resolve_string(node.value, string_constants)
+                elif node.target.id == "MIGRATIONS":
+                    module_migrations = _resolve_list_node(node.value, list_constants)
 
     if module_schema and module_migrations:
         schema_tables = _extract_schema_tables(module_schema)
@@ -227,6 +258,12 @@ def find_violations(path: Path) -> list[Violation]:
                             class_schema = _resolve_string(item.value, string_constants)
                         elif target.id == "MIGRATIONS":
                             class_migrations = _resolve_list_node(item.value, list_constants)
+            elif isinstance(item, ast.AnnAssign):
+                if isinstance(item.target, ast.Name):
+                    if item.target.id == "SCHEMA":
+                        class_schema = _resolve_string(item.value, string_constants)
+                    elif item.target.id == "MIGRATIONS":
+                        class_migrations = _resolve_list_node(item.value, list_constants)
 
         if class_schema and class_migrations:
             schema_tables = _extract_schema_tables(class_schema)
--- a/scripts/check_retrofit_migrations.py
+++ b/scripts/check_retrofit_migrations.py
@@ -320,7 +320,9 @@ def find_all_violations(root: Path = STORES_ROOT) -> list[Violation]:
 
 
 def main(argv: list[str] | None = None) -> int:
-    violations = find_all_violations()
+    args = argv if argv is not None else sys.argv[1:]
+    root = Path(args[0]) if args else STORES_ROOT
+    violations = find_all_violations(root)
     if not violations:
         print("retrofit-migration-guard: clean")
         return 0
--- a/scripts/check_retrofit_migrations.py
+++ b/scripts/check_retrofit_migrations.py
@@ -1,0 +1,296 @@
+#!/usr/bin/env python3
+"""Static guard against retrofit migrations placed in MIGRATIONS (#tsk-gqsuah).
+
+``BaseStore.init()`` runs a store's ``SCHEMA`` string (CREATE TABLE +
+CREATE INDEX) BEFORE the migration runner.  The runner uses
+baseline-at-latest semantics: pre-existing databases are stamped at the
+latest version WITHOUT executing any SQL, so any migration that ALTERs
+or adds an index to a table that ALSO appears in SCHEMA is silently
+skipped on exactly the databases that need it.  The tracking table then
+records the migration as applied -- a false success.
+
+The correct pattern (mandated in the BaseStore docstring and
+``db_migrations.py`` footgun #2) is a guarded ``_post_init`` coroutine:
+``PRAGMA table_info`` check + ``ALTER TABLE`` only when the column is
+absent.  Such a migration runs every init, stays permanently in the
+codebase, and self-checks whether it already ran.
+
+This script statically inspects every Python file under ``tinyagentos/``
+that defines a store and flags the dangerous pattern: a MIGRATIONS entry
+that ALTERs or adds a CREATE INDEX to a table that ALSO appears in that
+store's SCHEMA -- i.e. a retrofit to a pre-existing table that
+baseline-at-latest will skip.
+
+It does NOT flag:
+  - CREATE TABLE IF NOT EXISTS in MIGRATIONS (genuinely new tables are
+    safe: idempotent on existing DBs).
+  - Migrations that reference the SCHEMA constant itself (the bootstrap
+    pattern ``MIGRATIONS = [(1, SCHEMA)]``), since that is the initial
+    schema creation, not a retrofit.
+
+Usage:
+    python scripts/check_retrofit_migrations.py
+Prints ``retrofit-migration-guard: clean`` and exits 0 when no violations,
+or prints each violation and exits 1.
+
+Dependency-light: stdlib only (ast + re + pathlib).
+"""
+from __future__ import annotations
+
+import ast
+import re
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+STORES_ROOT = REPO_ROOT / "tinyagentos"
+
+# ALTER TABLE <table> ADD [COLUMN] <col> ...
+_ADD_COLUMN_RE = re.compile(
+    r"ALTER\s+TABLE\s+(\w+)\s+ADD\s+(?:COLUMN\s+)?(\w+)",
+    re.IGNORECASE,
+)
+
+# CREATE [UNIQUE] INDEX [IF NOT EXISTS] <name> ON <table> (col, col, ...)
+_CREATE_INDEX_RE = re.compile(
+    r"CREATE\s+(?:UNIQUE\s+)?INDEX\s+(?:IF\s+NOT\s+EXISTS\s+)?\w+\s+ON\s+(\w+)",
+    re.IGNORECASE,
+)
+
+# CREATE TABLE [IF NOT EXISTS] <table> ( ... )
+_CREATE_TABLE_RE = re.compile(
+    r"CREATE\s+TABLE\s+(?:IF\s+NOT\s+EXISTS\s+)?(\w+)\s*\(",
+    re.IGNORECASE,
+)
+
+
+@dataclass
+class Violation:
+    path: Path
+    store: str
+    version: int
+    table: str
+    kind: str  # "ALTER TABLE ADD COLUMN" or "CREATE INDEX"
+    detail: str
+
+    def __str__(self) -> str:
+        fix = (
+            "move this migration into a guarded _post_init coroutine "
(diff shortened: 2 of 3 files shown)


jaylfc:
**nemotron-ultra-orB review** 

VERDICT: The retrofit-migration-guard implementation is well-structured with good test coverage, but has several static analysis blind spots that could miss violations.

- **scripts/check_retrofit_migrations.py:130-133**: `_resolve_string` only handles `ast.Constant` and `ast.Name` — misses concatenated strings, f-strings, method calls, or any expression that evaluates to SQL. A migration like `MIGRATIONS = [(1, "ALTER TABLE " + "widgets" + " ADD COLUMN x TEXT")]` would not be checked.

- **scripts/check_retrofit_migrations.py:142-147**: `_resolve_list_node` has the same limitation — only resolves direct list literals or constant references. Dynamically built MIGRATIONS lists bypass the check entirely.

- **scripts/check_retrofit_migrations.py:83-95**: Regex patterns use `\w+` for identifiers — won't match quoted SQLite identifiers (`"my-table"`, `table-name`, `column_name` with special chars). Could miss `ALTER TABLE "user-data" ADD COLUMN ...` or `CREATE INDEX idx ON "order" (...)`.

- **scripts/check_retrofit_migrations.py:293-300**: In `_check_migrations_list`, unresolvable SQL nodes (callables, complex expressions) are silently skipped with `continue`. No warning or logging — violations in such migrations go undetected.

- **scripts/check_retrofit_migrations.py:250-260**: `ast.walk(tree)` scans entire class bodies including nested classes/functions. Could falsely flag SCHEMA/MIGRATIONS in inner scopes that aren't actual store definitions (though unlikely in practice).

- **tests/test_check_retrofit_migrations.py:327,330**: Test relies on hardcoded relative paths (`parent.parent`, `scripts/`) and mutates `sys.path`. Fragile if file structure changes; consider using a pytest fixture for the scripts module import.

- **tests/test_check_retrofit_migrations.py:248-257**: `test_unparseable_file_with_real_violation_returns_empty` contains invalid syntax (`[[[`) in the test string itself — this is intentional but confusing; add a comment clarifying it's deliberate malformed SQL.

No blocking issues found.

jaylfc:
Lead verification per the gate rule - ran the script from this branch in a constructed repo layout, all directions:
RED module-scope AnnAssign retrofit (the shape every real store uses, which the previous version was blind to): fires, exit 1.
RED module-scope plain Assign retrofit: fires, exit 1.
RED class-scope AnnAssign (the invite_store idiom): fires, exit 1.
GREEN new-table CREATE TABLE IF NOT EXISTS migration: clean, exit 0.
GREEN bootstrap MIGRATIONS = [(1, SCHEMA)] with index in schema: clean, exit 0.
Wired: doc-gate.yml invokes it. The doc-gate check on THIS PR doubles as the no-false-positive proof against the real tree, since every real store is AnnAssign-declared.
Two non-blocking nits for a later pass: main(argv) accepts argv and ignores it (inert param - either honor a root override or drop the param), and the CREATE INDEX violation still prints the ADD COLUMN fix text (card item 4, looks unaddressed - confirm or fold before I merge).
Merging on green CI + bot adjudication, minus the item-4 answer.

jaylfc:
Correction to my last comment: card item 4 IS addressed - probed a CREATE INDEX violation and it prints the index-specific fix text (PRAGMA index_list + CREATE INDEX IF NOT EXISTS), not the ADD COLUMN advice. Only remaining nit is the inert main(argv) param, non-blocking. This merges on green CI.

jaylfc:

I'll review the diff directly since the files are provided in the diff content.

---

**VERDICT: Several issues found — logic gaps in SQL parsing, incomplete test coverage for edge cases, and a potential false-negative pattern.**

- **scripts/check_retrofit_migrations.py:58-62** — `_CREATE_INDEX_RE` misses `CREATE UNIQUE INDEX` without `IF NOT EXISTS` because the regex requires `\w+\s+ON` (index name then ON), but `CREATE UNIQUE INDEX idx ON table(col)` has `UNIQUE` between `INDEX` and the name. The pattern `INDEX\s+(?:IF\s+NOT\s+EXISTS\s+)?\w+\s+ON` fails to match `CREATE UNIQUE INDEX idx ON t(c)` because `UNIQUE` is not optional in the regex — it's part of `(?:UNIQUE\s+)?` before `INDEX`, but the group captures `INDEX` then expects optional `IF NOT EXISTS` then name. Actually wait: `CREATE\s+(?:UNIQUE\s+)?INDEX\s+(?:IF\s+NOT\s+EXISTS\s+)?\w+\s+ON` — this matches `CREATE INDEX idx ON`, `CREATE UNIQUE INDEX idx ON`, `CREATE INDEX IF NOT EXISTS idx ON`, `CREATE UNIQUE INDEX IF NOT EXISTS idx ON`. But it requires a word `\w+` (the index name) before `ON`. What about `CREATE INDEX ON table(col)` (anonymous index)? SQLite doesn't allow that. But it misses `CREATE INDEX idx ON table (col)` with parentheses spacing — the `\w+\s+ON` requires whitespace before ON, which is fine. **Main issue**: the regex doesn't handle `CREATE INDEX IF NOT EXISTS` with `UNIQUE` combined — wait it does. Let me re-read: `(?:UNIQUE\s+)?` is before `INDEX`, `(?:IF\s+NOT\s+EXISTS\s+)?` is after `INDEX`. So `CREATE UNIQUE INDEX IF NOT EXISTS idx ON t(c)` matches. **Real gap**: the regex captures the table name in group 1 via `ON\s+(\w+)`. But `CREATE INDEX idx ON "table-name" (col)` with quoted table names (SQLite allows quoted identifiers) won't match. Also `CREATE INDEX idx ON schema.table (col)` won't match. Low risk but a gap.

- **scripts/check_retrofit_migrations.py:53-56** — `_ADD_COLUMN_RE` misses `ALTER TABLE table ADD COLUMN col TYPE` with column constraints (e.g., `NOT NULL`, `DEFAULT`). The regex captures `ADD\s+(?:COLUMN\s+)?(\w+)` — stops at the column name. That's fine for detection purposes (we only need table + column name). But it misses `ALTER TABLE "table" ADD COLUMN col` (quoted table name). Same quoted identifier gap.

- **scripts/check_retrofit_migrations.py:50-52** — `_CREATE_TABLE_RE` only matches `CREATE TABLE [IF NOT EXISTS] name (`. Misses `CREATE TABLE "name" (` (quoted), `CREATE TABLE schema.name (`, and `CREATE TEMPORARY TABLE`. Since this is used to extract schema tables, missing a table in SCHEMA means false negatives (retrofit on that table won't be flagged). **This is a correctness bug.**

- **scripts/check_retrofit_migrations.py:185-198** — Constant collection walks `ast.walk(tree)` *after* already collecting module-level constants. This double-collects module-level constants (harmless but wasteful). More importantly, it collects constants from *nested functions/classes* into the global `string_constants`/`list_constants` dicts. If a nested function defines `SCHEMA = "..."`, it pollutes the module namespace. Could cause false positives/negatives if a nested constant shadows a module one. Should scope constant collection to module/class level only.

- **scripts/check_retrofit_migrations.py:220-245** — Module-level store detection only checks `ast.Assign` and `ast.AnnAssign` directly in `tree.body`. Misses `SCHEMA`/`MIGRATIONS` defined inside `if __name__ == "__main__":` blocks or other nested blocks at module level. Not a real-world pattern but a gap.

- **scripts/check_retrofit_migrations.py:247-275** — Class-level store detection uses `ast.walk(node)` which traverses *into nested classes/functions*. If a nested class defines `SCHEMA`, it'll be picked up as the outer class's schema. Should iterate `node.body` directly.

- **scripts/check_retrofit_migrations.py:283** — `_check_migrations_list` skips non-tuple entries and tuples with `<2` elements silently. If `MIGRATIONS = [1, 2, 3]` (malformed), it returns clean. Should at least warn or error. But since this is a guard script, silent skip is a false negative risk.

- **scripts/check_retrofit_migrations.py:296** — `sql = _resolve_string(sql_node, string_constants)` returns `None` for callables (functions), call expressions, joined strings (f-strings), binops. The script silently skips these. **Major false negative gap**: if a migration SQL is built via `f"ALTER TABLE {table} ADD COLUMN {col}"` or a helper function, it's not analyzed. The docstring says "Dependency-light: stdlib only" but this limitation should be documented or handled.

- **scripts/check_retrofit_migrations.py:308** — `find_all_violations` uses `sorted(root.rglob("*.py"))`. If `tinyagentos/` has symlinks pointing outside, `rglob` follows them. Could read arbitrary files. Low risk in CI but a security consideration.

- **tests/test_check_retrofit_migrations.py:16** — `sys.path.insert(0, ...)` modifies global state. Tests run in isolation (pytest) so mostly fine, but could leak if tests run in same process. Better to use `importlib` or a fixture.

- **tests/test_check_retrofit_migrations.py** — No tests for:
  - Quoted identifiers in SQL (`CREATE TABLE "widgets"`, `ALTER TABLE "widgets" ADD COLUMN`)
  - Schema-qualified tables (`CREATE TABLE main.widgets`)
  - F-string / callable migrations (expected false negatives)
  - Nested class constant shadowing
  - Malformed MIGRATIONS entries (non-tuples, wrong arity)
  - `CREATE TEMPORARY TABLE` in SCHEMA
  - Multiple tables in one migration SQL (e.g., `ALTER TABLE a ADD COLUMN x; ALTER TABLE b ADD COLUMN y;`)

- **tests/test_check_retrofit_migrations.py:247-253** — `test_unparseable_file_returns_empty` and `test_unparseable_file_with_real_violation_returns_empty` test SyntaxError handling. But they don't test `OSError` (permission denied, etc.) which the `except (OSError, SyntaxError)` also catches. Minor.

- **.github/workflows/doc-gate.yml:37-39** — The new step runs `python scripts/check_retrofit_migrations.py` without `continue-on-error: false` (default) so it blocks on failure. Good. But no `timeout-minutes` — if the script hangs on a huge codebase, the workflow hangs. Add a timeout.

- **Style**: `scripts/check_retrofit_migrations.py` has inconsistent docstring format (module docstring is a mix of narrative and usage). The `_normalize` function collapses whitespace but doesn't remove comments — SQL comments in SCHEMA vs migration could cause bootstrap detection to fail. E.g., `SCHEMA = "CREATE TABLE t (id INT); -- comment"` vs migration `(1, SCHEMA)` — `_normalize` keeps `-- comment` so they won't match. Bootstrap exemption fails → false positive. **Correctness bug.**

- **Style**: `_extract_schema_tables` returns a set of table names. If SCHEMA has `CREATE TABLE t (id INT); CREATE TABLE t (id INT);` (duplicate), set dedupes. Fine. But if SCHEMA uses `CREATE TABLE IF NOT EXISTS t`, the regex captures `t`. Good.

- **Style**: `Violation.__str__` includes fix advice that mentions `_post_init` coroutine — assumes the codebase pattern. If a store doesn't use `BaseStore`, the fix advice is misleading. But this is taOS-specific so acceptable.

- **Missing test**: The bootstrap exemption test (`test_exemption_bootstrap_schema_reference`) only tests `MIGRATIONS = [(1, SCHEMA)]`. What about `MIGRATIONS = [(1, SCHEMA_CONST)]` where `SCHEMA_CONST` is a different constant pointing to the same string? The normalization compares strings, so it works. But what about `MIGRATIONS = [(1, SCHEMA + "; CREATE INDEX ...")]`? Not the bootstrap pattern, correctly flagged.
I'll review the diff directly since the files are provided in the diff content.

---

**VERDICT: Several issues found — logic gaps in SQL parsing, incomplete test coverage for edge cases, and a potential false-negative pattern.**

- **scripts/check_retrofit_migrations.py:58-62** — `_CREATE_INDEX_RE` misses `CREATE UNIQUE INDEX` without `IF NOT EXISTS` because the regex requires `\w+\s+ON` (index name then ON), but `CREATE UNIQUE INDEX idx ON table(col)` has `UNIQUE` between `INDEX` and the name. The pattern `INDEX\s+(?:IF\s+NOT\s+EXISTS\s+)?\w+\s+ON` fails to match `CREATE UNIQUE INDEX idx ON t(c)` because `UNIQUE` is not optional in the regex — it's part of `(?:UNIQUE\s+)?` before `INDEX`, but the group captures `INDEX` then expects optional `IF NOT EXISTS` then name. Actually wait: `CREATE\s+(?:UNIQUE\s+)?INDEX\s+(?:IF\s+NOT\s+EXISTS\s+)?\w+\s+ON` — this matches `CREATE INDEX idx ON`, `CREATE UNIQUE INDEX idx ON`, `CREATE INDEX IF NOT EXISTS idx ON`, `CREATE UNIQUE INDEX IF NOT EXISTS idx ON`. But it requires a word `\w+` (the index name) before `ON`. What about `CREATE INDEX ON table(col)` (anonymous index)? SQLite doesn't allow that. But it misses `CREATE INDEX idx ON table (col)` with parentheses spacing — the `\w+\s+ON` requires whitespace before ON, which is fine. **Main issue**: the regex doesn't handle `CREATE INDEX IF NOT EXISTS` with `UNIQUE` combined — wait it does. Let me re-read: `(?:UNIQUE\s+)?` is before `INDEX`, `(?:IF\s+NOT\s+EXISTS\s+)?` is after `INDEX`. So `CREATE UNIQUE INDEX IF NOT EXISTS idx ON t(c)` matches. **Real gap**: the regex captures the table name in group 1 via `ON\s+(\w+)`. But `CREATE INDEX idx ON "table-name" (col)` with quoted table names (SQLite allows quoted identifiers) won't match. Also `CREATE INDEX idx ON schema.table (col)` won't match. Low risk but a gap.

- **scripts/check_retrofit_migrations.py:53-56** — `_ADD_COLUMN_RE` misses `ALTER TABLE table ADD COLUMN col TYPE` with column constraints (e.g., `NOT NULL`, `DEFAULT`). The regex captures `ADD\s+(?:COLUMN\s+)?(\w+)` — stops at the column name. That's fine for detection purposes (we only need table + column name). But it misses `ALTER TABLE "table" ADD COLUMN col` (quoted table name). Same quoted identifier gap.

- **scripts/check_retrofit_migrations.py:50-52** — `_CREATE_TABLE_RE` only matches `CREATE TABLE [IF NOT EXISTS] name (`. Misses `CREATE TABLE "name" (` (quoted), `CREATE TABLE schema.name (`, and `CREATE TEMPORARY TABLE`. Since this is used to extract schema tables, missing a table in SCHEMA means false negatives (retrofit on that table won't be flagged). **This is a correctness bug.**

- **scripts/check_retrofit_migrations.py:185-198** — Constant collection walks `ast.walk(tree)` *after* already collecting module-level constants. This double-collects module-level constants (harmless but wasteful). More importantly, it collects constants from *nested functions/classes* into the global `string_constants`/`list_constants` dicts. If a nested function defines `SCHEMA = "..."`, it pollutes the module namespace. Could cause false positives/negatives if a nested constant shadows a module one. Should scope constant collection to module/class level only.

- **scripts/check_retrofit_migrations.py:220-245** — Module-level store detection only checks `ast.Assign` and `ast.AnnAssign` directly in `tree.body`. Misses `SCHEMA`/`MIGRATIONS` defined inside `if __name__ == "__main__":` blocks or other nested blocks at module level. Not a real-world pattern but a gap.

- **scripts/check_retrofit_migrations.py:247-275** — Class-level store detection uses `ast.walk(node)` which traverses *into nested classes/functions*. If a nested class defines `SCHEMA`, it'll be picked up as the outer class's schema. Should iterate `node.body` directly.

- **scripts/check_retrofit_migrations.py:283** — `_check_migrations_list` skips non-tuple entries and tuples with `<2` elements silently. If `MIGRATIONS = [1, 2, 3]` (malformed), it returns clean. Should at least warn or error. But since this is a guard script, silent skip is a false negative risk.

- **scripts/check_retrofit_migrations.py:296** — `sql = _resolve_string(sql_node, string_constants)` returns `None` for callables (functions), call expressions, joined strings (f-strings), binops. The script silently skips these. **Major false negative gap**: if a migration SQL is built via `f"ALTER TABLE {table} ADD COLUMN {col}"` or a helper function, it's not analyzed. The docstring says "Dependency-light: stdlib only" but this limitation should be documented or handled.

- **scripts/check_retrofit_migrations.py:308** — `find_all_violations` uses `sorted(root.rglob("*.py"))`. If `tinyagentos/` has symlinks pointing outside, `rglob` follows them. Could read arbitrary files. Low risk in CI but a security consideration.

- **tests/test_check_retrofit_migrations.py:16** — `sys.path.insert(0, ...)` modifies global state. Tests run in isolation (pytest) so mostly fine, but could leak if tests run in same process. Better to use `importlib` or a fixture.

- **tests/test_check_retrofit_migrations.py** — No tests for:
  - Quoted identifiers in SQL (`CREATE TABLE "widgets"`, `ALTER TABLE "widgets" ADD COLUMN`)
  - Schema-qualified tables (`CREATE TABLE main.widgets`)
  - F-string / callable migrations (expected false negatives)
  - Nested class constant shadowing
  - Malformed MIGRATIONS entries (non-tuples, wrong arity)
  - `CREATE TEMPORARY TABLE` in SCHEMA
  - Multiple tables in one migration SQL (e.g., `ALTER TABLE a ADD COLUMN x; ALTER TABLE b ADD COLUMN y;`)

- **tests/test_check_retrofit_migrations.py:247-253** — `test_unparseable_file_returns_empty` and `test_unparseable_file_with_real_violation_returns_empty` test SyntaxError handling. But they don't test `OSError` (permission denied, etc.) which the `except (OSError, SyntaxError)` also catches. Minor.

- **.github/workflows/doc-gate.yml:37-39** — The new step runs `python scripts/check_retrofit_migrations.py` without `continue-on-error: false` (default) so it blocks on failure. Good. But no `timeout-minutes` — if the script hangs on a huge codebase, the workflow hangs. Add a timeout.

- **Style**: `scripts/check_retrofit_migrations.py` has inconsistent docstring format (module docstring is a mix of narrative and usage). The `_normalize` function collapses whitespace but doesn't remove comments — SQL comments in SCHEMA vs migration could cause bootstrap detection to fail. E.g., `SCHEMA = "CREATE TABLE t (id INT); -- comment"` vs migration `(1, SCHEMA)` — `_normalize` keeps `-- comment` so they won't match. Bootstrap exemption fails → false positive. **Correctness bug.**

- **Style**: `_extract_schema_tables` returns a set of table names. If SCHEMA has `CREATE TABLE t (id INT); CREATE TABLE t (id INT);` (duplicate), set dedupes. Fine. But if SCHEMA uses `CREATE TABLE IF NOT EXISTS t`, the regex captures `t`. Good.

- **Style**: `Violation.__str__` includes fix advice that mentions `_post_init` coroutine — assumes the codebase pattern. If a store doesn't use `BaseStore`, the fix advice is misleading. But this is taOS-specific so acceptable.

- **Missing test**: The bootstrap exemption test (`test_exemption_bootstrap_schema_reference`) only tests `MIGRATIONS = [(1, SCHEMA)]`. What about `MIGRATIONS = [(1, SCHEMA_CONST)]` where `SCHEMA_CONST` is a different constant pointing to the same string? The normalization compares strings, so it works. But what about `MIGRATIONS = [(1, SCHEMA + "; CREATE INDEX ...")]`? Not the bootstrap pattern, correctly flagged.

jaylfc:
**nemotron-super review** 

VERDICT: Pass
No blocking issues found.

PR #2187, opened 2026-07-28
tsk-oju3st  [OPEN]  Consent approve: 'Assign later' / defer project bi

jaylfc:
Autonomous build of board card tsk-oju3st.

> **REVIEW WARNING (automated):** this card's text asks for tests, but the diff changes no test file. Either the acceptance criteria are unmet or the card needs correcting. Do not merge without resolving this.

Files:
 tinyagentos/routes/agent_auth_requests.py | 55 ++++++++++++++++++++++---------
 1 file changed, 39 insertions(+), 16 deletions(-)

----
## Summary by Gitar

- **Consent Approval Updates:**
    - Added `defer_binding` flag to `ApproveBody` and `approve_request_record` to allow minting unbound tokens.
    - Skipped project scope validation and membership creation when binding is deferred.

<sub>This will update automatically on new commits.</sub>

## Summary by CodeRabbit

- **New Features**
  - Added an optional **defer binding** setting during external-agent consent approval.
  - Administrators can now approve project-scoped access without immediately linking it to a specific project.
  - Deferred approvals avoid creating project membership, canvas settings, and channel associations until a project is selected.
  - Existing approval behavior remains unchanged by default.

PR #2235, opened 2026-08-02
tsk-5bup3r  [OPEN]  Registry tokens never expire: taOS mints JWTs with

jaylfc:
Autonomous build of board card tsk-5bup3r.

Files:
 tests/test_agent_registry.py         |  38 ++++++++++++
 tests/test_agent_registry_store.py   | 114 ++++++++++++++++++++++++++++++++++-
 tinyagentos/agent_registry_store.py  |  92 ++++++++++++++++++++++------
 tinyagentos/agent_token_auth.py      |  11 +++-
 tinyagentos/config.py                |   4 ++
 tinyagentos/routes/agent_registry.py |  31 +++++++++-
 6 files changed, 269 insertions(+), 21 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Registry tokens now expire after a configurable lifetime, defaulting to 24 hours.
  * Added token renewal through a dedicated API endpoint.
  * Renewed tokens preserve agent identity and metadata.
  * Legacy tokens without expiration claims can be supported during a configurable migration period.

* **Bug Fixes**
  * Expired, malformed, invalidly signed, and unauthorized tokens are now rejected appropriately.
  * Added validation for registry slug lookups and token claims.

### DIFF
--- a/tests/test_agent_registry.py
+++ b/tests/test_agent_registry.py
@@ -273,6 +273,8 @@ class TestTokenRoundTrip:
         assert payload["sub"] == "agent-20260609-120000"
         assert payload["iss"] == "taos-registry"
         assert "iat" in payload
+        assert "exp" in payload
+        assert payload["exp"] > payload["iat"]
         assert payload["user_id"] == "user-1"
         assert payload["framework"] == "openclaw"
 
@@ -577,6 +579,42 @@ class TestAgentRegistryRoutes:
     async def test_revoked_feed_admin_can_read(self, registry_client):
         resp = await registry_client.get("/api/agents/registry/revoked")
         assert resp.status_code == 200
+
+    async def test_renew_token_returns_new_token(self, registry_client):
+        """POST /api/agents/registry/token/renew issues a fresh token."""
+        reg_resp = await registry_client.post(
+            "/api/agents/registry/register",
+            json={"framework": "openclaw", "display_name": "Renew Me"},
+        )
+        assert reg_resp.status_code == 200
+        old_token = reg_resp.json()["token"]
+
+        renew_resp = await registry_client.post(
+            "/api/agents/registry/token/renew",
+            headers={"Authorization": f"Bearer {old_token}"},
+        )
+        assert renew_resp.status_code == 200
+        new_token = renew_resp.json()["token"]
+        assert new_token != old_token
+
+        pubkey_resp = await registry_client.get("/api/agents/registry/pubkey")
+        pub_pem = pubkey_resp.json()["public_key"].encode()
+        old_payload = verify_registry_token(old_token, pub_pem)
+        new_payload = verify_registry_token(new_token, pub_pem)
+        assert new_payload["sub"] == old_payload["sub"]
+        assert new_payload["user_id"] == old_payload["user_id"]
+        assert new_payload["exp"] > new_payload["iat"]
+
+    async def test_renew_token_requires_bearer(self, registry_client):
+        resp = await registry_client.post("/api/agents/registry/token/renew")
+        assert resp.status_code == 401
+
+    async def test_renew_token_rejects_invalid_token(self, registry_client):
+        resp = await registry_client.post(
+            "/api/agents/registry/token/renew",
+            headers={"Authorization": "Bearer not-a-valid-token"},
+        )
+        assert resp.status_code == 401
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_agent_registry_store.py
+++ b/tests/test_agent_registry_store.py
@@ -1,4 +1,5 @@
 import json
+import time
 from datetime import datetime, timezone
 from pathlib import Path
 
@@ -17,6 +18,7 @@ from tinyagentos.agent_registry_store import (
     load_or_create_signing_keypair,
     mint_canonical_id,
     mint_registry_token,
+    renew_registry_token,
     verify_registry_token,
     VALID_STATUSES,
 )
@@ -255,6 +257,14 @@ class TestTokenMinting:
         payload = verify_registry_token(token, pub)
         assert "iat" in payload
         assert isinstance(payload["iat"], int)
+
+    def test_token_has_exp(self, signing_keypair):
+        priv, pub = signing_keypair
+        token = mint_registry_token("agent-008", priv)
+        payload = verify_registry_token(token, pub)
+        assert "exp" in payload
+        assert isinstance(payload["exp"], int)
+        assert payload["exp"] > payload["iat"]
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


jaylfc:

VERDICT: The changes implement token expiration (exp claim), renewal endpoint, and migration window — mostly solid but with several correctness and security concerns.

- **Security: Token renewal accepts expired tokens without checking agent status first** (`tinyagentos/routes/agent_registry.py:765-774`). The `renew_registry_token` call verifies signature only (via `_verify_signature_only`), which explicitly skips exp check. The agent status check happens *after* renewal. An attacker with a stolen expired token could renew it even if the agent was deactivated between token expiry and renewal call. Fix: check agent status *before* calling `renew_registry_token`.

- **Security: Migration window uses server time without skew tolerance** (`tinyagentos/agent_registry_store.py:378-385`). `allow_no_exp_until` compares against `time.time()` directly. If server clock drifts or is set incorrectly, tokens without exp could be rejected prematurely or accepted too long. Consider adding a small skew buffer (e.g., ±60s) or document that NTP is required.

- **Correctness: `renew_registry_token` doesn't validate `sub` exists in payload** (`tinyagentos/agent_registry_store.py:400-418`). It assumes `payload["sub"]` is present. If a malformed but validly-signed token lacks `sub`, this will KeyError. Add defensive check or document that only registry-minted tokens are expected.

- **Correctness: `renew_registry_token` preserves `project_id` as `None` when absent** (`tinyagentos/agent_registry_store.py:413`). `payload.get("project_id")` returns `None` if missing, which `mint_registry_token` then includes as `"project_id": null` in the new token. Original tokens omit the key entirely. This changes the token structure. Fix: only pass `project_id` if present in original.

- **Test gap: No test for `renew_registry_token` with missing optional claims** (`tests/test_agent_registry_store.py`). The renewal tests only cover tokens with all fields. Should test renewal of minimal token (only required claims).

- **Test gap: No test for clock skew during migration window** (`tests/test_agent_registry_store.py:330-355`). The migration tests use exact cutoff boundaries. Real deployments need skew tolerance.

- **Style: `_verify_signature_only` duplicates JWT parsing logic** (`tinyagentos/agent_registry_store.py:283-312`). It reimplements header/payload splitting and base64url decode that `verify_registry_token` also does. Could extract shared parsing to avoid drift.

- **Config: `registry_token_migration_cutoff_ts` loaded as float but config may provide int** (`tinyagentos/config.py:203`). `float(data["registry_token_migration_cutoff_ts"])` works for both but explicit `float()` on int is fine. No issue.

- **Missing test: Renewal endpoint with deactivated agent** (`tests/test_agent_registry.py`). Should verify 403 when agent status != active.
VERDICT: The changes implement token expiration (exp claim), renewal endpoint, and migration window — mostly solid but with several correctness and security concerns.

- **Security: Token renewal accepts expired tokens without checking agent status first** (`tinyagentos/routes/agent_registry.py:765-774`). The `renew_registry_token` call verifies signature only (via `_verify_signature_only`), which explicitly skips exp check. The agent status check happens *after* renewal. An attacker with a stolen expired token could renew it even if the agent was deactivated between token expiry and renewal call. Fix: check agent status *before* calling `renew_registry_token`.

- **Security: Migration window uses server time without skew tolerance** (`tinyagentos/agent_registry_store.py:378-385`). `allow_no_exp_until` compares against `time.time()` directly. If server clock drifts or is set incorrectly, tokens without exp could be rejected prematurely or accepted too long. Consider adding a small skew buffer (e.g., ±60s) or document that NTP is required.

- **Correctness: `renew_registry_token` doesn't validate `sub` exists in payload** (`tinyagentos/agent_registry_store.py:400-418`). It assumes `payload["sub"]` is present. If a malformed but validly-signed token lacks `sub`, this will KeyError. Add defensive check or document that only registry-minted tokens are expected.

- **Correctness: `renew_registry_token` preserves `project_id` as `None` when absent** (`tinyagentos/agent_registry_store.py:413`). `payload.get("project_id")` returns `None` if missing, which `mint_registry_token` then includes as `"project_id": null` in the new token. Original tokens omit the key entirely. This changes the token structure. Fix: only pass `project_id` if present in original.

- **Test gap: No test for `renew_registry_token` with missing optional claims** (`tests/test_agent_registry_store.py`). The renewal tests only cover tokens with all fields. Should test renewal of minimal token (only required claims).

- **Test gap: No test for clock skew during migration window** (`tests/test_agent_registry_store.py:330-355`). The migration tests use exact cutoff boundaries. Real deployments need skew tolerance.

- **Style: `_verify_signature_only` duplicates JWT parsing logic** (`tinyagentos/agent_registry_store.py:283-312`). It reimplements header/payload splitting and base64url decode that `verify_registry_token` also does. Could extract shared parsing to avoid drift.

- **Config: `registry_token_migration_cutoff_ts` loaded as float but config may provide int** (`tinyagentos/config.py:203`). `float(data["registry_token_migration_cutoff_ts"])` works for both but explicit `float()` on int is fine. No issue.

- **Missing test: Renewal endpoint with deactivated agent** (`tests/test_agent_registry.py`). Should verify 403 when agent status != active.

jaylfc:
I have converted this to DRAFT so it cannot be merged by accident. As written, deploying it takes the entire fleet offline at once, and the agents cannot recover themselves. Please do not undraft until blockers 1 to 5 are done.

Credit where it is due first: the store-level crypto is correct and genuinely red-first. I mutated the expiry enforcement four different ways and each one went red, including deleting the exp block entirely (3 failures) and forcing the migration window permanently closed (1 failure). That is the part the card cared most about and you delivered it properly.

BLOCKING 1, the renewal path does not work for agents. POST /api/agents/registry/token/renew is not in _AGENT_TOKEN_PATHS in auth_middleware.py, so the middleware rejects it before the handler runs. Proven with a cookie-less bearer-only client against the real app: 401 Authentication required. Your tests pass because the registry_client fixture carries an ADMIN SESSION COOKIE, so they authenticate as the human admin, not as the agent. Renewal therefore requires exactly the manual human re-mint the card said must not be required. Add the route to the allowlist and add a test with a cookie-less client.

BLOCKING 2, the migration window defaults to ZERO. registry_token_migration_cutoff_ts defaults to None, which means allow_no_exp_until is None, which means every legacy token without exp is rejected on the first request after deploy. Every agent on the Pi, every internal driver identity, everything on the A2A bus goes to 401 simultaneously, and per blocker 1 they cannot renew out of it. Ship a non-None effective default (a computed now+30d at first start, or a documented mandatory upgrade step). Not None.

BLOCKING 3, the operator escape hatch is not durable. AppConfig.to_dict() does not emit either new field, and save_config writes yaml.dump(to_dict()). There are 11 save_config call sites, so an operator who sets a migration cutoff loses it the next time anything touches settings, mid-migration and silently. Add both fields to to_dict() plus a load-save-load round-trip test.

BLOCKING 4, renewal re-creates the eternal credential this PR exists to remove. renew_registry_token does a signature-only check with no time bound at all. Probed: a token that expired TEN YEARS ago renews into a fresh 24h token, and a legacy no-exp token renews regardless of the cutoff. So a leaked bearer is still permanent, the attacker just renews forever. Bound it: reject tokens expired beyond a configurable grace, and pass allow_no_exp_until through.

BLOCKING 5, this conflicts with hognek's PR 2208 in both senses. Semantically, 2208 revokes by rejecting tokens whose iat is older than token_min_iat; your renewal mints iat=now from a signature-only check that never consults it, so a rotated-out token can be laundered into a valid post-rotation one, defeating rotation entirely. It does not bite today only because renewal is unreachable, so fixing blocker 1 without this opens it. They must land together. Textually you also conflict in routes/agent_registry.py and tests/test_agent_registry_store.py.

SHOULD FIX: registry_token_lifetime_seconds is read by nothing, so the configurable lifetime is inert (probed: config set to 60, minted lifetime still 86400) - wire it into all five mint/renew sites or delete the field. Add clock-skew leeway (60s) to both comparisons; today exp+2s is rejected, which will cause spurious 401s across machines. Add HTTP-layer migration-window tests, because mutating the cutoff getter to infinity currently leaves 222 tests green. Guard the numeric coercions in load_config, since an ISO date string in a field named _ts crashes taOS at startup. Audit-log renewal the way minting already is. Refresh user_id/framework/project_id from the registry record rather than replaying the old token's claims. And three docstrings you edited still say the token carries no exp claim.

Also please revert the unrelated comment typo in tests/test_agent_registry_store.py where metacharacters became metachrs.

jaylfc:
**nemotron-ultra-orB review** 

VERDICT: Code review complete - several correctness and test issues found.

- **tests/test_agent_registry_store.py:270** - `_build_token_no_exp` uses `_b64url_encode` which is not imported (private function in `agent_registry_store.py`). Tests will fail with `NameError`.
- **tinyagentos/agent_registry_store.py:280-310** - `_verify_signature_only` does not validate the JWT header `alg` field; a token with `{"alg":"none"}` would pass signature verification (though EdDSA verify would fail, explicit alg check is defense-in-depth).
- **tinyagentos/routes/agent_registry.py:744-778** - Renewal endpoint checks agent status *after* minting new token. If agent is deactivated, a token is minted then discarded (minor inefficiency, not a security issue).
- **tests/test_agent_registry.py:580-620** - Missing integration test for renewal with an *expired* token (the primary use case). Current test only uses a fresh token.
- **tinyagentos/agent_token_auth.py:103,185** - Migration cutoff defaults to `None`, causing immediate rejection of tokens without `exp` claim. This is a breaking change for existing deployments unless `registry_token_migration_cutoff_ts` is explicitly configured.
- **tinyagentos/config.py:196** - No validation that `registry_token_migration_cutoff_ts` is a reasonable future timestamp if provided.
- **tinyagentos/agent_registry_store.py:395** - `verify_registry_token` parameter `allow_no_exp_until` default `None` combined with `agent_token_auth.py` passing `None` means migration window is closed by default. Document this clearly or flip default to a far-future timestamp for smoother migration.

jaylfc:
**nemotron-super review** 

jaylfc:
**nemotron-super review** 

jaylfc:
## Not merging: this PR's green is 47 commits older than the branch it would land in.

Automated triage flagged this as a verify-then-merge candidate — its only red is the known Kilo Code Review rate-limit flake — and my merge gate agrees:

```
bash gate_merge.sh --check 2235
GATE PASS (check-only): PR #2235 passes every gate and WOULD merge.   EXIT=0
```

Both are wrong, for a reason neither of them checks:

```
latest check completed : 2026-08-02T14:34:12Z
origin/dev head        : 2026-08-05T04:25:59Z
dev is ahead by        : 47 commits
```

Every check on this head passed against a `dev` that no longer exists. `mergeStateStatus` reports `UNSTABLE` rather than conflicted, so there is no textual conflict — but "merges cleanly" and "was verified against what it merges into" are different claims, and only the first one has been established.

That matters more here than it usually would. This PR changes **registry token expiry** — `taOS mints JWTs that never expire`. Landing a security change whose entire verification predates 47 commits of the target branch is the case where a stale green is least acceptable.

**What it needs:** rebase onto current `dev` (or push to re-trigger), so the checks run against the base it will actually merge into. The work itself is not in question — only the age of its evidence.

Logged as the second live instance on the base-staleness card `tsk-qmgvwt`, alongside #2187 (113 commits behind). That card now has a complete fixture set — two red cases and one green — so the rule can be implemented and proven in both directions rather than relying on someone noticing by hand, which is what happened here.

jaylfc:
Closing: this draft has been on a DO-NOT-MERGE hold and CONFLICTING, and its card tsk-5bup3r is already closed. It has been the standing Section B row on every orphan check run. Closing resolves that row and frees a fleet throttle slot; if the work is still wanted it should be re-cut from current origin/dev rather than rebased from here.

PR #2236, opened 2026-08-02
check_doc_gate: add opt-in on_modify per-rule and extend tests

jaylfc:
Autonomous build of board card tsk-sxuexd.

evaluate_rules now honours an optional on_modify flag on each rule
(default false) so plain modifications can trigger the gate without
changing the behaviour of existing rules. Added three rules covering
user-visible behaviour, agent-facing behaviour, and CI/packaging
documentation, each with on_modify = true. Covered the five required
test cases (M-only fail, M-only with doc pass, M-only with trailer
pass, M-only default-still-pass, A/D still-work) plus edge cases
for test-path exclusion and mixed rules.

Files:
 docs/doc-gate.toml           |  28 ++++++--
 scripts/check_doc_gate.py    |  35 +++++-----
 tests/test_check_doc_gate.py | 157 +++++++++++++++++++++++++++++++++++++++++++
 3 files changed, 201 insertions(+), 19 deletions(-)

## Summary by CodeRabbit

* **Documentation**
  * Updated documentation requirements to cover user-facing changes, agent-facing components, CI and packaging updates, and contribution rules.
  * Documentation checks now recognize relevant modified files and accept corresponding documentation or changelog entries.
* **Bug Fixes**
  * Improved consistency in documentation validation while preserving exclusions for test-only changes.
* **Tests**
  * Expanded coverage for documentation checks involving modified, added, and deleted files.

### DIFF
--- a/docs/doc-gate.toml
+++ b/docs/doc-gate.toml
@@ -4,10 +4,10 @@
 # entries here to cover new feature areas without touching the script.
 #
 # Precision over recall: `when_changed` only triggers a rule on a structural
-# change (a file added or deleted), never a plain modification, and every
-# glob below is scoped to the specific tree it names. This is deliberate --
-# a noisy gate gets disabled, so false positives are worse than an occasional
-# missed doc update.
+# change (a file added or deleted, or a plain modification when the rule sets
+# `on_modify = true`), and every glob below is scoped to the specific tree it
+# names. This is deliberate -- a noisy gate gets disabled, so false positives
+# are worse than an occasional missed doc update.
 
 [gate]
 trailer = "Docs-Reviewed:"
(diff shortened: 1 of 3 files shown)


jaylfc:
Verified independently rather than taking the tests' word for it, since this is a gate and a gate that has only been seen passing is unproven. I re-ran the exact probe that showed MY OWN version of these rules was inert, against your branch, in both directions:

RED, all three fire where mine did nothing:
- M-only change to tinyagentos/routes/themes.py with no doc -> exit 1
- M-only change to .github/workflows/ci.yml with no doc -> exit 1
- M-only change to tinyagentos/agent_token_auth.py with no doc -> exit 1

GREEN, no false positives and the default is genuinely unchanged:
- same route change WITH a CHANGELOG entry -> exit 0
- same route change WITH a 'Docs-Reviewed:' trailer -> exit 0
- M-only change to a file matched only by a rule WITHOUT on_modify -> exit 0

8 tests pass locally. That last probe is the one I care most about: it proves the opt-in default really is off, so no existing PR suddenly turns red.

This closes a real gap. Before it, the gate caught a NEW undocumented route but was completely blind to a behaviour change in an existing one, which is the common case and the one Jay actually asked for. It is also the gap that let today's device-bearer PR ship while docs/agent-coordination.md still said browser sessions and agent JWTs are not accepted on that surface.

Merging on green CI. Two follow-ups, neither blocking: the sibling repos should adopt this once it lands (cards tsk-nwg6ef for taosmd and tsk-fe6bi3 for taos-website, both now carrying a correct 'REPO:' first line), and it is worth watching the first week of real PRs for rules that are too broad, since a gate that cries wolf gets routed around rather than obeyed.

jaylfc:

VERDICT: Approved — clean implementation with thorough tests and no regressions.

- No blocking issues found.
VERDICT: Approved — clean implementation with thorough tests and no regressions.

- No blocking issues found.

PR #2237, opened 2026-08-02
reserve user-, human-, admin-, taos- prefixes at agent registration

jaylfc:
Autonomous build of board card tsk-qvq7r6.

Choosing rejection with a clear ValueError over deterministic re-prefixing,
because explicit refusal is simpler and prevents accidental namespace
collisions. The check lives in the single register() choke point so no
caller can bypass it.

A new find_reserved_prefix_identities() method surfaces any pre-existing
collisions for manual review without auto-renaming live rows.

Audit of the codebase and test data found no identities already matching a
reserved prefix.

Files:
 tests/test_agent_registry_store.py  | 79 +++++++++++++++++++++++++++++++++++++
 tinyagentos/agent_registry_store.py | 53 ++++++++++++++++++++++++-
 2 files changed, 131 insertions(+), 1 deletion(-)

## Summary by CodeRabbit

* **New Features**
  * Registration now blocks names using reserved identity prefixes, including names that normalize to those prefixes.
  * Added an audit capability to identify existing records with reserved-prefix identities without modifying them.

* **Bug Fixes**
  * Improved registration consistency by validating the canonical identifier derived from the agent’s display name or framework.

### DIFF
--- a/docs/agent-coordination.md
+++ b/docs/agent-coordination.md
@@ -235,6 +235,13 @@ further project via `POST /api/projects/{project_id}/members/assign-agent`
 active identity (the existing canonical_id and token are reused instead of
 409ing).
 
+Reserved name prefixes: registration rejects any name whose slug is or starts
+with `user-`, `human-`, `admin-` or `taos-` (including casing, spacing and
+punctuation obfuscations like `U s e r`), so an external agent cannot mint an
+identity that reads as a person or as an internal taOS agent. The public
+register route returns 422. The admin-only internal mint/seed path is exempt -
+internal driver agents (`taos-dev`, ...) legitimately live under `taos-`.
+
 ## Device bearer self-service (second, narrower passthrough)
 
 Beyond the `EXEMPT_PATHS` entry for `GET /api/share/destinations`, a paired
--- a/tests/test_agent_internal_mint.py
+++ b/tests/test_agent_internal_mint.py
@@ -38,7 +38,7 @@ class TestGetByHandle:
     async def test_returns_active_match(self, tmp_path):
         s = await self._store(tmp_path / "reg.db")
         try:
-            rec = await s.register(framework="taos-internal", display_name="taos-dev", handle="@taOS-dev")
+            rec = await s.register(framework="taos-internal", display_name="taos-dev", handle="@taOS-dev", allow_reserved=True)
             got = await s.get_by_handle("@taOS-dev")
             assert got is not None
             assert got["canonical_id"] == rec["canonical_id"]
@@ -200,6 +200,9 @@ class TestMintInternalRoute:
         rec = await mint_client._app.state.agent_registry.register(
             framework="claude-code", display_name="taos-dev",
             origin="external-selfjoin", handle="@taOS-dev",
+            # Simulates a row minted BEFORE the reserved-prefix era; the guard
+            # (rightly) blocks creating this via register() today.
+            allow_reserved=True,
         )
         # external-selfjoin starts 'pending'; the real @taOS-dev was approved -> active.
         await mint_client._app.state.agent_registry.set_status(rec["canonical_id"], "active")
@@ -228,6 +231,9 @@ class TestMintInternalRoute:
         rec = await mint_client._app.state.agent_registry.register(
             framework="claude-code", display_name="taos-dev",
             origin="external-selfjoin", handle="@taOS-dev",
+            # Simulates a row minted BEFORE the reserved-prefix era; the guard
+            # (rightly) blocks creating this via register() today.
+            allow_reserved=True,
         )
         await mint_client._app.state.agent_registry.set_status(rec["canonical_id"], "active")
         with patch("tinyagentos.routes.agent_registry._audit_governance",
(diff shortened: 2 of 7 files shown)


jaylfc:
Verified independently and this one is clean. Merging on green CI.

I probed the guard directly rather than trusting the suite, given how many tests today have passed while testing the wrong thing. Sixteen inputs through the real _slugify and _check_reserved_prefix:

BLOCKED, correctly: User, USER, user, uSeR, 'user!', '  user  ' (all slugify to 'user'), user-bot, Admin, taos, human.
ALLOWED, also correctly: Userland, username, users, 'U s e r' (slugifies to u-s-e-r), normal-agent, Jay.

That second list is why I am satisfied. The guard is scoped to exactly what would produce a user- prefixed canonical id, and does not over-block lookalikes. 'username' becomes username-YYYYMMDD-HHMMSS, which does not start with 'user-' and so is not classified as human downstream; blocking it would have been a false positive that annoyed real users for no security gain. Getting that boundary right is the difference between a guard people keep and one they route around.

Mutation-tested: neutering _check_reserved_prefix to a no-op turns 6 tests red. 139 pass unmutated.

One note for whoever picks up the taosmd side: this closes the taOS half, so taOS can no longer mint an identity in a namespace it does not own. The taosmd half still needs its human-principal branch to fail closed when human_iss is unset, which is item 4 on the tsk-legqtr rework list. Neither is sufficient alone: taOS must not mint into the namespace, and taosmd must not trust the prefix without verification.

jaylfc:

VERDICT: The reserved-prefix guard is a sensible defense-in-depth measure, but the implementation has a logic gap in the raw-name normalization and the tests miss a framework-name collision case.

- **tinyagentos/agent_registry_store.py:386** The raw-name normalization `re.sub(r"[^a-z0-9]", "", raw_name.lower())` removes hyphens, so a display name like `"user-agent"` normalizes to `"useragent"` which does **not** equal bare `"user"` — the reserved check passes even though the slug `"user-agent"` would be caught by the slug `startswith(prefix)` check. This is harmless (double-checked) but the docstring claim "obfuscate with punctuation" is misleading; only spacing/punctuation that fully collapses to the bare word is caught.

- **tinyagentos/agent_registry_store.py:495** When `display_name=""` (empty string), `source_name` falls back to `framework`. If `framework` is a reserved word (e.g., `"user"`), the guard triggers — but there is **no test** covering this path.

- **tests/test_agent_registry_store.py:412** `test_find_reserved_prefix_identities_surfaces_collisions` inserts rows with `user_id=""` — verify the schema allows empty `user_id` or the test will fail on constraint violation.

- **tinyagentos/agent_registry_store.py:699** `find_reserved_prefix_identities` runs one query per prefix (4 total). Acceptable for 4 prefixes, but consider a single `WHERE canonical_id LIKE 'user-%' OR canonical_id LIKE 'human-%' ...` for maintainability.

- **tests/test_agent_registry_store.py:389** `test_bypass_spacing_rejected` relies on raw-name normalization; if `_slugify("U s e r")` produces `"u-s-e-r"` instead of `"user"`, the slug check passes and only the raw-name check rejects. This works but is an implicit dependency on the normalization behavior — add a comment or assertion.

No blocking issues found.
VERDICT: The reserved-prefix guard is a sensible defense-in-depth measure, but the implementation has a logic gap in the raw-name normalization and the tests miss a framework-name collision case.

- **tinyagentos/agent_registry_store.py:386** The raw-name normalization `re.sub(r"[^a-z0-9]", "", raw_name.lower())` removes hyphens, so a display name like `"user-agent"` normalizes to `"useragent"` which does **not** equal bare `"user"` — the reserved check passes even though the slug `"user-agent"` would be caught by the slug `startswith(prefix)` check. This is harmless (double-checked) but the docstring claim "obfuscate with punctuation" is misleading; only spacing/punctuation that fully collapses to the bare word is caught.

- **tinyagentos/agent_registry_store.py:495** When `display_name=""` (empty string), `source_name` falls back to `framework`. If `framework` is a reserved word (e.g., `"user"`), the guard triggers — but there is **no test** covering this path.

- **tests/test_agent_registry_store.py:412** `test_find_reserved_prefix_identities_surfaces_collisions` inserts rows with `user_id=""` — verify the schema allows empty `user_id` or the test will fail on constraint violation.

- **tinyagentos/agent_registry_store.py:699** `find_reserved_prefix_identities` runs one query per prefix (4 total). Acceptable for 4 prefixes, but consider a single `WHERE canonical_id LIKE 'user-%' OR canonical_id LIKE 'human-%' ...` for maintainability.

- **tests/test_agent_registry_store.py:389** `test_bypass_spacing_rejected` relies on raw-name normalization; if `_slugify("U s e r")` produces `"u-s-e-r"` instead of `"user"`, the slug check passes and only the raw-name check rejects. This works but is an implicit dependency on the normalization behavior — add a comment or assertion.

No blocking issues found.

jaylfc:
The red was a real conflict with the internal-mint work that landed on dev: driver agents are deliberately named under taos-, so the new guard broke mint-internal/seed-internal on exactly the names the prefix is reserved for. Fixed by an explicit allow_reserved keyword on register() that only the admin-only internal mint call site passes - not reachable from any request body, so the public route still rejects every reserved name, now as a 422 instead of an unhandled 500 (handler removed = test goes red, verified). Merged dev in while I was there.

PR #2258, opened 2026-08-03
Unit tests for restart_orchestrator

jaylfc:
CARD TITLE (intent, not commit subject): Unit tests for restart_orchestrator

Autonomous build of board card tsk-vigony.

Files:
 tests/test_restart_orchestrator.py | 587 +++++++++++++++++++++++++++++++++++++
 1 file changed, 587 insertions(+)

### DIFF
--- a/tests/test_restart_orchestrator.py
+++ b/tests/test_restart_orchestrator.py
@@ -1,0 +1,587 @@
+"""Unit tests for tinyagentos.restart_orchestrator.
+
+Tests cover pure logic and public functions/methods that can run in-process.
+External services (network, containers, GPU, LLM) are mocked.
+"""
+from __future__ import annotations
+
+import asyncio
+import json
+import os
+import time
+from pathlib import Path
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+import tinyagentos.restart_orchestrator as ro
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+def _app_state(tmp_path, *, agents=None, data_dir=None):
+    """Minimal app_state namespace for tests."""
+    if agents is None:
+        agents = []
+    if data_dir is None:
+        data_dir = tmp_path / "data"
+        data_dir.mkdir(parents=True, exist_ok=True)
+    notifications = SimpleNamespace(add=AsyncMock())
+    config = SimpleNamespace(
+        agents=agents,
+        config_path=tmp_path / "config.yaml",
+    )
+    return SimpleNamespace(
+        config=config,
+        data_dir=data_dir,
+        notifications=notifications,
+        _background_tasks=set(),
+    )
+
+
+@pytest.fixture(autouse=True)
+def _no_config_write(monkeypatch):
+    """Prevent real config writes during tests."""
+    import tinyagentos.config as cfg
+
+    monkeypatch.setattr(cfg, "save_config_locked", AsyncMock())
+
+
+# ---------------------------------------------------------------------------
+# _pending_restart_path
+# ---------------------------------------------------------------------------
+
+class TestPendingRestartPath:
+    def test_uses_env_var(self, monkeypatch, tmp_path):
+        monkeypatch.setenv("TAOS_DATA_DIR", str(tmp_path))
+        p = ro._pending_restart_path()
+        assert p == tmp_path / "pending-restart.json"
+
+    def test_uses_install_data_dir_when_available(self, monkeypatch, tmp_path):
+        monkeypatch.delenv("TAOS_DATA_DIR", raising=False)
+        install_root = tmp_path / "tinyagentos"
+        install_root.mkdir()
+        data_dir = install_root / "data"
+        data_dir.mkdir()
+        (tmp_path / "data").mkdir()
+        monkeypatch.setattr(ro, "__file__", str(install_root / "restart_orchestrator.py"))
+        p = ro._pending_restart_path()
+        assert p == tmp_path / "data" / "pending-restart.json"
+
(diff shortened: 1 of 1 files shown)


jaylfc:
Reviewed: purely additive new file, 37 tests over the restart orchestrator with the failure paths that actually matter for update safety - prepare timeout/exception status, controller-note fallback on host non-200/exception, post-restart SHA mismatch posts a warning and subprocess failure is treated as mismatch (fail-safe direction), save-failure keeps the note while still unpausing, and corrupt/missing note synthesis. Mocks sit at the genuine external boundaries (network, subprocess) while the note/status plumbing runs real against tmp_path. Merging on green.

jaylfc:

VERDICT: Tests are well-structured with good coverage but have flaky time-dependent assertions and some mock gaps.

- `tests/test_restart_orchestrator.py:494` — Flaky assertion `data["paused_at"] == int(time.time())` depends on exact test execution second; will fail across second boundaries
- `tests/test_restart_orchestrator.py:218-219` — Hardcoded timeout value `300` in test couples test to implementation detail; should reference the actual timeout constant from the module
- `tests/test_restart_orchestrator.py:213-224` and `227-238` — Tests create private `pytest.MonkeyPatch()` instances instead of using the fixture; unconventional but functional due to explicit `undo()`
- `tests/test_restart_orchestrator.py:135-140` — `_pending_restart_path` fallback chain not fully tested: no test for when `HOME` is unset and no install dir exists (would raise or return unexpected path)
- `tests/test_restart_orchestrator.py:537-545` — `test_hostless_agent_unpaused_without_resume_call` doesn't verify `_unpause` was actually invoked; only checks side effects
- `tests/test_restart_orchestrator.py` — Missing test for concurrent `prepare()` calls / idempotency; missing negative case for agent dict missing required keys (`name`, `host`, `port`)
VERDICT: Tests are well-structured with good coverage but have flaky time-dependent assertions and some mock gaps.

- `tests/test_restart_orchestrator.py:494` — Flaky assertion `data["paused_at"] == int(time.time())` depends on exact test execution second; will fail across second boundaries
- `tests/test_restart_orchestrator.py:218-219` — Hardcoded timeout value `300` in test couples test to implementation detail; should reference the actual timeout constant from the module
- `tests/test_restart_orchestrator.py:213-224` and `227-238` — Tests create private `pytest.MonkeyPatch()` instances instead of using the fixture; unconventional but functional due to explicit `undo()`
- `tests/test_restart_orchestrator.py:135-140` — `_pending_restart_path` fallback chain not fully tested: no test for when `HOME` is unset and no install dir exists (would raise or return unexpected path)
- `tests/test_restart_orchestrator.py:537-545` — `test_hostless_agent_unpaused_without_resume_call` doesn't verify `_unpause` was actually invoked; only checks side effects
- `tests/test_restart_orchestrator.py` — Missing test for concurrent `prepare()` calls / idempotency; missing negative case for agent dict missing required keys (`name`, `host`, `port`)

jaylfc:
**nemotron-ultra-orB review** 

VERDICT: The test file is comprehensive and well-structured, but has several correctness and style issues that should be fixed.

- `tests/test_restart_orchestrator.py:151-158`: `test_uses_install_data_dir_when_available` incorrectly constructs expected path — uses `tmp_path / "data"` but the logic derives from `ro.__file__` parent's `data` dir (install_root/data), not tmp_path/data. Test will pass only by accident of temp dir layout.
- `tests/test_restart_orchestrator.py:159-167`: `test_falls_back_to_home` has same issue — constructs fake_module under `tmp_path/tinyagentos/...` but expects `home/.config/taos/...`. The fallback logic uses `Path(ro.__file__).parent.parent / "data"` which would resolve to `tmp_path/tinyagentos/data`, not home dir.
- `tests/test_restart_orchestrator.py:267-277`: `test_prepare_timeout_sets_status` creates a new `pytest.MonkeyPatch()` instance instead of using the fixture `monkeypatch` — breaks isolation and leaks patches.
- `tests/test_restart_orchestrator.py:292-302`: `test_prepare_exception_sets_status` same issue — manual `MonkeyPatch()` instead of fixture.
- `tests/test_restart_orchestrator.py:403-404`: `test_host_200_uses_note_path_from_response` mocks `httpx.AsyncClient` as a context manager but patches the class, not the instance — `mock_cm.__aenter__` returns mock_client but the code uses `async with httpx.AsyncClient() as client:`, so `client` is the `__aenter__` return value. This works but is fragile; prefer patching `httpx.AsyncClient.post` directly like other tests.
- `tests/test_restart_orchestrator.py:433-434`, `463-464`: Same context manager mock pattern repeated — inconsistent with simpler `patch("httpx.AsyncClient.post")` used in `TestPostResume`.
- `tests/test_restart_orchestrator.py:507-508`: `test_subprocess_failure_treated_as_mismatch` patches `asyncio.create_subprocess_exec` but the code likely uses `asyncio.create_subprocess_exec` with `stdout=asyncio.subprocess.PIPE` — the mock returns a coroutine that raises, but the actual call signature expects specific args. Should verify the real function signature.
- `tests/test_restart_orchestrator.py:529-530`, `552-553`, `571-572`: `test_save_failure_keeps_note_and_unpauses`, `test_note_none_still_saves` patch `"tinyagentos.config.save_config_locked"` as string — should patch the imported reference in `restart_orchestrator` module (i.e., `ro.save_config_locked` or `tinyagentos.restart_orchestrator.save_config_locked`) since the module imports it locally.
- Missing test coverage: No tests for `_prepare_agent` timeout (5min default), no test for `resume_agents_from_notes` when `_post_resume` returns False (network failure path), no test for concurrent prepare/execute race conditions, no test for `apply_pending_restart_check` when git subprocess returns non-zero exit code.

jaylfc:
**nemotron-super review** 

VERDICT: Blocking issue found: flaky test due to use of time.time() without mocking.
- tests/test_restart_orchestrator.py:412

PR #2266, opened 2026-08-03
Deploy route registers deployed agents into agent_registry (dec-h3jlcy)

jaylfc:
CARD TITLE (intent, not commit subject): Deploy route registers deployed agents into agent_registry (dec-h3jlcy)

Autonomous build of board card tsk-qdvfg4.

Files:
 tests/test_routes_agents.py  | 173 +++++++++++++++++++++++++++++++++++++++++++
 tinyagentos/routes/agents.py |  31 ++++++++
 2 files changed, 204 insertions(+)

## Summary by CodeRabbit

* **New Features**
  * Deployed agents are now registered with the agent registry and assigned a persistent canonical identity.
  * Agents with duplicate display names receive distinct identities.

* **Bug Fixes**
  * Deployments using reserved names now return a clear HTTP 400 error without creating an agent.
  * Other registry failures return an appropriate server error while allowing pending deployment requests to complete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
 
 - Observatory fleet view for agents holding a global `observatory_control`
   grant; project-scoped grants see only their granted projects (#2267).
+- Deployed agents are registered into the agent registry at deploy time, each
+  minting its own canonical identity; names that resolve to a reserved prefix
+  are rejected with a 400 (#2266).
 
 ### Added
 
--- a/tests/test_routes_agents.py
+++ b/tests/test_routes_agents.py
@@ -620,6 +620,179 @@ class TestDeployPersistence:
 
 
 @pytest.mark.asyncio
+class TestDeployRegistryRegistration:
+    """Deploy endpoint must mint an agent_registry canonical_id on first deploy
+    and skip re-registration on redeploy / model swap."""
+
+    async def test_deploy_creates_registry_row(self, client, app, monkeypatch):
+        import re
+        import taosmd.agents as tm_agents
+        from unittest.mock import AsyncMock, MagicMock
+
+        app.state.archive = MagicMock(
+            record=AsyncMock(), query=AsyncMock(return_value=[{}])
+        )
+
+        register_calls = []
+        canonical_ids = []
+
+        async def fake_register(*, framework, display_name, **kw):
+            register_calls.append((framework, display_name))
+            cid = f"{display_name.lower().replace(' ', '-')}-20260101-000000"
+            canonical_ids.append(cid)
+            return {"canonical_id": cid, "framework": framework, "display_name": display_name}
+
+        mock_registry = MagicMock()
+        mock_registry.register = fake_register
+        mock_registry.get = AsyncMock(return_value=None)
+        app.state.agent_registry = mock_registry
+
+        def fake_register_agent(name, **kwargs):
+            pass
+        monkeypatch.setattr(tm_agents, "register_agent", fake_register_agent)
+
+        async def fake_deploy(req):
+            return {"success": True, "name": req.name, "ip": "10.0.0.42",
+                    "llm_key": "sk-test", "steps": ["deployment_complete"],
+                    "container": f"taos-agent-{req.name}"}
+        monkeypatch.setattr("tinyagentos.deployer.deploy_agent", fake_deploy)
+
+        class _FakeCatalog:
+            def all_models(self, capability=None):
+                return [{"name": "test-model", "id": "test-model"}]
+        app.state.backend_catalog = _FakeCatalog()
+        app.state.cluster_manager._workers.clear()
+
+        resp = await client.post("/api/agents/deploy", json={
+            "name": "RegistryTest",
+            "framework": "none",
+            "model": "test-model",
+        })
+        assert resp.status_code == 200
+        import asyncio
+        await asyncio.sleep(0.2)
+
+        assert len(register_calls) == 1
+        assert re.match(r"^[a-z0-9-]+-\d{8}-\d{6}(-\\d{2})?$", canonical_ids[0])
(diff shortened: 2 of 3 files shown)


jaylfc:

VERDICT: Several correctness and reliability issues; tests are flaky and production code has undefined variable references.

- **tests/test_routes_agents.py:649** - Flaky test uses `await asyncio.sleep(0.2)` instead of proper async synchronization; will fail under load or CI variance
- **tests/test_routes_agents.py:647** - Regex pattern `r"^[a-z0-9-]+-\d{8}-\d{6}(-\\d{2})?$"` has double-escaped `\\d` in raw string; matches literal `\d` not digits
- **tinyagentos/routes/agents.py:581** - `display_name` used but not defined in diff context (likely defined earlier, but verify it's set before this block)
- **tinyagentos/routes/agents.py:582** - `config` used but not shown in diff; verify it's `request.app.state.config` or similar
- **tinyagentos/routes/agents.py:592** - `unique_slug` referenced in logger but not defined in this scope; will cause NameError at runtime
- **tinyagentos/routes/agents.py:593** - Error response leaks exception details: `f"Could not register agent in registry: {e}"` — security info leak
- **tinyagentos/routes/agents.py:580** - `getattr(ar, "_db", None)` leaks implementation detail; prefer a public `is_ready()` or similar method
- **tests/test_routes_agents.py** - No test for `ar.register()` raising exception (error path at line 590-596)
- **tests/test_routes_agents.py** - No test for `ar.get()` raising exception
- **tinyagentos/routes/agents.py:596** - `scoped_key` and `idempotency_cache` used but not shown in diff; verify they're in scope
VERDICT: Several correctness and reliability issues; tests are flaky and production code has undefined variable references.

- **tests/test_routes_agents.py:649** - Flaky test uses `await asyncio.sleep(0.2)` instead of proper async synchronization; will fail under load or CI variance
- **tests/test_routes_agents.py:647** - Regex pattern `r"^[a-z0-9-]+-\d{8}-\d{6}(-\\d{2})?$"` has double-escaped `\\d` in raw string; matches literal `\d` not digits
- **tinyagentos/routes/agents.py:581** - `display_name` used but not defined in diff context (likely defined earlier, but verify it's set before this block)
- **tinyagentos/routes/agents.py:582** - `config` used but not shown in diff; verify it's `request.app.state.config` or similar
- **tinyagentos/routes/agents.py:592** - `unique_slug` referenced in logger but not defined in this scope; will cause NameError at runtime
- **tinyagentos/routes/agents.py:593** - Error response leaks exception details: `f"Could not register agent in registry: {e}"` — security info leak
- **tinyagentos/routes/agents.py:580** - `getattr(ar, "_db", None)` leaks implementation detail; prefer a public `is_ready()` or similar method
- **tests/test_routes_agents.py** - No test for `ar.register()` raising exception (error path at line 590-596)
- **tests/test_routes_agents.py** - No test for `ar.get()` raising exception
- **tinyagentos/routes/agents.py:596** - `scoped_key` and `idempotency_cache` used but not shown in diff; verify they're in scope

jaylfc:
**nemotron-ultra-orB review** 

VERDICT: Multiple correctness bugs and test flaws blocking merge

- **BUG tinyagentos/routes/agents.py:586** — Lookup uses `display_name` (user-facing) instead of unique `name` (slug). Multiple agents can share a display_name; this causes wrong canonical_id reuse or false "already registered" skips. Should match on `a.get("name") == unique_slug`.

- **BUG tinyagentos/routes/agents.py:580-598** — Check-then-act race: two concurrent deploys for same agent both see `canonical_id=None`, both call `register()`, creating duplicate registry rows. The registry's collision guard only protects canonical_id uniqueness, not display_name/framework combo.

- **TEST FLAW tests/test_routes_agents.py:640,680,720** — `asyncio.sleep(0.2)` flaky wait for background deploy. Use proper async synchronization or mock the deploy to run synchronously.

- **TEST FLAW tests/test_routes_agents.py:636,676,716** — Mock registry missing `_db` attribute. `MagicMock()` auto-creates truthy `_db`, so `getattr(ar, "_db", None) is not None` passes but doesn't test real init check. Real store raises `RuntimeError` if `_db is None`. Tests should set `mock_registry._db = object()` or use a real store.

- **MISSING TEST** — No test for `agent_registry` present but uninitialized (`_db is None`). Code silently skips registration; should either test this path or fail fast.

- **STYLE tinyagentos/routes/agents.py:595** — Error log references `unique_slug` but `register()` call uses `display_name` + `framework`. Misleading for debugging.

- **MINOR tinyagentos/routes/agents.py:596-600** — Registry failure caches error under idempotency key. Retry with same key returns stale registry error even if registry recovered. Consider not caching this specific error or using separate key.

jaylfc:
Reviewed at 5237fa44; defect found and lead-completed at 02887839.

The mint path and the first test were on-spec: deploy registers the agent and stores registry_canonical_id, guard + failure path match the taosmd registration idiom above it.

The 'idempotent on redeploy' branch was an identity-collision defect. Display names are not unique (validate_agent_name allows duplicates; unique_agent_slug suffixes -2, -3...), and every deploy creates a NEW config entry, so the display_name match could only ever hit a DIFFERENT agent - handing a second agent named 'Helper' the first Helper's canonical_id. canonical_id is the JWT sub and the memory-namespace key; two live agents must never share one. Two of the three tests pinned that behavior as desired (the 'model swap' test was also a same-name second deploy - model swaps don't go through this route at all).

Fix: always mint on deploy; reserved-prefix ValueError from the registry now returns 400 (user error) instead of 500. Tests rewritten: same-display-name deploys must produce two register calls and DISTINCT stored ids (red-proven against 5237fa44 - it fails with the shared id), reserved-name deploy returns 400 with the registry message and adds no agent (red-proven: was 500). Also fixed the escaped-backslash regex in the format assertion ((-\\d{2})? matched a literal backslash-d, never a real collision suffix).

Full tests/test_routes_agents.py: 71 passed on the branch.

Follow-up noted, not blocking: a register that succeeds followed by a later deploy failure leaves an orphaned registry row (same shape as the taosmd registration above); registry hygiene sweep can reap rows no config entry references.

Auto-merge armed on green.

PR #2131, opened 2026-07-26
feat(feedback): add per-user 24h submission cap

jaylfc:
Autonomous build of board card tsk-tlkemm.

Files:
 tests/test_feedback_route.py   | 48 +++++++++++++++++++++++++++++++++++++++++-
 tinyagentos/feedback_store.py  | 13 ++++++++++++
 tinyagentos/routes/feedback.py | 13 +++++++++++-
 3 files changed, 72 insertions(+), 2 deletions(-)

PR #2269, opened 2026-08-03
Fix decorative test from #2131: rate_limit_per_user passes with no rate limit at all

jaylfc:
CARD TITLE (intent, not commit subject): Fix decorative test from #2131: rate_limit_per_user passes with no rate limit at all

Autonomous build of board card tsk-p7fity.

The test asserted 201 for every request, so it passed whether or not the
rate limit existed. Add the missing assertion that a further admin post
after hitting the cap returns 429, while user2's first post still returns
201. This pair is the actual per-user scoping claim.

Files:
 tests/test_feedback_route.py | 6 ++++++
 1 file changed, 6 insertions(+)

### DIFF
--- a/tests/test_feedback_route.py
+++ b/tests/test_feedback_route.py
@@ -182,6 +182,12 @@ async def test_post_feedback_rate_limit_per_user(client, app):
         )
         assert resp.status_code == 201
 
+    resp = await client.post(
+        "/api/feedback",
+        json={"type": "bug", "title": "Admin over limit", "body": ""},
+    )
+    assert resp.status_code == 429
+
     invite_code = app.state.auth.add_user_invite("user2", "admin")
     app.state.auth.complete_invite("user2", invite_code, "User Two", "", "password")
     record = app.state.auth.find_user("user2")


jaylfc:
Reviewed at 8ae6c94d. ON-SPEC, textbook Distrust-Green fix.

The original test set MAX_FEEDBACK_PER_USER_PER_DAY, posted exactly up-to-limit (all 201), then checked a SECOND user was independently limited - but never asserted the first user gets 429 when over. So the per-user limiter could have been entirely dead and the test stayed green (the exact decorative-test class in our CLAUDE.md). The added 6 lines post one past the limit and assert 429.

Red-first proven locally: neutered the limiter (feedback.py:87 if item is None -> if False) and the new assertion fails; reverted, passes. Diff is +6/-0, one focused assertion, no scope creep.

Auto-merge armed on green.

jaylfc:
**nemotron-ultra-orB review** 

VERDICT: Test has a critical logic bug - it tests admin rate limiting without authenticating as admin first.

- `tests/test_feedback_route.py:188-193`: New test expects 429 for "Admin over limit" but makes unauthenticated request. No login/setup for admin user exists in this test block. The prior test creates `user2` (non-admin) after this assertion, so this request runs as anonymous or previous user context. Test will likely fail or test wrong behavior.

PR #2274, opened 2026-08-03
Device token write passthrough for share targets (currently discovery-only)

jaylfc:
CARD TITLE (intent, not commit subject): Device token write passthrough for share targets (currently discovery-only)

Autonomous build of board card tsk-avjhlv.

Files:
 tinyagentos/auth_middleware.py             |  21 ++--
 tinyagentos/routes/chat.py                 |  22 +++-
 tinyagentos/routes/decisions.py            |  70 +++---------
 tinyagentos/routes/library.py              |   8 +-
 tinyagentos/routes/project_files.py        |  29 ++++-
 tinyagentos/taosnet/mesh.py                |   3 +-
 tinyagentos/taosnet/mesh_credentials.py    |   3 +-
 12 files changed, 258 insertions(+), 146 deletions(-)

### DIFF
--- a/tests/test_device_share_write.py
+++ b/tests/test_device_share_write.py
@@ -1,0 +1,176 @@
+"""Tests proving a device token can complete share writes and cannot write elsewhere."""
+from __future__ import annotations
+
+import pytest
+from httpx import ASGITransport, AsyncClient
+
+from tinyagentos.app import create_app
+
+
+@pytest.fixture
+def app(tmp_data_dir):
+    return create_app(data_dir=tmp_data_dir)
+
+
+@pytest.mark.asyncio
+class TestDeviceShareWritePassthrough:
+    """Device bearer tokens may write to the three share destinations
+    advertised by /api/share/destinations and cannot reach any other route.
+    """
+
+    async def _register_device(self, client) -> str:
+        reg = await client.post("/api/devices/register", json={"platform": "ios"})
+        assert reg.status_code == 200
+        return reg.json()["scoped_token"]
+
+    async def _device_client(self, app, token):
+        transport = ASGITransport(app=app)
+        return AsyncClient(
+            transport=transport,
+            base_url="http://test",
+            headers={"Authorization": f"Bearer {token}"},
+        )
+
+    async def _user_id(self, app) -> str:
+        primary = app.state.auth.get_primary_user()
+        return primary["id"] if primary else "admin"
+
+    # -- Positive: device token can write to each share destination --
+
+    async def test_device_can_ingest_library_url(self, client, app, tmp_data_dir):
+        token = await self._register_device(client)
+        async with await self._device_client(app, token) as dc:
+            resp = await dc.post("/api/library/ingest", data={"url": "https://example.com/page"})
+            assert resp.status_code == 202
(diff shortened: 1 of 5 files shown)


jaylfc:
HELD for full security review, NOT armed - and NOT for beta.46 (that train is already bumped/promoting; this rides the next one).

Reviewed so far (at 189c2ebc):
- auth_middleware.py: adds exactly 3 POST routes to the device-bearer allowlist (library/ingest, chat/messages, projects/*/files/upload). The non-admin, user_id=None model is preserved and the allowlist is explicit and narrow - good shape.
- project_files.py: VERIFIED correct - the device path in _authorize_files_actor gates on project.user_id != device.user_id (404), the SAME ownership check as the session path, so a device can only write to its owner's projects. Does NOT touch decisions grant-approval, so it does not pre-empt the open decision dec-uwovql.

Remaining before I arm (the review this security surface earns):
1. chat/messages and library/ingest: confirm each applies the SAME owner-scoping as project_files - a device must not post to another user's channel or ingest into another user's library. If either trusts a body-supplied target id without checking device.user_id, that is the bug.
2. current_user_or_device: confirm it 401s a bad/expired device token and cannot be reached without request.state._device set.
3. Red-first: the new test_device_share_write.py must prove a device writing to a NON-owned target gets 404/403, not just the happy path.

Will complete this pass once beta.46 is cut. Good build - flagging thoroughly, not blocking on doubt.

jaylfc:
**nemotron-super review** 

VERDICT: FAIL
- tests/test_device_share_write.py: line 176: missing test case for device cannot upload files to another project's files

jaylfc:
Closing to rebuild - this is a REAL REGRESSION, not the usual doc-gate miss (though that is red too).

FAILING, and they are pre-existing tests, not new ones:
  tests/test_routes_project_files_agent_scope.py::TestAgentFilesWrite::test_upload_allowed_with_files_write
  tests/test_routes_project_files_agent_scope.py::TestAgentFilesWrite::test_upload_without_write_scope_is_403  (assert 401 == 403)

DIAGNOSIS: adding Depends(current_user_or_device) to api_project_upload_file made the dependency run BEFORE the handler. current_user_or_device resolves a SESSION first, else requires a DEVICE bearer via require_device - an AGENT registry JWT is neither, so it raises 401 and the request never reaches _authorize_files_actor, which is the code that honours a files_write grant. Net effect: agents can no longer upload to project files at all, and the 403-without-scope contract degrades to 401. You gated devices in correctly and broke agents out in the process.

FIX DIRECTION: the dependency must not hard-fail for a bearer it does not recognise. Make it FALL THROUGH (return None / leave request.state._device unset) when there is no session and no valid device token, and let _authorize_files_actor make the decision - it already handles all three actors (session owner/admin, device via request.state._device which your PR added, and agent JWT via the files scope). Do not special-case by inspecting the token prefix in the route.

KEEP: the auth_middleware allowlist addition and the project_files device ownership gate (project.user_id != device.user_id -> 404) are correct and I verified them; the same ownership check still needs proving for chat/messages and library/ingest, plus a red-first test that a device writing to a NON-owned target is refused. Guidance on the card tsk-avjhlv.

PR #2276, opened 2026-08-03
Library UI: settings pane (preferred quality + per-source rules)

jaylfc:
CARD TITLE (intent, not commit subject): Library UI: settings pane (preferred quality + per-source rules)

Autonomous build of board card tsk-g3olu7.

Files:
 CHANGELOG.md                                  |   4 -
 desktop/src/apps/LibraryApp.settings.test.tsx | 194 ++++++++++++++++++++
 desktop/src/apps/LibraryApp.tsx               | 247 +++++++++++++++++++++++++-
 docs/agent-manual/11-files-api.md             |  35 ----
 docs/agent-manual/index.md                    |   1 -
 docs/taos-agent-manual.md                     |  35 ----
 tests/test_agent_manual_compiled.py           |   2 +-
 7 files changed, 436 insertions(+), 82 deletions(-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
 
 ### Added
 
+- **Library**: settings pane for download preferences: preferred quality and per-source rules (#2276).
+
 - Observatory fleet view for agents holding a global `observatory_control`
   grant; project-scoped grants see only their granted projects (#2267).
 - Deployed agents are registered into the agent registry at deploy time, each
--- a/desktop/src/apps/LibraryApp.settings.test.tsx
+++ b/desktop/src/apps/LibraryApp.settings.test.tsx
@@ -1,0 +1,194 @@
+import { render, screen, fireEvent, waitFor, within } from "@testing-library/react";
+import { describe, it, expect, vi, afterEach, beforeEach } from "vitest";
+import { LibraryApp } from "./LibraryApp";
+
+const MOCK_ITEMS = [
+  {
+    id: "lib-item-1",
+    title: "YouTube Video",
+    source_type: "youtube",
+    source_url: "https://youtube.com/watch?v=1",
+    source_id: "yt-1",
+    author: "YT Author",
+    summary: "Summary 1",
+    content: "Content 1",
+    media_path: null,
+    thumbnail: null,
+    categories: [],
+    tags: [],
+    metadata: {},
+    status: "ready",
+    monitor: { current_interval: 0, frequency: 0, decay_rate: 0, pinned: false, last_poll: null, last_hash: "" },
+    created_at: 1700000000,
+    updated_at: 1700000000,
+  },
+];
+
+function createFetchMock() {
+  return vi.fn((input: RequestInfo | URL) => {
+    const url = String(input);
+    if (url.startsWith("/api/knowledge/items")) {
+      return Promise.resolve({
+        ok: true,
+        status: 200,
+        headers: new Map([["content-type", "application/json"]]),
+        json: () => Promise.resolve({ items: MOCK_ITEMS, count: MOCK_ITEMS.length }),
+      } as Response);
+    }
+    if (url === "/api/agents") {
+      return Promise.resolve({
+        ok: true,
+        status: 200,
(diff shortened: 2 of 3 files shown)


jaylfc:
**nemotron-super review** 

VERDICT: Found issues requiring fixes
- desktop/src/apps/LibraryApp.tsx:1226: Storage cap input allows zero/negative values, causing incorrect storage view calculation
- desktop/src/apps/LibraryApp.tsx:144: loadLibrarySettings lacks validation for parsed settings (type safety)
- desktop/src/apps/LibraryApp.tsx:1255: Non-unique key in rules table for duplicate rules (same source/action/quality)
- desktop/src/apps/LibraryApp.settings.test.tsx:46: Weak table selector (querySelector) risks false positives if multiple tables exist
- desktop/src/apps/LibraryApp.settings.test.tsx:64: Weak table selector (querySelector) risks false positives if multiple tables exist

jaylfc:
HELD, not armed - this builds the CLIENT half of a feature whose SERVER half is in an open PR, and as written the two never meet.

WHAT THIS PR DOES: the settings pane persists preferred quality and per-source rules to localStorage only (LIBRARY_SETTINGS_KEY). There are no API calls in the diff, and dev's tinyagentos/routes/library.py has no settings or rules endpoints at all - I grepped.

WHAT ALREADY EXISTS ELSEWHERE: hognek's OPEN #2070 'feat(library): P3 - heavy tier download, quality preferences, storage accounting, per-source rules' adds precisely the server side, including POST /api/library/rules and GET /api/library/rules, with rules carrying source_pattern, quality and auto_download - the same three concepts this pane edits.

CONSEQUENCE IF MERGED AS-IS: the pane would look complete and change nothing. The library pipeline downloads server-side, so rules kept in one browser's localStorage are never consulted; a user would set 720p for youtube, see it saved, and get whatever the server defaults to. That is the failure mode where the working half hides the broken half, and it is also the collision the contributing guide warns about - two PRs implementing one feature from opposite ends without knowing about each other.

NOT A CRITICISM OF THE BUILD: the pane itself is sound - 194 lines of real tests, no CSRF exposure precisely because it makes no network calls, and the only CI red besides Kilo's known rate-limit fake-red is the missing changelog.

TO LAND IT: wire the pane to #2070's endpoints - GET /api/library/rules to populate, POST /api/library/rules to save - and keep localStorage at most as an optimistic cache, never as the source of truth. That means this PR should either wait for #2070 to merge, or be rebased onto it. I am holding rather than closing because the UI work is worth keeping. Sequencing is mine to coordinate with hognek.

jaylfc:
@coderabbitai full review

jaylfc:
@coderabbitai full review

jaylfc:
**Lead review — APPROVED, removing `lead-blocked`.** Real review per the fake-green policy (CodeRabbit's status is a rate-limited pass that never ran; a retrigger is in flight but this review is the gate).

Scope is honest scaffolding: client half of #2070 against a mock contract, explicitly labeled until #2060 lands, consistent with the Library storage view which is already mock-derived on dev. Wired in correctly (view toggle + render chain), tests exercise the real component, a11y is right (label/htmlFor pairs, aria-labels on destructive buttons, radiogroup).

**Minor — fold into the #2060 real-backend card, none block merge:**
1. `handleAddSourceRule` allows duplicate/conflicting rules for the same source (youtube→skip and youtube→download can coexist); the winner is undefined. Dedup by source or last-write-wins when the contract goes real.
2. Storage cap input accepts 0 and negatives (`parseInt || 0`); cap 0 renders as perpetually paused-at-cap. Clamp to a sane minimum.
3. `newSourceRule` is typed by literal inference (`action: "download" as const`), forcing `as typeof prev.action` casts in the onChange handlers — a type-level lie that happens to be runtime-correct. Type it `(typeof ACTION_OPTIONS)[number]` instead.
4. `saveLibrarySettings` is called inside setState updaters (add/delete handlers) — updaters should be pure; Strict Mode double-invokes them (harmless here since the write is idempotent, but move the save to an effect or after the set when touching this next).
5. Product-polish note: the literal string "(mock until #2060)" is end-user-visible in the pane. Fine for beta; must not survive #2060.

The "(#2276)" changelog line rides CHANGELOG.md directly rather than a fragment — acceptable, gate satisfied either way.

PR #2275, opened 2026-08-03
taOStalk s1: ToolCallBlock + StatusBlock/question renderers (cards 5+6)

jaylfc:
CARD TITLE (intent, not commit subject): taOStalk s1: ToolCallBlock + StatusBlock/question renderers (cards 5+6)

Autonomous build of board card tsk-3lkyq2.

Files:
 docs/agent-manual/11-files-api.md                  | 35 --------
 docs/agent-manual/index.md                         |  1 -
 docs/design/taosgo-mesh-join-foundation.md         |  8 +-
 docs/taos-agent-manual.md                          | 35 --------
 tests/test_agent_manual_compiled.py                |  2 +-
 tinyagentos/taosnet/mesh.py                        |  3 +-
 tinyagentos/taosnet/mesh_credentials.py            |  3 +-
 14 files changed, 330 insertions(+), 102 deletions(-)

----
## Summary by Gitar

- **Components:**
    - Added `ToolCallBlock` component to render tool call content blocks with status indicators and previews
    - Added `StatusBlock` component to render status and question content blocks with reply hints
- **MessagesApp:**
    - Integrated `ToolCallBlock` and `StatusBlock` into `renderContentBlock` dispatch logic
    - Added `QuestionContentBlock` type definition and updated exports
- **Tests:**
    - Added test suites for `ToolCallBlock`, `StatusBlock`, and updated render helper tests

<sub>This will update automatically on new commits.</sub>

## Summary by CodeRabbit

* **New Features**
  * Added dedicated displays for tool activity, including tool names, inputs, statuses, results, and errors.
  * Added styled status and question messages with contextual hints and metadata.
  * Improved message rendering for tool calls, statuses, and questions.

* **Bug Fixes**
  * Unknown message types now retain a clear fallback display.

* **Tests**
  * Added coverage for tool-call states, status messages, questions, result previews, and fallback behavior.

PR #2282, opened 2026-08-04
taOStalk s1: TextBlock + ThinkingBlock renderers (cards 3+4)

jaylfc:
CARD TITLE (intent, not commit subject): taOStalk s1: TextBlock + ThinkingBlock renderers (cards 3+4)

Autonomous build of board card tsk-6ylbbx.

Files:
 desktop/src/apps/MessagesApp.tsx                   |  79 +++++++++++--
 .../apps/chat/__tests__/content-blocks.test.tsx    | 127 +++++++++++++++++++++
 .../apps/chat/__tests__/render-helpers.test.tsx    |  16 +--
 3 files changed, 207 insertions(+), 15 deletions(-)

## Summary by CodeRabbit

- **New Features**
  - Chat messages now render text content with inline Markdown formatting.
  - Thinking content appears in a collapsible panel that is closed by default and can be expanded when needed.
  - Added accessible expand/collapse controls with appropriate labels and relationships.

- **Bug Fixes**
  - Supported text and thinking content blocks no longer display as unsupported content.
  - Improved rendering consistency for chat content blocks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
 
 ### Added
 
+- Chat renders `text` and `thinking` content blocks: thinking is a
+  collapsed-by-default disclosure with a proper ARIA expand/collapse contract (#2282).
 - Messages sidebar shows a live "thinking" badge on channels whose bound
   taOStalk agent is currently working, on desktop and mobile (#2281).
 - Admin-only `POST /api/notifications` so orchestrators and lead agents can
--- a/desktop/src/apps/MessagesApp.tsx
+++ b/desktop/src/apps/MessagesApp.tsx
@@ -416,7 +416,7 @@ export function ThinkingBlock({ block, index }: { block: ThinkingContentBlock; i
         aria-controls={contentAria}
         aria-label={open ? "Collapse thinking" : "Expand thinking"}
         onClick={() => setOpen((o) => !o)}
-        className="flex w-full items-center gap-2 px-3 py-2 text-[12px] font-semibold text-shell-text-tertiary hover:text-shell-text-secondary hover:bg-white/[0.06] focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-accent/40"
+        className="flex w-full items-center gap-2 px-3 py-2 text-[12px] font-semibold text-shell-text-tertiary hover:text-shell-text-secondary hover:bg-shell-surface-hover focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-accent/40"
       >
         <ChevronDown
           size={14}
--- a/desktop/src/apps/MessagesApp.tsx
+++ b/desktop/src/apps/MessagesApp.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef, useCallback } from "react";
+import React, { useState, useEffect, useRef, useCallback, useId } from "react";
 import {
   MessageCircle,
   Hash,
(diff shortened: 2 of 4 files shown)


jaylfc:
**nemotron-super review** 

VERDICT: Test expectation for fallback lines is incorrect (should be 2, not 1)
- desktop/src/apps/chat/__tests__/render-helpers.test.tsx:98

jaylfc:
Reviewed at 0e1dc714; changelog added by me at 92101bf4. Good build - the strongest of today's lane batch.

VERIFIED:
- Replay check clean: one commit on the CURRENT dev tip, files match the card.
- Tests exercise REAL code: content-blocks.test.tsx imports { TextBlock, ThinkingBlock } from ../../MessagesApp and those are genuinely exported there, so this is not a suite that passes against nothing. 23/23 pass locally.
- The edit to the EXISTING render-helpers.test.tsx is a legitimate strengthening, not the net-weakening pattern I have been closing PRs for: the old assertions expected the 'unsupported block: text' placeholder, the new ones assert the real rendered content, and the still-unimplemented kinds (tool_call, status) keep their fallback assertions. The test name was updated to match what it now proves.
- ACCESSIBILITY DONE RIGHT, unprompted: ThinkingBlock uses a real <button type=button> (so keyboard works natively), aria-expanded, aria-controls, an aria-label that changes with state, and the decorative chevron is aria-hidden. That is the textbook disclosure contract - noting it because I had to fix exactly this on #2281.

RED, neither of them yours:
- doc-gate was the missing CHANGELOG line, now added.
- dependency-audit fails on 3 CVEs in cryptography 48.0.1, which litellm[proxy] pins <49.0 so it cannot be upgraded. It fails on every PR right now and is NOT caused by this one. Tracked as tsk-4kifi6 with decision dec-26leeh for Jay; it is not a required check on dev, so it does not block this merge.

Auto-merge armed on green.

jaylfc:
Rebased onto dev at d5984f17 after #2275 landed. This was NOT a changelog-only conflict - it was a semantic one, which is why spa-build went red.

WHAT COLLIDED: #2275 and this PR both rewrote the same renderContentBlock switch from opposite ends. #2275's version renders tool_call/status/question through their card components and lets text/thinking fall through; this PR's renders text/thinking and lets tool_call/status fall through. Git could not reconcile them and, more importantly, KEEPING BOTH NAIVELY WOULD HAVE BEEN WRONG in the test file: #2275 asserts text/thinking produce 'unsupported block', this PR asserts tool_call/status do, and after the union NONE of the four do. Two green-looking test suites that contradict each other.

RESOLVED as the union, which is what both cards together were always meant to produce: all four slice-1 kinds (text, thinking, tool_call, status/question) now dispatch to their dedicated components, and only genuinely unrecognised kinds hit the fallback. The contradictory assertions were replaced with one test that renders all four and asserts NO 'unsupported block' appears, plus a retained single-unknown-kind test proving the fallback still fires exactly once.

VERIFIED, since the failing check was a build rather than a test: 38/38 across all four affected suites, and 'npm run build' completes (built in 22.7s) - so the TypeScript that broke spa-build compiles now. I ran the build itself rather than trusting the unit tests, because the tests passed even while spa-build was failing.

Remaining red is dependency-audit only - the fleet-wide cryptography CVE issue (tsk-4kifi6 / dec-26leeh), not required on dev and not caused by this PR. Auto-merge stays armed.

jaylfc:
Fixed the real spa-build failure at bd89688e, and correcting my own earlier claim on this PR.

I SAID 'the TypeScript that broke spa-build compiles now' after running npm run build plus four targeted test files. That verified the wrong thing: spa-build does not run a build at all, it runs the FULL vitest suite (npx vitest run). My green proved nothing about the check that was failing, which is the exact mistake I have been calling out elsewhere today.

THE ACTUAL DEFECT, unrelated to the semantic conflict: MessagesApp.palette.test.ts forbids hardcoded white/black Tailwind utilities in this file, and the new thinking-disclosure button used hover:bg-white/[0.06]. That is why spa-build failed both times - 380 files passed, 1 failed, every run.

FIXED with the token the file already uses elsewhere (hover:bg-shell-surface-hover) rather than the palette-ok comment the test also honours. The waiver would have silenced the check instead of satisfying it, and the pre-existing waived line (the canvas iframe, legitimately white) is the kind of case that escape hatch is actually for.

VERIFIED PROPERLY THIS TIME: the entire suite, as CI runs it - 381 files, 3279 tests, all passing. Remaining red is dependency-audit only (fleet-wide cryptography CVEs, tsk-4kifi6 / dec-26leeh, not required on dev) and Kilo's known rate-limit fake-red. Auto-merge stays armed.

PR #2284, opened 2026-08-04
Settings: Notifications pane with per-type toggles (disable e.g. task.claimed)

jaylfc:
CARD TITLE (intent, not commit subject): Settings: Notifications pane with per-type toggles (disable e.g. task.claimed)

Autonomous build of board card tsk-t52zqq.

Files:
 desktop/src/apps/SettingsApp.tsx    | 112 ++++++++++++++++++++++++++++++++++--
 tests/test_notifications.py         |  39 ++++++++++++-
 tests/test_notifications_prefs.py   |   7 ++-
 tests/test_routes_notifications.py  |  14 ++---
 tests/test_store_upgrades.py        |  53 +++++++++++++----
 tinyagentos/notifications.py        |  89 +++++++++++++++++-----------
 tinyagentos/routes/notifications.py |   8 +--
 7 files changed, 258 insertions(+), 64 deletions(-)

## Summary by CodeRabbit

- **New Features**
  - Added a Notifications section to Settings with controls for enabling or muting event types.
  - Notification preferences are now personalized per account.
  - `task.claimed` notifications are muted by default.
  - Broadcast notifications continue to be delivered regardless of personal preferences.

- **Bug Fixes**
  - Existing notification preferences are preserved during upgrades.
  - Improved handling of missing system information and advanced-settings text.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
 
 ### Added
 
+- Settings gains a Notifications pane with per-type toggles, stored per user;
+  preferences set before the per-user migration keep applying until that user
+  overrides them (#2284).
 - Chat renders `text` and `thinking` content blocks: thinking is a
   collapsed-by-default disclosure with a proper ARIA expand/collapse contract (#2282).
 - Messages sidebar shows a live "thinking" badge on channels whose bound
--- a/desktop/src/apps/SettingsApp.tsx
+++ b/desktop/src/apps/SettingsApp.tsx
@@ -17,6 +17,7 @@ import {
   UserCircle,
   ScrollText,
   Save,
+  Bell,
 } from "lucide-react";
 import {
   Button,
@@ -40,7 +41,7 @@ import { LogsSection } from "@/apps/SettingsApp/LogsPanel";
 /*  Types                                                              */
 /* ------------------------------------------------------------------ */
 
-type Section = "account" | "system" | "storage" | "memory" | "backup" | "updates" | "advanced" | "shortcuts" | "accessibility" | "desktop" | "users" | "themes" | "logs";
+type Section = "account" | "system" | "storage" | "memory" | "backup" | "updates" | "advanced" | "shortcuts" | "accessibility" | "desktop" | "users" | "themes" | "logs" | "notifications";
 
 interface SectionDef {
   id: Section;
@@ -82,6 +83,7 @@ const SECTIONS: SectionDef[] = [
   { id: "users", label: "Users", icon: Users },
   { id: "themes", label: "Themes", icon: Palette },
   { id: "logs", label: "Logs", icon: ScrollText },
+  { id: "notifications", label: "Notifications", icon: Bell },
 ];
 
 // Sections that require admin privileges to view (GHSA-47g9: backend already
@@ -148,7 +150,7 @@ function SystemInfoSection() {
         hw.npu?.tops && hw.npu.tops > 0 ? ` · ${hw.npu.tops} TOPS` : "";
       const diskType = hw.disk?.type ? ` ${hw.disk.type}` : "";
       const osParts = [hw.os?.distro, hw.os?.version].filter(Boolean);
-      const osStr = osParts.length > 0 ? osParts.join(" ") : "—";
+      const osStr = osParts.length > 0 ? osParts.join(" ") : "--";
       setInfo({
         cpu: `${cpuModel}${cpuCores}${cpuArch}`,
         ram:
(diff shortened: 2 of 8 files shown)


jaylfc:
Reviewed at d736205a; lead-completed at 24d729aa. Real upgrade-regression found - worth reading before this pattern repeats.

THE DEFECT: the migration rebuilds notification_prefs from global (event_type PK) to per-user ((user_id, event_type) PK) and copies existing rows under user_id='default'. But every read path - _is_event_muted and get_event_prefs - queries strictly WHERE user_id = <caller's id>, and nothing in the system ever has the id 'default': real users get their record id, and the legacy password-file path returns a dict with NO 'id' key at all, so those callers use "". The migrated rows are therefore read by nobody. Effect on a real upgrade: every notification type a user had muted starts arriving again, silently - no error, no log, the row still sitting in the table.

WHY CI WOULD NEVER HAVE CAUGHT IT: test_upgrade_preserves_notification_prefs_data asserts the ROW EXISTS with user_id == 'default'. It tests the migration's mechanics and passes cleanly through the regression. That is a decorative test in the precise sense - green while the thing it is named for is broken.

FIXED: both read paths now fall back to the sentinel row (named MIGRATED_GLOBAL_USER rather than a bare 'default' literal) when the caller has no row of their own, with the user's OWN row winning when present. Added test_upgrade_keeps_the_pre_upgrade_mute_IN_EFFECT which asserts the MUTE APPLIES for a real user id, that a user's own later choice overrides it, and that the override does not leak to a second user. RED-PROVEN: it fails against your original notifications.py and passes with the fix.

The migration itself is otherwise well built - PRAGMA-detected, transactional with rollback, idempotent on the user_id column. Only the read side was missing.

58 tests pass across all four notification suites. Changelog added. Auto-merge armed on green.

jaylfc:
**nemotron-super review** 

VERDICT: One blocking issue found.
- desktop/src/apps/SettingsApp.tsx:933 - The NotificationsSection component attempts to map over null when the fetch for notification prefs fails, causing a runtime error.

jaylfc:
Closing to rebuild. The pane is well made, but as written EVERY TOGGLE IN IT IS INERT, and it also regresses the global mute that works on dev today. This needs a design decision, not a patch, which is why I am bouncing rather than fixing it for you.

THE CORE PROBLEM. This PR makes prefs per-user and gates the check on having a user:
  _is_event_muted(event_type, user_id): if user_id is None: return False
  emit_event(..., user_id: str | None = None) -> _is_event_muted(event_type, user_id)
Then I grepped every emit_event call site in tinyagentos/: task lifecycle (routes/projects.py:899), health.py, disk_quota.py, auto_update.py, knowledge_ingest.py. ZERO of them pass user_id. So every system-emitted event arrives with user_id=None, the mute check bails out immediately, and no preference is ever consulted.

Concretely: the example in this PR's own title - disable task.claimed - does not work. routes/projects.py:899 calls emit_event("task.claimed", ...) with no user, so muting it in the new pane changes nothing.

AND IT IS A REGRESSION. On dev, _is_event_muted(event_type) does a global lookup with no user gate, so muting task.claimed works TODAY. After this PR it stops working, for every event type, for everyone.

YOUR OWN TEST CAUGHT IT and was not run: tests/test_task_lifecycle_notifications.py::test_mute_suppresses_claim_notification fails on the new signature (TypeError, missing user_id). That test is the feature's acceptance criterion. It was left un-updated, along with a production caller - routes/settings.py:484/494 still called get_event_prefs()/set_event_muted() unqualified, which is what turned test (3.12), test (3.13) and shards red.

THE DESIGN CALL for the rebuild - pick one deliberately:
(a) Fan out per user: emit_event resolves the recipients and checks each one's prefs. Most correct for multi-user, most work.
(b) Keep a GLOBAL row as the fallback for user-less events: when user_id is None, consult a sentinel row rather than returning False, and have the pane write that row as well as the user's. Restores today's behaviour and adds per-user on top. Cheapest coherent option, and my recommendation for now given taOS instances are overwhelmingly single-user.
(c) Pass the owning user at every emit site. Correct but touches five subsystems.

WHAT I ALREADY BUILT, keep it: the migration in this branch parks old global prefs under user_id='default' but nothing ever read them, so pre-upgrade mutes silently stopped applying - the row survived, unread. I added a sentinel fallback (MIGRATED_GLOBAL_USER) in both read paths, fixed the two settings.py callers, and added test_upgrade_keeps_the_pre_upgrade_mute_IN_EFFECT, which is RED-PROVEN against the original. Note your existing upgrade test asserted only that the ROW EXISTS - it stayed green through that regression. All of that is on the branch at 24d729aa and option (b) builds directly on it.

Guidance is on the card tsk-t52zqq.

PR #2288, opened 2026-08-04
fix(theme-store): use injected get() to break circular self-reference

hognek:
Task: t_71a7eb3a

Replace the `useThemeStore.getState()` call inside the theme-store create closure with the `get()` argument injected by zustand's `create`. This eliminates the TS7022 / TS7006 cascade caused by referencing the store within its own initializer.

Changed:
- `create<ThemeStore>((set) => ({...}))` -> `create<ThemeStore>((set, get) => ({...}))`
- `getWallpapersBySection` now uses `get()` instead of `useThemeStore.getState()`

Only the in-closure occurrence changes. Module-level `getState()` calls in exported helpers (outside the creator) keep referencing the hook directly — they are not circular.

Verification:
- `tsc -b --force` exits 0 (full uncached typecheck, no TS7022/TS7006)
- `vitest run src/stores/theme-store.test.ts src/stores/__tests__/wallpaper.test.ts`: 19/19 pass

## Summary by CodeRabbit

* **Refactor**
  * Improved theme state handling when generating wallpaper sections.
  * No visible changes to the user experience.


jaylfc:
Reviewed. ON-SPEC, arming. Correct fix and the right idiom.

getWallpapersBySection called useThemeStore.getState() from INSIDE the store's own initializer, which is a circular self-reference: the module binding is not initialised while create() is still running, so the call depends on module-evaluation order rather than on anything the code controls. Swapping to zustand's injected get() is the standard fix and is semantically identical for callers - same state object, no behaviour change - while removing the cycle.

Verified rather than eyeballed: 191 tests across all 16 store suites pass with the change, including restore-theme.test.ts which is the one that would notice if getWallpapersBySection started returning a different shape.

No changelog needed (internal correctness fix, no user-visible behaviour change) and doc-gate agrees - it is not red. CI is otherwise clean on this one.

Auto-merge armed on green.

jaylfc:
recreate-cla

hognek:
Pushed empty commit to retrigger CLA check.

hognek:

jaylfc:
recreate-cla

hognek:
recreate-cla

jaylfc:
recreate-cla

jaylfc:
Refreshing CI against the current base (dev moved 14 times since these checks ran). The cla check cannot structurally appear on this PR (fork PRs do not fire pull_request_target on synchronize, and the comment-triggered run binds to the default branch); hognek's CLA signature is on record in cla-signatures, so this will merge on fresh green with a documented gate exception.

jaylfc:
Merging with the cla check absent, documented exception: fork PRs cannot fire pull_request_target on synchronize (measured earlier today), and the comment-triggered run binds to the default branch, so cla is structurally unable to appear here. hognek's CLA signature is verified in the cla-signatures branch (signatures/version1/cla.json), which satisfies the check's actual purpose. All other required checks are green on the current head after the close/reopen refresh; base drift since those runs is one docs-only merge (#2327).

PR #2289, opened 2026-08-04
fix(desktop): CSRF protection + Headers spread fix in x-monitor and fetch wrappers

hognek:
Task: t_4a8783ff (successor to closed #2165).

Rebases fix/2126-library-csrf onto origin/dev tip (ad7e4fea) and completes the Headers spread fix:
- x-monitor.ts fetchJson no longer spreads the withCsrf() Headers instance into an object literal (which silently dropped every header). It now builds a fresh Headers via new Headers(init?.headers) + forEach, so the CSRF token actually reaches the wire.
- Same normalization applied to memory/knowledge/github fetch wrappers.
- Updated Content-Type test assertions to the dual-mode pattern (headers?.get?.() ?? headers?.[...]) so tests pass for both Headers-instance and plain-object callers.

Tests: desktop vitest 3205/3205 pass.

### DIFF
--- a/desktop/src/__tests__/entry-guards.test.ts
+++ b/desktop/src/__tests__/entry-guards.test.ts
@@ -1,0 +1,70 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+// Use vi.hoisted so the mock reference is stable across vi.resetModules() calls.
+// vi.mock factories only run once; vi.resetModules clears the module cache but
+// does not re-invoke the factory, so the returned vi.fn() instance persists.
+// mockClear() between tests resets the call count so each assertion sees only
+// the calls from the current test's dynamic import().
+const { installAuthGuard: mockGuard } = vi.hoisted(() => ({
+  installAuthGuard: vi.fn(),
+}));
+
+vi.mock("../lib/auth-guard", () => ({
+  installAuthGuard: mockGuard,
+  SESSION_EXPIRED_EVENT: "taos-session-expired",
+}));
+
+// Entry modules call createRoot(document.getElementById("root")!).render(...).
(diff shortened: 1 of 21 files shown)


jaylfc:
Reviewed in depth. CORRECT FIX for a genuinely subtle bug - arming. One missed site, carded rather than blocking you on it.

WHY THIS MATTERS, since it reads as cosmetic: 'headers: { Accept: ..., ...init?.headers }' looks like a normal merge, but when a caller passes a Headers INSTANCE the spread yields {} - Headers keeps its entries internally and exposes no enumerable own properties - so every caller-supplied header is silently dropped, with nothing failing. 'new Headers(init?.headers)' then .set() is right because the constructor ingests all three legal forms (Headers, plain object, entry array). This is the call-site half of the CSRF work; csrf.ts itself was always fine.

VERIFIED: full desktop suite on your branch, 377 files / 3205 tests, all passing. Only red is dependency-audit, the fleet-wide cryptography CVE issue (tsk-4kifi6 / dec-26leeh), not yours and not required on dev.

MISSED SITE - same class, one file you did not touch: desktop/src/lib/projects.ts:137-145 still does 'headers: { "Content-Type": "application/json", ...(init?.headers || {}) }' with init typed RequestInit, so a Headers instance drops the same way. It is actually worse there: '...init' appears AFTER the headers key, so ANY caller passing init.headers overwrites the merged object entirely and loses Content-Type: application/json. Carded as a follow-up so it cannot be forgotten; if you would rather fold it into this PR before it merges, that is better still - same sweep, three lines.

Not blocking on it: the fix in front of me is correct and the tests back it. Auto-merge armed on green.

PR #2299, opened 2026-08-04
desktop/src/lib/projects.ts: same Headers-spread bug + init overwrites merged headers

jaylfc:
CARD TITLE (intent, not commit subject): desktop/src/lib/projects.ts: same Headers-spread bug + init overwrites merged headers

Autonomous build of board card tsk-ux4a7q.

projectsApi.http built its default headers with a plain-object spread
(...(init?.headers || {})). When init.headers was a Headers instance
(as returned by withCsrf downstream), the spread silently dropped every
entry, and the trailing ...init spread overwrote the merged headers key
entirely, clobbering Content-Type. Mirror the x-monitor/library fix:
build a fresh Headers via new Headers(init?.headers), set Content-Type
on it, and spread ...init before headers so merged headers win.

Files:
 desktop/src/lib/projects.test.ts | 47 ++++++++++++++++++++++++++++++++++++++++
 desktop/src/lib/projects.ts      |  4 +++-
 2 files changed, 50 insertions(+), 1 deletion(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP request handling to consistently preserve custom headers and JSON content-type settings.
  * Ensured CSRF protection is correctly applied to mutating requests without adding unnecessary headers to read-only requests.

* **Tests**
  * Added regression coverage for CSRF behavior and header preservation.

### DIFF
--- a/desktop/src/lib/projects.test.ts
+++ b/desktop/src/lib/projects.test.ts
@@ -609,3 +609,50 @@ describe("projectsApi.activity", () => {
     await expect(projectsApi.activity("p-1")).rejects.toThrow("500");
   });
 });
+
+describe("projectsApi.http CSRF + header merge (Headers-spread regression)", () => {
+  const originalDocument = global.document;
+
+  afterEach(() => {
+    vi.stubGlobal("document", originalDocument);
+  });
+
+  const headerValue = (opts: unknown, name: string): string | null => {
+    const headers = (opts as { headers?: Headers | Record<string, string> }).headers;
+    if (headers instanceof Headers) return headers.get(name);
+    return headers ? (headers as Record<string, string>)[name] ?? null : null;
+  };
+
+  it("attaches X-CSRF-Token and preserves Content-Type on POST", async () => {
+    vi.stubGlobal("document", { cookie: "csrf_token=hdr-token" } as unknown as Document);
+    const fetchMock = vi.fn().mockResolvedValue({
+      ok: true,
+      status: 200,
+      json: async () => ({ id: "p-1", name: "New", slug: "new", description: "", status: "active", created_by: "u-1", created_at: 1, updated_at: 1 }),
+      text: async () => "",
+      statusText: "OK",
+    });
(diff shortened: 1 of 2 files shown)


jaylfc:
Reviewed. The FIX IS CORRECT and I am arming it - but the tests named for the regression do not actually catch the regression, and you should know why before trusting them.

THE FIX IS RIGHT, including the subtlety I care about most: '...init' now comes BEFORE 'headers', so a caller-supplied init.headers can no longer clobber the merged object. That ordering was half the bug and it is correct here. new Headers(init?.headers) also handles all three legal forms. Setting Content-Type unconditionally rather than only-if-absent is fine in this file specifically - I checked, no caller passes its own Content-Type and there is no FormData path - though it does mean this wrapper is JSON-only by construction now.

THE TESTS DO NOT PROVE IT. I reverted projects.ts to the ORIGINAL buggy form and re-ran your suite: 73/73 still passed. The block is named 'Headers-spread regression' but no test passes caller headers, so nothing exercises a spread of a Headers instance or the init-ordering clobber. It asserts Content-Type and X-CSRF-Token arrive, which they do in BOTH versions.

WHY THAT IS NOT REALLY YOUR FAULT, and why I am not asking for a rewrite: http() is not exported and NO public method passes init.headers - I grepped, zero call sites. So the bug is LATENT: unreachable through today's API, a trap for whoever next adds a method that threads headers. You cannot write a red-first test through the current public surface without exporting http() purely for testing, which is its own smell.

SO, honestly stated: this is a correct DEFENSIVE fix with no regression coverage, and that is acceptable here because the defect is not reachable yet. If http() ever gains a caller that passes headers, that PR owes the test - a Headers instance in, both the caller's header and Content-Type out.

73/73 pass, CI clean. Auto-merge armed.

jaylfc:
**nemotron-super review** 

VERDICT: Missing test case for init.headers scenario in projectsApi.http CSRF + header merge test suite
- desktop/src/lib/projects.test.ts:612

PR #2303, opened 2026-08-04
Edge: reduce-effects / performance-mode toggle in Settings (#361)

jaylfc:
CARD TITLE (intent, not commit subject): Edge: reduce-effects / performance-mode toggle in Settings (#361)

Autonomous build of board card tsk-kur32h.

Files:
 desktop/src/App.tsx                                |  2 +-
 desktop/src/components/Desktop.tsx                 |  3 ++-
 .../src/stores/__tests__/reduce-effects.test.ts    | 23 ++++++++++++++++++++++
 3 files changed, 26 insertions(+), 2 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Reduced-effects mode now suppresses animated particle wallpapers for a calmer, less resource-intensive experience.

* **Bug Fixes**
  * Reduced-effects preferences now apply consistently and persist correctly when enabled or disabled.
  * The reduced-performance setting is reflected accurately across the interface.

PR #2304, opened 2026-08-04
doc-gate: user-visible-changelog misses desktop shell changes (App.tsx, components/, stores/)

jaylfc:
CARD TITLE (intent, not commit subject): doc-gate: user-visible-changelog misses desktop shell changes (App.tsx, components/, stores/)

Autonomous build of board card tsk-wnqh3d.

Add desktop/src/App.tsx, desktop/src/components/**, and desktop/src/stores/**
to the when_changed glob for the user-visible-changelog rule so modifications
to the desktop shell require a changelog entry, matching the existing coverage
for desktop/src/apps/*/**.

Files:
 docs/doc-gate.toml     |  2 +-
 tests/test_doc_gate.py | 41 ++++++++++++++++++++++++++++++++++++++++-
 2 files changed, 41 insertions(+), 2 deletions(-)

## Summary by CodeRabbit

* **Documentation**
  * Improved release documentation checks to cover changes across the desktop app, components, stores, and installer files.
  * Added validation to ensure applicable desktop updates include user-visible changelog entries.

### DIFF
--- a/docs/doc-gate.toml
+++ b/docs/doc-gate.toml
@@ -66,7 +66,7 @@ hint = "the agent-token route allowlist changed; update the agent-facing API sur
 [[rules]]
 name = "user-visible-changelog"
 on_modify = true
-when_changed = ["tinyagentos/routes/*.py", "desktop/src/apps/*/**", "tinyagentos/installers/*"]
+when_changed = ["tinyagentos/routes/*.py", "desktop/src/apps/*/**", "desktop/src/App.tsx", "desktop/src/components/**", "desktop/src/stores/**", "tinyagentos/installers/*"]
 require_doc = ["CHANGELOG.md", "changelog.d/*.md"]
 hint = "user-visible behaviour changed; add a changelog.d/<pr>-<slug>.md fragment (preferred) or a CHANGELOG.md line (or add a Docs-Reviewed trailer explaining why not)"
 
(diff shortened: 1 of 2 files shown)


jaylfc:
Reviewed. ON-SPEC, arming - and it respected the caution in the card rather than pasting the widening I suggested, which I appreciate.

WHAT IT DOES: adds desktop/src/App.tsx, desktop/src/components/** and desktop/src/stores/** to the user-visible-changelog rule, so a shell change like #2303 (which stopped an animated wallpaper animating and merged with NO release-note trail, because nothing asked) now demands a changelog or a fragment.

NARROWER THAN MY SUGGESTION, correctly: I listed desktop/src/registry/** too and this leaves it out. Registry edits are mostly app-manifest plumbing rather than behaviour a user sees, so excluding it keeps the gate quieter. Narrower is the right instinct on a rule whose own config header warns that a noisy gate gets disabled.

PROVEN AGAINST THE REAL GATE, both directions, not just the unit fixtures:
- appended a line to desktop/src/App.tsx with no changelog -> DOC-GATE FAIL, exit 1. The rule genuinely fires now.
- appended a line to a shell __tests__ file -> doc-gate clean, exit 0. Test-only edits do NOT trigger it, which was the noise case I was most worried about. That works because check_doc_gate.py already treats test paths as never-structural regardless of on_modify - worth knowing, it means the widening could not have made tests noisy even if the PR had not thought about it.
- 32 unit tests pass, including the three new positive cases and the test-only negative.

Auto-merge armed on green.

jaylfc:
**nemotron-super review** 

VERDICT: Blocking issue found: test_desktop_shell_test_file_does_not_trigger incorrectly expects no trigger for test files
- tests/test_doc_gate.py:289   (test expects no trigger but pattern matches)

PR #2306, opened 2026-08-05
check_doc_gate.py: a broken/unparseable config exits 1 identically to a real violation

jaylfc:
CARD TITLE (intent, not commit subject): check_doc_gate.py: a broken/unparseable config exits 1 identically to a real violation

Autonomous build of board card tsk-thh54d.

An unparseable or missing doc-gate config previously raised an unhandled
traceback that exited 1 -- identical to a real documentation-drift violation,
so a typo in docs/doc-gate.toml was indistinguishable from a missing
changelog. Catch tomllib.TOMLDecodeError and OSError in main(), print a
clear error to stderr, and exit 2 instead.

Files:
 scripts/check_doc_gate.py | 17 ++++++++++++++---
 tests/test_doc_gate.py    | 31 +++++++++++++++++++++++++++++++
 2 files changed, 45 insertions(+), 3 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration files that are missing or invalid now produce a distinct error status and a clear error message.
  * Genuine documentation-rule violations continue to return the appropriate violation status.

* **Tests**
  * Added coverage to verify separate statuses for configuration errors and documentation violations.

### DIFF
--- a/scripts/check_doc_gate.py
+++ b/scripts/check_doc_gate.py
@@ -37,6 +37,13 @@ REPO_ROOT = Path(__file__).resolve().parent.parent
 DEFAULT_CONFIG = REPO_ROOT / "docs" / "doc-gate.toml"
 DEFAULT_TRAILER = "Docs-Reviewed:"
 
+# Exit codes: 0 clean, 1 a doc-gate violation, 2 a config/usage error
+# (broken, missing, or unparseable config). Distinct so a misconfigured gate
+# is never mistaken for a real documentation-drift violation.
+EXIT_OK = 0
+EXIT_VIOLATION = 1
+EXIT_CONFIG_ERROR = 2
+
 # A path-like token: one of the four known repo prefixes followed by a run of
 # non-whitespace / non-quoting characters. The negative lookbehind stops us
 # matching a prefix that is actually embedded inside a larger path (e.g. the
@@ -254,10 +261,10 @@ def get_trailer(config: dict) -> str:
 def _report(failures: list[str]) -> int:
     if not failures:
         print("doc-gate: clean")
-        return 0
+        return EXIT_OK
     for failure in failures:
         print(f"DOC-GATE FAIL: {failure}")
-    return 1
+    return EXIT_VIOLATION
 
 
 def main(argv: list[str] | None = None) -> int:
@@ -273,7 +280,11 @@ def main(argv: list[str] | None = None) -> int:
     group.add_argument("--base", help="Compare <base>...HEAD (CI / commit-msg)")
 
     args = parser.parse_args(argv)
-    config = load_config(args.config)
+    try:
+        config = load_config(args.config)
+    except (tomllib.TOMLDecodeError, OSError) as e:
+        print(f"doc-gate: config error: {args.config}: {e}", file=sys.stderr)
+        return EXIT_CONFIG_ERROR
 
     if args.command == "invariants":
         files_to_scan = config.get("invariants", {}).get("referenced_paths_scan", [])
--- a/tests/test_doc_gate.py
+++ b/tests/test_doc_gate.py
@@ -295,3 +295,34 @@ class TestChangelogFragments:
             ("A", "desktop/src/stores/__tests__/theme-store.test.ts"),
         ]
         assert dg.evaluate_rules(changed, [], CHANGELOG_RULE_CONFIG) == []
+
+
+class TestConfigErrorExitCode:
+    """A broken/unparseable config must exit distinctly from a real violation.
+
+    Regression: previously an unparseable config raised an unhandled traceback
+    that exited 1 -- identical to a genuine doc-gate violation -- so a typo in
+    docs/doc-gate.toml looked just like a missing changelog."""
+
+    def test_unparseable_config_returns_config_error_code(self, tmp_path, capsys):
+        bad = tmp_path / "bad.toml"
+        bad.write_text('key = "unterminated string\n')
+        rc = dg.main(["--config", str(bad), "print-trailer"])
+        captured = capsys.readouterr()
+        assert rc == dg.EXIT_CONFIG_ERROR
+        assert rc != dg.EXIT_VIOLATION
+        assert "config error" in captured.err
+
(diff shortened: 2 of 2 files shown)


jaylfc:
**nemotron-super review** 

VERDICT: Pass
No blocking issues found.

jaylfc:
Reviewed against `tsk-thh54d`'s red-first list. The core works and I proved it both ways rather than reading it. One defect blocks merge, and it is the same class the card exists to eliminate.

**The feature is real, verified red-then-green.** On this branch:

```
--config /tmp/nope.toml diff-gate --staged
  -> doc-gate: config error: /tmp/nope.toml: [Errno 2] No such file or directory   exit=2
--config /tmp/bad.toml  diff-gate --staged
  -> doc-gate: config error: /tmp/bad.toml: Expected '=' after a key in a key/value pair (at line 1, column 6)   exit=2
genuine violation -> exit=1        clean -> exit=0
```

On dev, both config cases produce a raw `FileNotFoundError` / `tomllib.TOMLDecodeError` traceback and **exit 1**, indistinguishable from a rule violation. So the card's premise is confirmed and the fix addresses it. Messages name the path and, for the parse error, the line and column. `load_config` opens and parses directly, so `OSError` and `TOMLDecodeError` genuinely propagate and the `except` clause is live rather than inert. Exit 9 is correctly avoided.

---

### BLOCKING — `EXIT_CONFIG_ERROR = 2` collides with argparse's own exit code

argparse exits **2** on any usage error. So after this change:

```
check_doc_gate.py diff-gate --nonsense-flag
  -> check_doc_gate.py diff-gate: error: one of the arguments --staged --base is required
  -> exit=2          # identical to a broken config
```

"The config is missing or unparseable" and "you invoked the command wrong" now return the same code. The card's stated goal is that these stop looking alike; this collapses two of them again, one step to the left.

This is not hypothetical and it nearly fooled me. My first test run put `--config` after the subcommand, argparse rejected it as an unrecognised argument, and I got `exit=2` on all three cases — which looked exactly like the feature working. It was argparse, and the config path never executed. I only caught it because the printed message was a usage block rather than the `doc-gate: config error:` line. A caller branching on the exit code has no such tell.

That matters most in CI, which is where this code is designed to be read by a machine: a workflow that grows a typo in its `check_doc_gate.py` invocation will report "config error" and send whoever is on call to inspect a `doc-gate.toml` that is perfectly fine.

**Fix:** use an exit code argparse does not claim. `EXIT_CONFIG_ERROR = 3` is the obvious choice (2 is argparse, 1 is a violation, 9 belongs to `gate_merge.sh` for stale-check aborts). Please also add a test asserting a malformed command line and a broken config return *different* codes — that is the property the card is actually buying, and nothing currently pins it.

### Smaller

The exit-code constants are a genuine improvement and worth carrying into `gate_merge.sh`'s expectations, since that script now has to distinguish 1 / 3 / 9. Worth confirming in this PR whether anything already branches on `check_doc_gate.py`'s exit code, so the new code does not surprise an existing caller.

---

Once the code moves off 2 and the "different codes" test lands, this is good to merge. Related work, not for this branch: I have carded three separate rule-engine defects found in the same file (`tsk-vs6c77` deletion satisfies a rule, `tsk-ymv3bb` rename bypass, `tsk-xe2tt5` git failures indistinguishable from violations). `tsk-xe2tt5` is deliberately blocked on this PR so it can reuse whatever code lands here — please leave it alone in this branch so this one stays reviewable against its own list.

jaylfc:
Correcting my own verdict above. I closed that review with "once the code moves off 2 and the 'different codes' test lands, this is good to merge." **That was premature.** Qodo and CodeRabbit each filed a finding I had missed, both are correct, and I verified both on this branch rather than taking their word.

Between them they mean the config-error path still has two holes of exactly the class this card exists to close.

**1. A structurally invalid config still exits 1** (qodo). Valid TOML, wrong shape:

```
# /tmp/struct.toml
rules = "not a list"
[gate]
trailer = "X:"
```
```
--config /tmp/struct.toml diff-gate --staged
  AttributeError: 'str' object has no attribute 'get'
  exit=1
```

`load_config` succeeds because the TOML parses; the failure happens later, in the rule loop. So a broken config is once again indistinguishable from a documentation violation, which is the precise thing the card asks to eliminate.

**2. Invalid UTF-8 in the config file still exits 1** (CodeRabbit). `tomllib.load` decodes before parsing, so a bad byte raises `UnicodeDecodeError`:

```
--config /tmp/badbytes.toml diff-gate --staged
  UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0
  exit=1
```

And the except clause cannot catch it:

```
issubclass(UnicodeDecodeError, OSError)                -> False
issubclass(UnicodeDecodeError, tomllib.TOMLDecodeError) -> False
```

`UnicodeDecodeError` is a `ValueError`. A truncated download, a file saved in the wrong encoding, or a mangled merge all land here.

**Revised asks**, replacing the single one in my earlier comment:

1. Move off exit code 2 (argparse owns it) — as before, exit 3.
2. Add `UnicodeDecodeError` (or `ValueError`) to the caught set around `load_config`.
3. Validate the config's *shape* after parsing, not just that it parsed: `rules` is a list of tables, `gate` is a table, each rule's `when_changed` / `require_doc` are lists. Report shape failures through the same config-error exit code, naming the offending key.
4. Tests for all three, each proven red on the current commit first: structurally invalid config, non-UTF-8 config, and malformed command line vs broken config returning *different* codes.

Point 3 is the one worth doing properly rather than minimally. "The config file loaded" and "the config file is usable" are different claims, and this PR currently only checks the first — which is the same absent-versus-negative distinction that keeps producing these bugs.

Credit to both bots; my review would have let this through with two of the three holes still open. Noting for the record that CodeRabbit's finding here came with an executed script demonstrating the decode failure, not just an assertion, which is why it was quick to confirm.

jaylfc:
Closing as **superseded by #2312** (its fix-forward), which merged with the corrected config-error exit paths plus the review fixes, STEP-0 verified (no commits lost). The doc-gate work continues on dev.

PR #2307, opened 2026-08-05
check_doc_gate invariants: file:line citations and cross-repo paths fail as nonexistent

jaylfc:
CARD TITLE (intent, not commit subject): check_doc_gate invariants: file:line citations and cross-repo paths fail as nonexistent

Autonomous build of board card tsk-zcnklp.

Files:
 scripts/check_doc_gate.py |  3 ++-
 tests/test_doc_gate.py    | 24 ++++++++++++++++++++++++
 2 files changed, 26 insertions(+), 1 deletion(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of file references with source-line citations, punctuation, and line ranges.
  * Preserved valid filenames containing colons.
  * Prevented path-like text in external Markdown URLs from being incorrectly flagged.

* **Tests**
  * Added coverage for file-and-line citations, punctuation variants, colon-containing filenames, and external URLs with path-like content.

### DIFF
--- a/scripts/check_doc_gate.py
+++ b/scripts/check_doc_gate.py
@@ -65,6 +65,12 @@ _GLOB_OR_PLACEHOLDER_CHARS = set("*?[]{}<>$~")
 # Trailing punctuation that is prose/markdown decoration, not part of the path.
 _TRAILING_PUNCT = ".,;:!?)]}'\"`"
 
+# A line-citation suffix: `:123`, `:12-20`, `:1,5-9`, and repeated groups so
+# file:line:col (ripgrep, compiler output) collapses in one pass. Applied
+# AFTER the trailing-punct strip — the anchor is defeated by a trailing full
+# stop or comma otherwise, and citations ending a sentence are the common case.
+_LINE_SUFFIX_RE = re.compile(r"(?::[0-9]+(?:-[0-9]+)?(?:,[0-9]+(?:-[0-9]+)?)*)+$")
+
 
 def _clean_token(raw: str) -> str | None:
     """Normalize a raw regex match into a bare repo-relative path, or None
@@ -74,9 +80,9 @@ def _clean_token(raw: str) -> str | None:
     for sep in ("#", "?", "::"):
         if sep in token:
             token = token.split(sep, 1)[0]
-    token = re.sub(r":[0-9]+(?:-[0-9]+)?(?:,[0-9]+(?:-[0-9]+)?)*$", "", token)
     while token and token[-1] in _TRAILING_PUNCT:
         token = token[:-1]
+    token = _LINE_SUFFIX_RE.sub("", token)
     if not token:
         return None
     if any(c in _GLOB_OR_PLACEHOLDER_CHARS for c in token):
--- a/scripts/check_doc_gate.py
+++ b/scripts/check_doc_gate.py
@@ -54,8 +54,9 @@ EXIT_CONFIG_ERROR = 3
 # exist in the repo itself. `-` is in the lookbehind for the same reason:
 # home-dir slugs like "-*-tinyagentos/memory/MEMORY.md" embed a repo prefix
 # after a hyphen, and the glob `*` sits BEFORE the match so the glob filter
-# never sees it.
-_TOKEN_RE = re.compile(r"(?<![\w/-])(?:scripts|tinyagentos|docs|desktop)/[^\s`\"'|]+")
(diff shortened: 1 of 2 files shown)


jaylfc:
**nemotron-super review** 

VERDICT: LGTM
No blocking issues found.

jaylfc:
Good call taking the dedicated line-citation pattern over splitting on the first colon. Splitting would also truncate any path that legitimately contains one, and `docs/weird:name.md` survives here correctly. That part is right and I verified it.

The strip is in the wrong place in `_clean_token`, though, and the result is that four common citation shapes still false-positive.

`re.sub(r":[0-9]+...$", ...)` is anchored to end-of-string but runs BEFORE the `_TRAILING_PUNCT` loop, so any trailing punctuation defeats the anchor. Measured against this branch's own `_clean_token`:

```
scripts/foo.sh:123      -> 'scripts/foo.sh'        OK
scripts/foo.sh:123.     -> 'scripts/foo.sh:123'    citation ending a sentence
scripts/foo.sh:123,     -> 'scripts/foo.sh:123'    citation in a list
scripts/foo.sh:12:34    -> 'scripts/foo.sh:12'     file:line:col (ripgrep, compilers)
scripts/foo.sh:12-20)   -> 'scripts/foo.sh:12-20'  citation closing a paren
docs/weird:name.md      -> 'docs/weird:name.md'    OK, correctly preserved
```

Every one of those leaks a suffix into the existence check and fails. "See `scripts/foo.sh:123`." with a full stop is the ordinary way this gets written in prose, so this will fire in practice.

Worth naming why CI is green on it: `test_file_line_citation_is_ignored` covers the one shape that already works, the bare backticked form. The test is real and it passes, and the gate is still broken for the other four. A passing test over the happy shape is not evidence the class is handled.

Two-part fix, both in `_clean_token`:

1. Move the `re.sub` to AFTER the `_TRAILING_PUNCT` while loop, so punctuation is gone before the anchor is applied.
2. Wrap the suffix group in `(?:...)+` so a repeated `:line:col` collapses in one pass.

```python
_LINE_SUFFIX_RE = re.compile(r"(?::[0-9]+(?:-[0-9]+)?(?:,[0-9]+(?:-[0-9]+)?)*)+$")
...
    while token and token[-1] in _TRAILING_PUNCT:
        token = token[:-1]
    token = _LINE_SUFFIX_RE.sub("", token)
```

Verified against all of the above plus `scripts/foo.sh:1,5-9`, `docs/f.md` and `docs/f.md.`: every citation shape reduces to `scripts/foo.sh`, and `docs/weird:name.md` is still preserved.

Please extend the test to the four failing shapes rather than only the bare one, and confirm each fails on the current commit before the reorder. A test that has only ever been seen passing does not tell us the reorder did anything.

Unrelated to the above, flagging rather than blocking: `_TOKEN_RE` whitelists `scripts|tinyagentos|docs|desktop`, so a broken reference under `app-catalog/`, `tests/` or `changelog.d/` is invisible to the scan. That is pre-existing and not this PR's job, but it bounds what the invariants pass can claim to cover, and it is worth knowing before anyone reads a green invariants run as "all referenced paths are good".

jaylfc:
@coderabbitai full review

jaylfc:
Lead fix-forward: rebased onto dev over the #2369 doc-gate rewrite (the conflict source). The two changes compose rather than collide — the token regex now carries BOTH #2369's hyphen lookbehind AND this PR's `)]` exclusion, and the `:123` line-citation strip sits alongside #2369's `::` symbol split. All 54 tests across both doc-gate suites green, invariants clean on the rebased tree.

jaylfc:
Block resolved (lead-completed, no worker on this conflicted branch). The suffix strip now runs AFTER the trailing-punct loop with the repeated group, exactly as the block prescribed. Red-first evidence — all four shapes measured on the pre-reorder commit:

```
'scripts/foo.sh:123.'   -> 'scripts/foo.sh:123'    LEAKS
'scripts/foo.sh:123,'   -> 'scripts/foo.sh:123'    LEAKS
'scripts/foo.sh:12:34'  -> 'scripts/foo.sh:12'     LEAKS
'scripts/foo.sh:12-20)' -> 'scripts/foo.sh:12-20'  LEAKS
```

Post-reorder: all five parametrized shapes pass, `docs/weird:name.md` preserved, 60 tests green across both doc-gate suites, invariants clean. Removing the lead-blocked label.

PR #2308, opened 2026-08-05
Fix-forward PR 2187 (consent defer): wire the flag + tests

jaylfc:
CARD TITLE (intent, not commit subject): Fix-forward PR 2187 (consent defer): wire the flag + tests

Autonomous build of board card tsk-5tj7qq.

_do_approve never passed defer_binding to approve_request_record, so the
flag added in PR 2187 was dead code: the project_id-required 400 guard
still fired on the defer combo, and when project_id was present the token
was bound anyway despite defer_binding:true.

- Pass defer_binding=body.defer_binding at the call site.
- Add a 400 guard for defer_binding:true with an explicit (non-blank)
  project_id -- the two are contradictory.
- Route tests covering the defer combo (unbound token, unbound grants,
  no membership row, no a2a channel), the 400 guard, and the unchanged
  non-deferred path.

Red run on buggy head (before the fix):
  test_defer_with_project_scopes_no_project_id_succeeds_unbound -> 400 (expected 200)
  test_defer_with_explicit_project_id_returns_400 -> 200 (expected 400)

Lifecycle note on never-bound deferred identities:
An unbound deferred identity is minted with project_id=None on both the
token and every grant. It is currently indistinguishable from a global
identity: no expiry, no pending binding flag, no visibility surface in
the Permissions app. A follow-up card should track (a) a lifecycle/expiry
mechanism so deferred-but-never-bound grants do not persist forever, and
(b) a visibility surface so operators can see and reap stale deferred
identities.

Files:
 tests/test_routes_agent_auth_requests.py  | 239 ++++++++++++++++++++++++++++++
 tinyagentos/routes/agent_auth_requests.py |  61 ++++++--
 2 files changed, 286 insertions(+), 14 deletions(-)

## Summary by CodeRabbit

- **New Features**
  - Added an option to approve agent access without immediately assigning it to a project.
  - Deferred approvals issue unbound tokens and grants, allowing project assignment later.
  - Added validation to prevent deferred approval when a specific project is selected.
  - Standard approvals continue to assign project access and synchronize related permissions.

- **Bug Fixes**
  - Improved approval handling to prevent inconsistent project bindings and registrations.

### DIFF
--- a/tests/test_routes_agent_auth_requests.py
+++ b/tests/test_routes_agent_auth_requests.py
@@ -1380,3 +1380,242 @@ class TestConcurrentApproveSameIdentity:
         await registry.close()
         await auth_store.close()
         await grants.close()
+
+
+class TestDeferBindingApproval:
+    """PR 2187 fix-forward: defer_binding must be wired through _do_approve so the
+    admin's 'Assign later' choice actually mints an unbound token instead of
+    silently binding to a project."""
+
+    @pytest.mark.asyncio
+    async def test_defer_with_project_scopes_no_project_id_succeeds_unbound(
+        self, client, monkeypatch, tmp_path
+    ):
+        """defer_binding=true + project scopes + no explicit project_id returns 200
+        with an unbound token (no project_id claim), unbound grants, no membership
+        row, and no a2a channel. The project_id-required 400 guard is skipped."""
+        from tinyagentos.agent_registry_store import (
+            AgentRegistryStore,
+            load_or_create_signing_keypair,
+            verify_registry_token,
+        )
+        from tinyagentos.auth_requests_store import AuthRequestsStore
+        from tinyagentos.agent_grants_store import AgentGrantsStore
+        from tinyagentos.projects.project_store import ProjectStore
+
+        registry = AgentRegistryStore(tmp_path / "reg-defer.db")
+        await registry.init()
+        auth_store = AuthRequestsStore(tmp_path / "auth-defer.db")
+        await auth_store.init()
+        grants = AgentGrantsStore(tmp_path / "grants-defer.db")
+        await grants.init()
+        pstore = ProjectStore(tmp_path / "projects-defer.db")
+        await pstore.init()
+        priv, pub = load_or_create_signing_keypair(tmp_path / "keys-defer")
+
+        project = await pstore.create_project(
+            name="Defer Proj", slug="defer-proj", created_by="u"
+        )
+
+        # The agent REQUESTED project_tasks against a project, but the admin
+        # defers binding: no explicit project_id on approve, defer_binding=true.
+        record = await auth_store.create(
+            identity_claim="@defer-bot",
+            framework="defer-cli",
+            requested_scopes=["project_tasks"],
+            requested_skills=None,
+            reason="",
+            duration_secs=None,
+            project_id=project["id"],
+        )
+
+        monkeypatch.setattr(client._transport.app.state, "agent_registry", registry)
+        monkeypatch.setattr(client._transport.app.state, "auth_requests", auth_store)
+        monkeypatch.setattr(client._transport.app.state, "agent_grants", grants)
+        monkeypatch.setattr(client._transport.app.state, "project_store", pstore)
+        monkeypatch.setattr(
+            client._transport.app.state, "agent_registry_keypair", (priv, pub)
+        )
+
+        resp = await client.post(
+            f"/api/agents/auth-requests/{record['id']}/approve",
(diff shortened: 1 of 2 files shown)


jaylfc:
**nemotron-super review** 

jaylfc:
Reviewed. The core is sound and the security-critical claim in the docstring is **true** — I verified it rather than trusting it. Two things to fix before merge, one of which is the red CI, plus one question.

**Verified good, stated explicitly because it is the part that could have been a privilege escalation.** The docstring says deferred grants are minted unbound and "project-scoped calls 403 until assign-agent binds it". A grant row with `project_id=None` could plausibly have meant *any* project rather than *none*, so I traced enforcement:

- `check_agent_scope_for_project` compares `g.get("project_id") == project_id`, so a `None` grant matches only a `None` request and 403s on any real project.
- The other checker (`check_agent_scope`) ignores `project_id` entirely, which *would* have been a hole, but it guards only non-project scopes (the registry feed scope, a2a bus).
- Every project-scoped route uses the project-aware checker: `projects.py:488`, `projects.py:532`, `project_files.py:80`, `decisions.py:150`.

So deferral authorizes nothing, as claimed. Good, and worth the trace.

---

### 1. doc-gate is legitimately red, not a flake

The CI log is unavailable (0 bytes), so I reproduced it locally against `origin/dev`, running **both** subcommands as CI does:

```
invariants -> doc-gate: clean, exit 0
diff-gate  -> exit 1
  DOC-GATE FAIL: user-visible-changelog
  DOC-GATE FAIL: agent-manual -- agent-facing behaviour changed
```

Both are correct for this change: `defer_binding` is a new field on a public approve body, and it alters agent-facing behaviour. Please add a `changelog.d/2308-<slug>.md` fragment (the fragment mechanism from #2290, not a direct `CHANGELOG.md` edit), and either touch `docs/agent-manual/*.md` / `docs/agent-coordination.md` or add a `Docs-Reviewed:` trailer explaining why not. The agent manual is the stronger option here — an operator-visible approve flag that changes how a token is bound is exactly what that doc is for.

### 2. Deferring against an existing ACTIVE handle changes behaviour, is untested, and the error steers the wrong way

The guard became:

```python
if not defer_binding and project_id and set(granted_scopes) & _PROJECT_SCOPES:
```

So with `defer_binding=True` and a handle that already maps to an active identity, the multi-project ADD branch is skipped and control falls through to:

```
409 "handle '<h>' is already in use by active agent <cid>; pick a different identity_claim"
```

The 409 is defensible — deferring names no project, so there is nothing to add the identity to. **The remedy in the message is not.** "Pick a different identity_claim" tells the operator to mint a *second identity for the same agent*, which is the outcome the multi-project model (#1862) exists to prevent. Someone following that advice creates exactly the duplicate this code was written to avoid.

Please either give this path its own message ("this handle already has an active identity; approve with an explicit project_id to add it to a project, or bind later via assign-agent") or state deliberately that deferring on an existing handle is unsupported.

Three defer tests were added and none covers this: `test_defer_with_project_scopes_no_project_id_succeeds_unbound`, `test_defer_with_explicit_project_id_returns_400`, `test_non_deferred_project_binding_unchanged`. The existing-active-handle tests (`test_non_project_handle_collision_still_409` and the multi-project ADD case) all predate `defer_binding` and do not set it. So the one branch whose behaviour this PR changed is the one branch with no test. Please add it, red first.

### 3. Question: the contradiction guard is asymmetric

```python
if body.defer_binding and body.project_id and body.project_id.strip():
    raise HTTPException(400, "defer_binding cannot be combined with an explicit project_id")
```

The comment reasons that naming a project while deferring is a mistake worth rejecting up front. Agreed. But `effective_project` is computed as `body.project_id if not None else record.get("project_id")`, and the guard only inspects `body.project_id`. So deferring on a request whose *record* already carries a project silently discards it, while naming that same project in the body is a 400.

Same operator intent, two different outcomes, and the silent one is the easier to hit. Is that deliberate? If deferring should always ignore a record-supplied project, a line saying so would help; if not, the guard should consider `effective_project`.

### Minor, not blocking

`_PROJECT_SCOPES = {"project_tasks"} | _CANVAS_SCOPES` is defined three times in this file (lines 379, 631, and as `_SCOPE_PROJECT_SCOPES` at 857). Identical today, so harmless, but three copies of a security-relevant set is a drift risk. Pre-existing and not this PR's job; worth a follow-up card if you agree.

---

Fix 1 and 2 and I am happy to merge. The unbound-grant design is right and the comments explaining *why* each `effective_project` became `binding_project` are unusually good.

PR #2309, opened 2026-08-05
Fix-forward PR 2220 (os-events SSE): doc-gate + hook fixes

jaylfc:
CARD TITLE (intent, not commit subject): Fix-forward PR 2220 (os-events SSE): doc-gate + hook fixes

Autonomous build of board card tsk-jmctoa.

- New authenticated SSE endpoint at GET /api/os/events streams typed
  change events (kind + id only, never payload) filtered by kinds query param
- New shared client hook useOsEvents(kinds, onEvent) returns connected/stale
  flags, manages single per-client EventSource with exponential backoff
- Reuses existing EventBus SSE plumbing and auth middleware
- Backend tests cover auth gate, kind filtering, and multi-subscriber delivery
- Frontend tests cover hook mount, kind filtering, dedup, stale flag, and reconnect

Files:
 CHANGELOG.md                            |   6 +
 desktop/src/hooks/use-os-events.test.ts | 228 ++++++++++++++++++++++++++
 desktop/src/hooks/use-os-events.ts      | 113 +++++++++++++
 tests/test_os_events.py                 | 282 ++++++++++++++++++++++++++++++++
 tinyagentos/routes/os_events.py         | 121 ++++++++++++++
 6 files changed, 753 insertions(+)

## Summary by CodeRabbit

* **New Features**
  * Added an authenticated, real-time operating-system event stream.
  * Events include identifiers, types, and timestamps while omitting payload details.
  * Added reliable connection monitoring with freshness status and automatic reconnection after interruptions.
  * Supports filtering for relevant event types and safely handles duplicate or invalid events.

* **Documentation**
  * Added an Unreleased changelog entry describing the new event-stream capabilities.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ Versions follow semver beta: `1.0.0-beta.N`, bumped on each dev->master promotio
   `POST /api/desktop/screenshot` prominently featured. The agent-manual index now
   points at both the OS skill and the existing `taos-development-skill`. Draft for
   @taOS-dev review.
+- **OS-level typed change-event stream + `useOsEvents` hook.** A new authenticated
+  SSE endpoint (`GET /api/os/events`) streams typed change events carrying only the
+  event kind and id, never the payload, so apps can opt into live updates with a
+  single hook call. The shared `useOsEvents(kinds, onEvent)` hook manages a single
+  per-client connection, exposes `connected` and `stale` flags, and handles
+  reconnect with exponential backoff.
 - Projects gain a Notes area: title + markdown notes per project, readable and
   writable by the project owner or by an agent holding a project-bound
   `project_notes` grant (new requestable scope) (#2285).
--- a/desktop/src/hooks/use-os-events.test.ts
+++ b/desktop/src/hooks/use-os-events.test.ts
@@ -1,0 +1,228 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { renderHook, act } from "@testing-library/react";
+import { useOsEvents, OsEvent } from "./use-os-events";
+
+// ---------------------------------------------------------------------------
+// Mock EventSource -- must use a regular function (not arrow) so `new` works
+// ---------------------------------------------------------------------------
+
+type MessageListener = (e: MessageEvent) => void;
+
+interface MockEventSource {
+  url: string;
+  onopen: (() => void) | null;
+  onmessage: MessageListener | null;
+  onerror: ((e: Event) => void) | null;
+  close: ReturnType<typeof vi.fn>;
+  readyState: number;
+  _fire: (data: unknown) => void;
+  _fireError: () => void;
+}
+
+let lastEs: MockEventSource | null = null;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const MockEventSourceCtor = vi.fn().mockImplementation(function (this: any, url: string) {
+  this.url = url;
+  this.onopen = null;
+  this.onmessage = null;
+  this.onerror = null;
+  this.close = vi.fn();
+  this.readyState = 0; // CONNECTING
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  this._fire = (data: unknown) => {
+    (this as MockEventSource).onmessage?.({ data: JSON.stringify(data) } as MessageEvent);
+  };
+  this._fireError = () => {
+    (this as MockEventSource).onerror?.(new Event("error"));
+  };
+  lastEs = this as MockEventSource;
+});
+
+beforeEach(() => {
+  vi.stubGlobal("EventSource", MockEventSourceCtor);
+  MockEventSourceCtor.mockClear();
+  lastEs = null;
+});
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+});
+
+describe("useOsEvents", () => {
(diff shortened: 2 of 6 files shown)


jaylfc:
**nemotron-super review** 

VERDICT: Found issues
- desktop/src/hooks/use-os-events.ts:102: Effect missing dependencies ['kinds', 'onEvent']

jaylfc:
Reviewed. This is careful work: cleanup is in a `finally`, only `kind`/`id`/`ts` cross the wire exactly as the docstring promises, auth fails closed, and the proxy-buffering headers are right. One blocking defect, one design question, and the red CI.

### 1. BLOCKING — subscriptions and relay tasks leak if the generator is never started

The subscriptions and the two relay tasks are created **in the handler body**, before `gen()` exists:

```python
user_q  = await event_bus.subscribe(user_ch)
bcast_q = await event_bus.subscribe("broadcast")
relay_tasks = [asyncio.create_task(_relay(user_q)), asyncio.create_task(_relay(bcast_q))]
```

but every one of them is torn down **inside** `gen()`'s `finally`. An async generator that is closed without ever being iterated does not execute its body, so that `finally` never runs. Verified rather than assumed:

```
never-started generator -> cleanup ran? False
started generator       -> cleanup ran? True
```

So if the client disconnects between the handler returning and `StreamingResponse` beginning to stream, or anything else closes the generator before first iteration, you leak **two EventBus subscriptions and two never-cancelled tasks, per occurrence**. It is not a bounded leak either: the bus keeps those queues in `self._queues[channel]` forever, so every subsequent published event is copied into queues nobody drains, and the relay tasks keep moving them into a `merged` queue nobody reads.

**Fix:** move the `subscribe` calls and `create_task` calls *inside* `gen()`, immediately inside the `try`. Then the `finally` is guaranteed to pair with them. Worth a test that closes the response without consuming it and asserts the bus has no leftover subscribers.

### 2. Unbounded queues make a slow consumer a memory leak

There are three unbounded queues per connection: `user_q` and `bcast_q` from `EventBus.subscribe` (which does a bare `asyncio.Queue()`), plus `merged` here. The relays move events between them as fast as they arrive, with no backpressure and no cap.

A client that stops reading but keeps the socket open, or a stalled proxy, accumulates events without limit. SSE clients doing exactly this is the normal failure mode, not an exotic one.

The bus's unbounded queue is pre-existing and not this PR's fault, but this route is the first long-lived consumer to make it matter, and `merged` *is* new here. At minimum give `merged` a `maxsize` and decide explicitly what happens when it fills: dropping the oldest and emitting a `kind: "events.lagged"` frame would let the client resync, which is friendlier than silently blocking the relay. Say in the PR which behaviour you chose, because "queue fills, relay blocks, events silently stop" is the failure a reader will not predict from the code.

### 3. doc-gate is legitimately red

CI's log is unavailable (0 bytes), so I reproduced locally, running both subcommands as CI does:

```
invariants -> doc-gate: clean, exit 0
diff-gate  -> DOC-GATE FAIL: routes -- an API route module was added or removed
              exit 1
```

Correct: `tinyagentos/routes/os_events.py` is a new route module and `docs/agent-coordination.md` is untouched. A new user-visible SSE endpoint is exactly what that doc is for, so please document it rather than adding a `Docs-Reviewed:` trailer.

### 4. Use a changelog fragment, not `CHANGELOG.md`

The +6 to `CHANGELOG.md` predates the `changelog.d/` mechanism from #2290 and is the rebase-conflict magnet that change removed. Please move it to `changelog.d/2309-<slug>.md`.

### 5. Question: `id: {seq}` invites a resume the server ignores

Each frame carries `id: {seq}` where `seq` is a per-connection counter. In SSE, emitting `id:` is what makes a browser send `Last-Event-ID` on reconnect, and the module docstring is explicit that resume is best-effort via the bus replay buffer and *not* `Last-Event-ID`. So we advertise a resume capability we do not implement, and the ids are not even stable across connections since `seq` restarts at 1.

Either drop the `id:` line, or emit `event.trace_id` and honour `Last-Event-ID`. Dropping it is the smaller change and matches the documented behaviour.

### Also worth confirming

Every authenticated user subscribes to `broadcast`, so all of them see `kind` + `trace_id` + `ts` for anything published there. That is activity metadata rather than content, and the no-payload rule limits the blast radius, but I could not establish from the code what actually publishes to `broadcast`. Could you confirm it is intended to be all-users, and that nothing user-specific reaches it? Asking rather than asserting; I did not find a publisher to check against.

---

Fix 1 and the two CI items and I am happy. 2 and 5 I would take in the same push, but say so if you disagree on either.

jaylfc:
SUPERSEDED by card **tsk-ronapx**, closing this PR.

Reason: the branch has been red and unattended since 2026-08-06 (lanes are one-shot - they post [REVIEW] and exit, so the blocking review above could never reach a worker), and dev has moved far enough since that a rebase is a bigger job than a re-cut: changelog.d/ fragments replaced direct CHANGELOG.md edits and doc-gate has been rewritten twice.

The work is not lost. `exec/tsk-jmctoa` stays on the remote and tsk-ronapx makes merging it STEP 0, then carries every finding from the review above as an explicit demand - the subscription/task leak (with a red-first proof required), the unbounded `merged` queue, the `id: {seq}` line, the doc-gate and changelog-fragment items, and the three use-os-events fixes including the vacuous disconnect assertions.

Nothing here needs re-deriving; the replacement PR supersedes this one.

PR #2310, opened 2026-08-06
ci(security): document and ignore three unreachable cryptography CVEs

jaylfc:
Resolves `dec-26leeh`. Jay's decision, taken today after I put the evidence and three options to him.

## Why

`dependency-audit` has been red on every dev PR and on the master promote since `cryptography` 48.0.1 picked up advisories with no upgrade path. It is currently the **sole** failing check on four contributor PRs (#2287, #2288, #2208, #2070) and on both open dependabot PRs, one of which (#2293) is itself carrying two `immutable` CVE fixes. A check that is red on everything and clearable by nobody has stopped carrying information.

## What

Adds three `--ignore-vuln` entries, each documented inline with **its own** reachability argument, because they are not in the same subsystem and one blanket justification would have been wrong for one of them.

| CVE | PYSEC | Fixed in | Subsystem |
|---|---|---|---|
| CVE-2026-69249 | 3553 | 49.0.0 | X.509 chain-building verifier (exponential path building, DoS) |
| CVE-2026-69248 | 3554 | 49.0.0 | X.509 chain-building verifier (wildcard SAN escapes permittedSubtrees) |
| CVE-2026-69247 | 3552 | **50.0.0** | PKCS#7 EnvelopedData **decryption** (Bleichenbacher oracle) |

**Unreachable, verified by grep over the fully populated venv, not asserted:**
- For the two verifier CVEs, searching `x509.verification`, `PolicyBuilder`, `ServerVerifier`, `ClientVerifier`, `verify_directly_issued_by` returns exactly one package: `cryptography` itself. taOS uses this library for Ed25519 sign/verify, RSA (GitHub App JWT), Fernet, X25519, hashes and serialization.
- For the PKCS#7 CVE, nothing in taOS references `pkcs7`, `EnvelopedData` or S/MIME. The only site-packages hit for the vulnerable decrypt API is a **pygments PHP builtin-name wordlist** (a string in a lexer, not a call). The `azure` and `pycryptodome` hits are PKCS#7 **block padding** (`cryptography.hazmat.primitives.padding.PKCS7`), an unrelated construct that shares the name.

**No upgrade path**, confirmed with the resolver rather than the changelog: forcing `cryptography>=49.0` fails with `tinyagentos[proxy] depends on cryptography>=48.0.1,<49.0`. Every `litellm[proxy]` release including the `1.96.0.dev1` prerelease carries that pin, and 69247 is not fixed until 50.0.0 anyway.

## Proven both ways

[...]

PR #2312, opened 2026-08-06
Fix-forward PR 2306 (doc-gate config errors): 2 config-error paths still exit 1

jaylfc:
CARD TITLE (intent, not commit subject): Fix-forward PR 2306 (doc-gate config errors): 2 config-error paths still exit 1

Autonomous build of board card tsk-aerar5.

Pr #2306 left two config-error paths that still exited 1 instead of the
config-error code:

- Structurally invalid config (valid TOML, wrong shape, e.g. rules is a
  string) passed load_config and died later in the rule loop with an
  AttributeError.  Add _validate_config to check the shape after parsing.

- Non-UTF-8 config bytes raised UnicodeDecodeError inside tomllib.load
  (which decodes before parsing), uncaught by the TOMLDecodeError-only
  handler.  Catch that alongside the other config errors.

Also move EXIT_CONFIG_ERROR from 2 to 3: exit code 2 is argparse's own
usage-error exit, so a typo'd flag was indistinguishable from a broken
config.  Now 1 (violation), 2 (argparse), and 3 (config error) are all
mutually distinct.

Supersedes exec/tsk-thh54d (pr #2306).

Files:
 scripts/check_doc_gate.py | 48 +++++++++++++++++++++++++++++--
 tests/test_doc_gate.py    | 72 +++++++++++++++++++++++++++++++++++++++++++++++
 2 files changed, 117 insertions(+), 3 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration validation for missing, malformed, unreadable, or incorrectly structured configuration files.
  * Configuration errors now produce a distinct exit status from document-gate violations and command-line usage errors.
  * Preserved clear status reporting for successful checks and detected violations.

* **Tests**
  * Added coverage for invalid configuration formats, encoding issues, missing files, valid violations, and invalid command-line usage.


jaylfc:
**nemotron-super review** 

VERDICT: Approved
No blocking issues found.

jaylfc:
## Review: the fix is correct. The one red check is not yours.

CI has now run properly — this PR had **zero** workflow runs until 00:05Z because it was opened during the GitHub Actions outage and its `pull_request` events were dropped; a close/reopen re-emitted them. Worth knowing the green you see is only ~40 minutes old.

**Result: 17 success, 1 failure.** The single failure is `dependency-audit`, and it is **not caused by this PR**.

`dev` currently runs `pip-audit --ignore-vuln CVE-2026-3219` — the pip CVE only. The three cryptography ignores live in #2310, which is under a hold. So `dependency-audit` fails on *every* PR against `dev` until #2310 lands. Nothing in this branch can fix it.

**The substance is right**, verified rather than skimmed:

- `EXIT_CONFIG_ERROR` moved 2 → 3, with 2 explicitly reserved for argparse. That was the collision I raised.
- `_validate_config` is **actually called** (line 315), not merely defined — I checked, because a validator that is never invoked passes inspection while doing nothing.
- `UnicodeDecodeError` is caught alongside `TOMLDecodeError`/`OSError`/`ValueError` at the call site, closing the non-UTF-8 hole.
- All three tests exist, including one asserting the three exit codes are mutually distinguishable and one asserting argparse's usage code differs.
- STEP 0 was honoured: `git rev-list --count origin/exec/tsk-thh54d ^origin/exec/tsk-aerar5` = **0**, so this contains all of #2306 and supersedes it rather than duplicating.

I have not merged it: `dependency-audit` is a required check and it is red, which my gate refuses on regardless of cause. That is correct behaviour — the queue is blocked by the #2310 decision, not by this work.

When #2310 lands, this should go green unchanged. `tsk-thh54d` carries the release condition: close #2306 **and** its card together at that point, or the card becomes claimable and a lane rebuilds work this PR already contains.

jaylfc:
recreate-cla

PR #2316, opened 2026-08-09
Auth sessions never pruned: 37,077 entries, 4.8MB and growing

jaylfc:
CARD TITLE (intent, not commit subject): Auth sessions never pruned: 37,077 entries, 4.8MB and growing

Autonomous build of board card tsk-7i3lsx.

Files:
 tests/test_auth.py  | 65 +++++++++++++++++++++++++++++++++++++++++++++++
 tinyagentos/auth.py | 72 ++++++++++++++++++++++++++++++++++++++++++++++++++++-
 2 files changed, 136 insertions(+), 1 deletion(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Expired sessions are now automatically removed during session storage reads, writes, and application startup.
  * Session cleanup safely handles malformed or unreadable session data.
  * Valid and long-lived sessions are preserved during cleanup.
  * Administrators are warned when the active session count exceeds 50,000.

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -188,6 +188,24 @@ class TestAuthManager:
         assert token1 in mgr2._sessions
         assert mgr2._sessions[token1]["long_lived"] is True
 
+    def test_prune_at_startup_shrinks_the_file(self, tmp_path):
+        """Startup pruning must persist: the sessions FILE shrinks at init,
+        not merely the in-memory view (which _load() already prunes)."""
+        expired = {
+            f"tok{i}": {"user_id": "u", "expires_at": time.time() - 10, "long_lived": False}
+            for i in range(50)
+        }
+        expired["live"] = {"user_id": "u", "expires_at": time.time() + 3600, "long_lived": False}
+        sessions_file = tmp_path / ".auth_sessions"
+        sessions_file.write_text(json.dumps(expired))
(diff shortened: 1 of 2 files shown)


jaylfc:
**Lead review — fix is real, cleanup applied in 01d7338c, merging on green.**

The effective mechanism (prune inside `_load()` and `_save()`) is correct and well-tested. Three defects fixed on the branch:
1. The two prune methods were pasted into **both** classes; the `_PersistentSessions` copy referenced `self._sessions_file`, which that class doesn't have — an AttributeError waiting to be called. Deleted.
2. `_prune_sessions_on_startup` loaded, pruned a **throwaway dict** and never wrote back — the 4.8MB file would only shrink on the next session mint, and its own test was actually passing via the read-path prune, not startup. It now loads (which prunes) + saves once; added `test_prune_at_startup_shrinks_the_file`, which fails against the decorative version.
3. Log tuple reported `after -> after`; now `before -> after`.

91 auth tests green locally.

jaylfc:
**nemotron-super review** 

VERDICT: Minor issues found.
- tests/test_auth.py:136   test misnamed: does not test startup pruning (tests in-memory cleanup instead)
- tinyagentos/auth.py:210   _prune_sessions_on_startup reads sessions file twice (inefficient)

PR #2342, opened 2026-08-10
AgentLoop wiring: single per-agent serialization owner (router + desktop chat + scoped status)

jaylfc:
Lead design card tsk-icpt4i (design decision recorded on the card 2026-08-10: option a — AgentLoop replaces AgentChatRouter's per-agent lock; two serialization mechanisms cannot both own one agent).

## What
- **AgentChatRouter**: `_agent_locks` → per-agent `AgentLoop`. `_run_acp_turn`: QUEUED → return (turn-holder drives it); IMMEDIATE → drive, then iteratively drive the safe-point drain, each queued message keeping its own trace id. `drive_turn` failure is caught + logged and the drain still runs.
- **routes/taos_agent.py**: the desktop agent's `/chat` previously had NO serialization — two concurrent POSTs raced on the shared opencode session id. One lazy `app.state.taos_agent_loop` now owns it: concurrent POST gets a queued-notice NDJSON frame; the turn-holder surfaces queued messages in its stream tail at the safe point (full redrive deliberately out of scope, commented).
- **`GET /api/taos-agent/status`**: scoped payload only — `state`, `current_turn_id`, `queued_count`, `subagents[{id,task,state,started_at}]`. `result`/`error` deliberately stripped (card requirement).
- Design doc + both changelog fragments updated (the #2335 'not yet wired' claim is now true-by-removal).

## Wedge invariant, red-proven
`reach_safe_point()` runs in `finally` on both paths — a raise that skips it wedges the loop in WORKING forever and the agent goes permanently silent. Proof: mutating `_run_acp_turn` to the no-except/no-finally shape fails exactly the two wedge-gate tests (`2 failed`); restored, `141 passed` across the loop/router/endpoint suites, zero skips.

## Sink change to review
The chat sink's `final` branch no longer enqueues `_DONE`; `_drive`'s `finally` owns stream termination so queued-message frames can land before it. `error` keeps its early `_DONE` (fast-fail preserved; queued surfacing best-effort on that path).

Known limitation (commented in code): a queued desktop message surfaces text only — attachments of a queued POST are not carried.

### DIFF
--- a/changelog.d/tsk-icpt4i-agent-loop-wiring.md
+++ b/changelog.d/tsk-icpt4i-agent-loop-wiring.md
@@ -1,0 +1,12 @@
+### Changed
+
+- **Agent loop wiring**: `AgentLoop` is now the single per-agent serialization
+  owner. `AgentChatRouter` drives OpenClaw ACP turns through a per-agent
+  `AgentLoop` (replacing the per-agent lock) and the turn-holder drives
+  messages queued mid-turn at its safe point. The desktop taOS agent chat
+  endpoint serializes on one `AgentLoop` too — fixing a race where two
+  concurrent POSTs shared the opencode session with no serialization —
+  queueing concurrent messages and surfacing them in the turn-holder's stream
+  tail. New `GET /api/taos-agent/status` endpoint returns the desktop loop's
+  status scoped to state / current turn / queue depth / subagent descriptors
+  (subagent result/error payloads stay server-side) (#tsk-icpt4i).
--- a/changelog.d/tsk-rl2lfb-agent-loop.md
+++ b/changelog.d/tsk-rl2lfb-agent-loop.md
@@ -1,6 +1,6 @@
 ### Added
 
 - **Agent loop infrastructure**: new `tinyagentos.agent_loop.AgentLoop` library
-  for subagent delegation and safe-point message queuing. This is standalone
-  infrastructure, not yet wired into the taOS chat agent or routes
-  (#tsk-rl2lfb).
+  for subagent delegation and safe-point message queuing. Landed as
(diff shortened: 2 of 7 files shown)


jaylfc:
@coderabbitai full review

jaylfc:
Independent adversarial review completed (CodeRabbit remained rate-limited even on the forced full review — qodo billing-blocked, gitar free-plan, kilo doesn't fire on feat/* — so a second reviewer attacked the concurrency semantics with a 400-message chaos harness against the real AgentLoop).

**Serialization core: SAFE, with the mechanism now stated** — the loop's lock is never held across an await, so `handle_message` and the safe-point drain never yield to the event loop; the entire holder stretch (drain → re-handle → next drive) is atomic. No turn-handoff race, no double-drive, no drop-by-mutual-assumption, no cancellation wedge (proven under repeated re-cancellation), no rapid-double-POST hole, status route leaks nothing beyond the scoped fields.

**Two real defects found and fixed in `99998819` (red-first: new test fails the unfixed route, 71 pass with it):**
1. A message queued during a turn that then ERRORS was silently lost — the sink's early `_DONE` stranded the surfacing frames in a queue nobody read, contradicting the queued-notice promise, on the most common failure (model-proxy error). The generator now drains the settled queue and yields leftovers before the closing done; client-disconnect drops are logged instead.
2. Bare `asyncio.create_task(_drive())` vs asyncio's weak-ref contract: a GC'd drive task = desktop loop wedged in WORKING forever (every later POST gets the queued notice, and defect 1 then eats those too). Now supervised, same rationale as the router's `_acp_tasks`.

Plus: queued-notice copy no longer oversells, and the router drain comment now states the real (atomic) mechanism instead of an impossible interleaving. Merging when CI settles on the new head.

PR #2338, opened 2026-08-09
Hailo backend slice 6 (per design doc)

jaylfc:
CARD TITLE (intent, not commit subject): Hailo backend slice 6 (per design doc)

Autonomous build of board card tsk-ba2p2g.

Files:
 .../manifest.yaml                                  | 28 +++++++++++++++++++++
 .../models/llama3.2-3b-instruct-hef/manifest.yaml  | 28 +++++++++++++++++++++
 .../models/qwen2-1.5b-instruct-hef/manifest.yaml   | 27 ++++++++++++++++++++
 .../models/qwen2.5-1.5b-instruct-hef/manifest.yaml | 28 +++++++++++++++++++++
 .../qwen2.5-coder-1.5b-instruct-hef/manifest.yaml  | 29 ++++++++++++++++++++++
 5 files changed, 140 insertions(+)

## Summary by CodeRabbit

* **New Features**
  * Added five Hailo-10H NPU-accelerated AI models to the catalog.
  * Introduced DeepSeek R1, Llama 3.2, Qwen2, Qwen 2.5, and Qwen 2.5 Coder variants.
  * Added model metadata, chat capabilities, tool-calling support where applicable, download details, hardware requirements, and recommended configurations.
  * Included support for 2048-token context windows on supported Qwen and Llama models.

PR #2344, opened 2026-08-10
CI sweep test: model catalog manifests must be resolvable and integrity-pinned

jaylfc:
CARD TITLE (intent, not commit subject): CI sweep test: model catalog manifests must be resolvable and integrity-pinned

Autonomous build of board card tsk-brblrv.

Files:
 tests/test_model_manifest_integrity.py | 200 +++++++++++++++++++++++++++++++++
 1 file changed, 200 insertions(+)

### DIFF
--- a/tests/test_model_manifest_integrity.py
+++ b/tests/test_model_manifest_integrity.py
@@ -1,0 +1,200 @@
+"""Sweep test for model-catalog manifest integrity.
+
+Ensures every variant in app-catalog/models/*/manifest.yaml satisfies the
+resolver schema contract: non-empty backends with known targets, a 64-char
+lowercase hex sha256, a non-empty https download_url, and a positive size_mb.
+
+A per-manifest allowlist tracks pre-existing sha256 debt until the catalog
+is filled in.  The intent is zero entries.
+"""
+from __future__ import annotations
+
+import glob
+import re
+from pathlib import Path
+
+import yaml
+
+# Known target enums -- the resolver only accepts values produced by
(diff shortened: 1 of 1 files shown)


jaylfc:
**nemotron-super review** 

jaylfc:
Post-merge lead audit (the dispatcher's gate auto-merged this as a green test/doc PR at 07:15:56Z, racing the review a gate-shaped PR owes — recording it here after the fact):

**Not inert, proven live**: green on the clean catalog; breaking one manifest (`size_mb: -5`) fails the sweep with the right error; restore, green again.

**Two defects found and dispositioned:**
1. `KNOWN_TARGETS` was hardcoded — the card explicitly said derive it from `capabilities.py`. The copy would have bounced the pending Hailo PR (#2338) the moment it merges (`targets: [hailo]` unknown to the literal set). Proven both directions with a simulated hailo branch + manifest: hardcoded FAILS the future case, derived set PASSES it and still fails genuinely-broken manifests. **Fixed on dev: `8b2d0a2c`** (derivation by regex over the capabilities source, with a collapse guard).
2. The `_SHA256_ALLOWLIST` quietly documents a real finding worth more than a buried comment: **~115 catalog manifests ship without sha256**, and `download_manager` treats empty as skip-verification — so essentially the whole model catalog downloads unverified today. The ratchet (new manifests must pin) is right; the burn-down is carded separately.

Sharp edge kept from the build: the allowlist keys on manifest id, so an allowlisted manifest adding a NEW variant also escapes the sha256 rule for that variant. Acceptable while the list burns down.

PR #2346, opened 2026-08-10
Chat migration S1: exporter + transformer for taOS chat -> bus (#2150)

jaylfc:
CARD TITLE (intent, not commit subject): Chat migration S1: exporter + transformer for taOS chat -> bus (#2150)

Autonomous build of board card tsk-yfy5j5.

- ChatExporter reads ChatMessageStore and produces import-batch envelopes
- Explicit identity_map, whole-batch failure on unmapped author
- Flattened body from content_blocks, non-empty when blocks exist
- 64KB serialized limit, oversized content delegated to file_writer
- Deterministic ordering by (created_at ASC, id ASC)
- reply_to preserved for non-deleted parents, omitted for orphans
- Idempotent on (source, source_id), re-run produces byte-identical output
- 18 pytest tests covering all acceptance criteria

Files:
 tests/test_chat_exporter.py       | 440 ++++++++++++++++++++++++++++++++++++++
 tinyagentos/chat/chat_exporter.py | 142 ++++++++++++
 2 files changed, 582 insertions(+)

### DIFF
--- a/tests/test_chat_exporter.py
+++ b/tests/test_chat_exporter.py
@@ -1,0 +1,440 @@
+from __future__ import annotations
+
+import json
+import time
+from pathlib import Path
+
+import pytest
+import pytest_asyncio
+
+from tinyagentos.chat.chat_exporter import ChatExportError, ChatExporter, flatten_body
+from tinyagentos.chat.message_store import ChatMessageStore
+
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+@pytest_asyncio.fixture
+async def store(tmp_path):
+    s = ChatMessageStore(tmp_path / "chat.db")
+    await s.init()
+    yield s
+    await s.close()
+
+
+def _make_file_writer(root: Path):
+    async def writer(source_id: str, content: bytes) -> str:
+        dest = root / "chat-export" / f"{source_id}.txt"
+        dest.parent.mkdir(parents=True, exist_ok=True)
+        dest.write_bytes(content)
+        return f"chat-export/{source_id}.txt"
+
+    return writer
+
+
+def _identity_map(user_ids: list[str], agent_ids: list[str] | None = None) -> dict[str, str]:
+    m: dict[str, str] = {}
+    for uid in user_ids:
+        m[uid] = f"@{uid}"
+    for aid in (agent_ids or []):
+        m[aid] = f"@{aid}"
+    return m
+
+
+# ---------------------------------------------------------------------------
+# flatten_body unit tests
+# ---------------------------------------------------------------------------
+
+@pytest.mark.asyncio
+async def test_flatten_empty_blocks():
+    assert flatten_body([]) == ""
+    assert flatten_body(None) == ""
+
+
+@pytest.mark.asyncio
+async def test_flatten_text_blocks():
+    blocks = [
+        {"type": "paragraph", "text": "Hello"},
+        {"type": "code", "lang": "py", "text": "print(1)"},
+    ]
+    assert flatten_body(blocks) == "Hello\nprint(1)"
+
+
+@pytest.mark.asyncio
+async def test_flatten_skips_empty_text():
+    blocks = [
+        {"type": "paragraph", "text": "Hello"},
+        {"type": "image", "url": "http://x/img.png"},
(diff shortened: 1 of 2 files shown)


jaylfc:
**nemotron-super review** 

VERDICT: No blocking issues found
- No blocking issues found

jaylfc:
Lead review: FIX-FIRST — three card-spec violations plus one undecided disposition, verified by running the code against the real ChatMessageStore. The skeleton is right (real store in tests, deterministic output, whole-batch author refusal exactly per spec), so this completes rather than bounces.

**Blocking, all traceable to card tsk-yfy5j5's own text:**
1. **Oversize path destroys blocks** (chat_exporter.py:129-137): card rule 5 says oversized content goes to project Files and the message carries a *ref* — this code sets `blocks=[]` and writes only the flattened text; a 70KB image block vanishes permanently. The PR's oversize test uses a pure-text block so it can't see this.
2. **Non-text-block messages brick the whole channel** (104-107): image/file-only messages (producible today via routes/chat.py:377) raise ChatExportError and fail the entire channel. Card rule 3's load-bearing rule is body is NEVER empty when blocks exist — the fix is flattening non-text blocks to descriptive text (`[image: …]`), not refusing. Whole-batch refusal is for forged authors, not valid message shapes.
3. **Timestamp-tie ordering violates causality** (sort key `(created_at, id)` with uuid ids): a reply can sort before its parent on a tie. Card rule 7 explicitly requires ordering survive 'including messages sharing a timestamp'. `test_thread_ordering_preserved` only proves determinism, not order — a weak test wearing a strong name.
4. **`state="streaming"/"error"` partials export as authentic history** (bridge_session.py creates both): interrupted agent responses become permanent bus records. Lead disposition: exclude `state != 'complete'` like deleted messages, documented and tested.

**Documented-drop, not blocking:** attachments/reactions/embeds/metadata/pins are outside the agreed envelope `{from, thread, body, blocks, ts, source, source_id, reply_to}` — deliberate control-plane minimalism, but the exporter must SAY what it drops (docstring + explicit test listing dropped fields), and attachment migration needs its own card before the live run.

**Adjudication of kilo's 4:** #1 confirmed (blocker 1), #4 confirmed (blocker 2), #3 confirmed-minor (the `content` fallback masks empty-body detection — dropping it in the fix round), #2 refuted (cross-channel reply chains cannot exist in this store; thread parents are channel-scoped by construction).

Lead is completing the fixes on this branch; label comes off after re-review.

jaylfc:
Lead pushed the fix batch (`30833341`, one commit): oversize path now writes the FULL envelope (blocks intact) to the ref file with the body carrying `[oversized content exported to: <ref>]` per card rule 5; non-text blocks flatten to `[type: identifier]` placeholders so image-only messages export instead of bricking the channel (card rule 3), with the `content` fallback dropped; `state != complete` messages excluded like deleted ones; stable topological tie-break within equal-timestamp groups so a reply never precedes its parent (card rule 7); module docstring now lists every intentionally-dropped field with a test documenting the drop.

All four fixes red-proven individually against the unfixed code (each failed for exactly the reviewed reason — REDPROOF.txt in the build record). 23/23 exporter tests + 43 store-adjacent tests green. Three pre-existing tests legitimately updated by the spec change (oversize ref-file contract, image placeholder) — each with rationale in the commit.

Label comes off when CI settles green; attachments migration gets its own card before any live run.

jaylfc:
Post-merge adjudication of kilo's re-review (2 findings on the fixed head — both CONFIRMED against the code, both now fixed forward on dev with red-first proof, `7f1544ea`):

1. **`_causal_tiebreak` dropped cycle members** — a `thread_id` cycle in a same-timestamp group (crafted self/mutual reference; the store doesn't validate) never reaches indegree 0 and vanished from the export. Cycle members now append in original order — no topological order exists for them, but a migration tool never silently drops a row.
2. **Oversize truncation could still exceed 64KB serialized** — the body was capped against the raw limit, then the ref note landed on top with envelope-field overhead uncounted; the bus would reject the result. Budget now derives from the serialized envelope overhead.

Process note, honestly: this is the second PR today I merged in the same command chain as the bot-verdict read — the check happened but nothing gated on it. Both kilo findings were real this time. The read-adjudicate-THEN-merge separation is now recorded discipline.

PR #2349, opened 2026-08-10
Alias editing UI in Agents app

jaylfc:
CARD TITLE (intent, not commit subject): Alias editing UI in Agents app

Autonomous build of board card tsk-kqffyv.

Files:
 desktop/src/apps/agents/RegistryPanel.test.tsx | 256 +++++++++++++++++++++++++
 desktop/src/apps/agents/RegistryPanel.tsx      | 119 +++++++++++-
 2 files changed, 374 insertions(+), 1 deletion(-)

PR #2351, opened 2026-08-10
Tests: harden collapse-revoked-agents coverage (Agents app)

jaylfc:
CARD TITLE (intent, not commit subject): Tests: harden collapse-revoked-agents coverage (Agents app)

Autonomous build of board card tsk-e2erm4.

Files:
 desktop/src/apps/agents/RegistryPanel.test.tsx | 43 ++++++++++++++++++++++++++
 desktop/src/apps/agents/RegistryPanel.tsx      | 34 ++------------------
 2 files changed, 46 insertions(+), 31 deletions(-)

### DIFF
--- a/desktop/src/apps/agents/RegistryPanel.test.tsx
+++ b/desktop/src/apps/agents/RegistryPanel.test.tsx
@@ -352,4 +352,47 @@ describe("RegistryPanel collapsed retired", () => {
     expect(retiredToggle).toHaveAttribute("aria-expanded", "true");
     expect(retiredPanel!).not.toHaveClass("hidden");
   }, 10_000);
+
+  it("treats unknown RegistryStatus as retired (robust to new statuses)", async () => {
+    const entries: RegistryEntry[] = [
+      { ...fakeEntry, canonical_id: "active-1", display_name: "ActiveAgent", status: "active" },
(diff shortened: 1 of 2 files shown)


jaylfc:
Lead block, and closing. This PR reverses a documented product decision under a "Tests:" title.

The deleted code carries the decision in its own comment: "Anything not explicitly retired (revoked/rejected/suspended) is visible by default **so future RegistryStatus values are never silently hidden**." The diff removes the `Other` catch-all section and folds every unknown status into collapsed `Retired` — exactly what PR #2300 attempted, which was closed for reversing that decision. The question (fail-open visible vs folded into Retired) is with Jay (recorded in the lead queue, item 2.5) and no build may pre-empt it.

The card (tsk-e2erm4) is a stale beta.42-era spec that predates the #2300 ruling; it also names `desktop/src/apps/AgentsApp/*` while this PR edits `desktop/src/apps/agents/RegistryPanel.tsx` — the same call sites PR #2349 is mid-flight on, so merging this would also conflict with that.

The legitimate half (vitest coverage of collapse/expand behavior) is being re-carded against the CURRENT documented semantics, including asserting the `Other` section renders for an unknown status rather than deleting it.

PR #2352, opened 2026-08-10
SECURITY: memory routes - path traversal + missing per-agent authz on the agent param (audit-verified)

jaylfc:
CARD TITLE (intent, not commit subject): SECURITY: memory routes - path traversal + missing per-agent authz on the agent param (audit-verified)

Autonomous build of board card tsk-6luty4.

Files:
 tinyagentos/routes/memory.py | 28 ++++++++++++++++++++++++++--
 1 file changed, 26 insertions(+), 2 deletions(-)

### DIFF
--- a/changelog.d/2352-memory-agent-traversal.md
+++ b/changelog.d/2352-memory-agent-traversal.md
@@ -1,0 +1,6 @@
+### Security
+
+- Memory routes reject any `agent` value that is not a single plain path
+  component (separators, `.`/`..`, NUL all 400): the caller-controlled name
+  becomes a filesystem path component of the qmd `dbPath`, and a traversal
+  value could previously address SQLite files outside `agent-memory/` (#2352).
--- a/tests/test_routes_memory.py
+++ b/tests/test_routes_memory.py
@@ -110,3 +110,56 @@ class TestAgentDbPath:
         req.app.state.agent_memory_dir = tmp_path / "agent-memory"
         p = _agent_db_path(req, "foo")
         assert p.endswith("agent-memory/foo/index.sqlite")
+
+
+@pytest.mark.asyncio
+class TestAgentPathTraversal:
+    """The ``agent`` param is caller-controlled and becomes a path component
+    of the dbPath handed to qmd serve. Anything that is not a single plain
+    component must 400 before any qmd call happens (#2352)."""
+
+    async def test_browse_rejects_dotdot(self, client_with_qmd):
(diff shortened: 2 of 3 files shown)


jaylfc:
Lead review + completion (tsk-6luty4). The traversal half was real and the fix direction right; three findings, all resolved on the branch (cfe4f60e):

1. **The authz half of the title was not implemented** — `find_agent` was imported and never used, advertising an intent the diff didn't carry. Verified against the auth middleware before deciding what to do about it: `/api/memory/*` is NOT in the agent-token allowlist, so agent bearer tokens cannot reach these routes at all — the `agent` param is reachable only by the authenticated desktop session, and that user browsing any agent's memory is by design (the Memory app). So "missing per-agent authz" is a non-finding under the current auth model; it becomes real work if memory paths ever join the agent allowlist or under multi-user separation. Recorded here so the next audit doesn't re-flag it blind. Import dropped; scope of the PR is honestly the traversal fix.
2. **The sanitizer accepted multi-segment names** (`a/b` passed through). No legitimate agent name contains a separator; rejecting outright is smaller and closes the widened surface. Rewritten to a single-component check (separator / dot-segment / NUL → 400) with the resolve+containment belt kept for symlinked entries.
3. **Zero tests.** Added six traversal tests through the real routes (browse GET, search POST body, delete-chunk) plus a clean-agent control, and a changelog fragment.

Red-proof against the unfixed helper (dev's `memory.py`, new tests only):

```
FAILED tests/test_routes_memory.py::TestAgentPathTraversal::test_browse_rejects_dotdot
FAILED tests/test_routes_memory.py::TestAgentPathTraversal::test_browse_rejects_backslash
FAILED tests/test_routes_memory.py::TestAgentPathTraversal::test_browse_rejects_bare_dots
FAILED tests/test_routes_memory.py::TestAgentPathTraversal::test_browse_rejects_multi_segment
FAILED tests/test_routes_memory.py::TestAgentPathTraversal::test_search_rejects_traversal_in_body
FAILED tests/test_routes_memory.py::TestAgentPathTraversal::test_delete_rejects_traversal
6 failed, 7 passed
```

After the fix: 13/13. All four `_agent_db_path` call sites resolve the path OUTSIDE their try blocks, so the 400 propagates rather than being swallowed into a 502 (checked, not assumed). Merges via gate on green; `awaiting-review` clears with this review once CI reports on cfe4f60e.

jaylfc:
**nemotron-super review** 

Issue #2106, opened 2026-07-22
feat(agents): consistent existence-hiding 404 across all agent-registry routes


hognek:
## Problem

The agent-registry surface currently leaks existence: an unauthorized caller can distinguish "agent ID does not exist" (404) from "agent ID exists but you do not own it" (403). This affects every route that fetches a registry record, checks it exists, then authorizes.

Affected routes in `routes/agent_auth_requests.py`:

- `create_scope_request` — fetches registry record → 404 on unknown → `require_owner_or_admin` → 403 on not-owned
- `approve_scope_request` — same pattern
- `deny_scope_request` — same pattern
- `approve_auth_request` (consent approve) — same pattern
- `deny_auth_request` (consent deny) — same pattern

The `GET /api/agents/registry/{id}` route in `routes/agent_registry.py` already documents "owner or admin; else 404" — it is existence-hiding. The scope-request and consent routes should match that.

## Context

PR #1921 added the scope-request flow. During review jaylfc noted the 404-vs-403 pattern and wrote (https://github.com/jaylfc/tinyagentos/pull/1921#issuecomment-5046717840):

> The 404-unknown vs 403-not-owner distinction is the file-wide pattern and matches the sibling consent approve/deny, so making only these two routes existence-hiding would introduce a new inconsistency rather than remove one. [...] **happy to switch the whole agent-registry surface to existence-hiding 404 as a separate pass if preferred.**

The create-path reorder in #1921 moved authz before the scope-vocabulary check, but not before the record-existence check — so the leak remains.

## Proposed solution

Make every agent-registry route existence-hiding: return 404 for both "does not exist" and "exists, not yours". The pattern is:

1. Fetch the registry record (without revealing whether it exists)
2. Authorize the caller (owner/admin check)
3. If authz fails → 404 (same as "not found")
4. Only then proceed with scope validation, grant writes, etc.

The `GET /api/agents/registry/{id}` route already does this — the scope-request and consent routes should follow the same pattern.

## Affected code

- `tinyagentos/routes/agent_auth_requests.py` — the scope-request create/approve/deny handlers and the consent approve/deny handlers
- Tests that assert 403 on not-owned will need to become 404 assertions

## References

- PR #1921 (merged): added scope-request flow
- jaylfc's comment: https://github.com/jaylfc/tinyagentos/pull/1921#issuecomment-5046717840
[...]

PR #2356, opened 2026-08-10
Existence-hiding 404 across ALL agent-registry routes (issue #2106, reported by hognek)

jaylfc:
CARD TITLE (intent, not commit subject): Existence-hiding 404 across ALL agent-registry routes (issue #2106, reported by hognek)

Autonomous build of board card tsk-vzkbao.

Convert create_scope_request, approve_scope_request, and deny_scope_request
to return the same 404 response for both non-existent canonical_ids and
authenticated non-owners, matching the pattern already used by
GET /api/agents/registry/{id}.

Server-side logs distinguish 403-not-owner from 404-unknown; only the
response is uniform.

Updated existing tests asserting 403 to assert 404, and added red-first
identical-response tests for each converted route.

Timing: non-owner path performs the same work as before (registry lookup
plus authz check); no new fast path on the not-found side.

Files:
 tests/test_agent_scope_requests.py        | 125 +++++++++++++++++++++++++++++-
 tinyagentos/routes/agent_auth_requests.py |  32 +++++---
 2 files changed, 144 insertions(+), 13 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Unauthorized scope-request actions now return `404` instead of `403`, preventing agent existence disclosure.
  * Create, approve, and deny operations consistently match the status and response body for unauthorized and nonexistent agents.
  * Registry updates, deletions, token rotations, and organization changes now return consistent `404` responses to unauthorized callers.
  * Unknown or inactive agents follow the same secure response behavior.

* **Documentation**
  * Documented resource-existence hiding across owner-gated registry routes.

## Red-first evidence (lead-run at merge ref)

All 7 new identical-response tests executed at the merge ref `origin/dev` (5910eb13), before the change:

```
FAILED tests/test_agent_scope_requests.py::test_create_scope_request_non_owner_and_nonexistent_identical
FAILED tests/test_agent_scope_requests.py::test_approve_scope_request_non_owner_and_nonexistent_identical
FAILED tests/test_agent_scope_requests.py::test_deny_scope_request_non_owner_and_nonexistent_identical
FAILED tests/test_agent_registry.py::TestRegistryWriteExistenceHiding::test_patch_non_owner_and_nonexistent_identical
FAILED tests/test_agent_registry.py::TestRegistryWriteExistenceHiding::test_delete_non_owner_and_nonexistent_identical
FAILED tests/test_agent_registry.py::TestRegistryWriteExistenceHiding::test_rotate_tokens_non_owner_and_nonexistent_identical
FAILED tests/test_agent_registry.py::TestRegistryWriteExistenceHiding::test_org_put_non_owner_and_nonexistent_identical
7 failed, 20 deselected in 17.72s
```

## Timing statement (card point 4)

No obviously faster branch was introduced on the not-found side. Every touched route performs the same single `store.get(canonical_id)` (or registry lookup) on both branches; the non-owner branch additionally does one in-memory id comparison before returning the identical response. Constant-time behaviour is not claimed.

### DIFF
--- a/changelog.d/2356-registry-existence-hiding.md
+++ b/changelog.d/2356-registry-existence-hiding.md
@@ -1,0 +1,7 @@
+### Security
+
+- All owner-gated agent-registry routes are now existence-hiding: a caller who
+  does not own an agent gets the same 404 as a nonexistent id, on the
+  scope-request create/approve/deny routes and on registry PATCH, revoke,
+  rotate-tokens, and org update. Previously a 403-vs-404 difference disclosed
+  whether an agent id existed (issue #2106, reported by hognek) (#2356).
--- a/docs/agent-coordination.md
+++ b/docs/agent-coordination.md
@@ -396,6 +396,15 @@ that SAME canonical_id instead:
 - `POST /api/agents/registry/{canonical_id}/scope-requests/{req_id}/deny`:
   owner/admin only.
 
+All owner-gated registry routes are existence-hiding (#2106): an authenticated
+caller who is not the owner gets the same 404 body as a nonexistent
+`canonical_id`, on the scope-request create/approve/deny routes above and on
+registry PATCH, DELETE (revoke), rotate-tokens, and `PUT /api/agents/{id}/org`.
+Agents must not treat a 404 from these routes as proof an id does not exist,
+and must not expect a 403 to distinguish "exists, not yours". Admin-only
+lifecycle routes (approve/reject/suspend/reactivate) still 403 non-admins
+before any lookup, which discloses nothing.
+
 Requested scopes are validated against the same closed `VALID_SCOPES` vocabulary
 as the consent flow. `project_tasks` and the canvas scopes still require an
 explicit `project_id`; `decisions_read` / `decisions_write` (and the other global
--- a/tests/test_agent_registry.py
+++ b/tests/test_agent_registry.py
@@ -870,7 +870,7 @@ class TestRegistryWriteExistenceHiding:
                 "/api/agents/registry/does-not-exist", json={"display_name": "Stolen"}
             )
         assert resp_owned.status_code == resp_missing.status_code == 404
-        assert resp_owned.json() == resp_missing.json()
+        assert resp_owned.content == resp_missing.content
         # And the record was not modified.
         check = await registry_client.get(f"/api/agents/registry/{cid}")
         assert check.json()["display_name"] == "Hidden Agent"
@@ -881,7 +881,7 @@ class TestRegistryWriteExistenceHiding:
             resp_owned = await mallory.delete(f"/api/agents/registry/{cid}")
             resp_missing = await mallory.delete("/api/agents/registry/does-not-exist")
         assert resp_owned.status_code == resp_missing.status_code == 404
-        assert resp_owned.json() == resp_missing.json()
+        assert resp_owned.content == resp_missing.content
         # And the record was not revoked.
         check = await registry_client.get(f"/api/agents/registry/{cid}")
         assert check.status_code == 200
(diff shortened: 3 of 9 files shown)


jaylfc:
**nemotron-super review** 

VERDICT: Log messages incorrectly state 403 when raising 404 for non-owner cases
- tinyagentos/routes/agent_auth_requests.py:951   (log message in _authorize_scope_request_creation function says 403-not-owner but raises 404)
- tinyagentos/routes/agent_auth_requests.py:983   (log message in approve_scope_request function says 403-not-owner but raises 404)
- tinyagentos/routes/agent_auth_requests.py:1189  (log message in deny_scope_request function says 403-not-owner but raises 404)

jaylfc:
Lead review (tsk-vzkbao, issue #2106). The lane's fix is correct and well-tested: the three scope-request routes now return the not-found response for authenticated non-owners, the true cause is logged server-side, and the byte-identical comparison tests (status + body, non-owner vs nonexistent) are exactly the right assertion shape. The consent approve/deny routes the issue also listed turn out not to leak: both check `user.is_admin` before any lookup, so a non-admin gets the same 403 whether the request id exists or not.

Pushed 0b632ebb completing the class the issue's own goal statement names ("make every agent-registry route existence-hiding"): four owner-gated write routes in `routes/agent_registry.py` still had the 404-then-403 pattern — PATCH, DELETE (revoke), rotate-tokens, and PUT `/api/agents/{id}/org`. All four now use the same idiom as the GET route the issue cites as the reference implementation. Byte-identical tests added for each with a real non-admin user (and side-effect checks that the record was not modified/revoked); the PATCH test was proven red against the old `require_owner_or_admin` behaviour before restoring. The orphaned import is removed, and the doc-gate changelog fragment is in. 73 tests across both files pass locally.

Merging via the gate once checks are green on the new head.

jaylfc:
@coderabbitai full review

jaylfc:
Bot findings adjudicated (999f7ab8):

**Folded** — both bots' byte-identity point was right: `.json()` compares normalized objects and cannot back the "byte-identical" claim, so all seven identical-response tests now compare `resp.content`. CodeRabbit's same-caller point was also right: the three scope-request tests sent their nonexistent probe from the admin fixture client; they now use the same non-owner client, since the contract under test is what one unprivileged caller can distinguish. The four registry write routes also gained the unknown-vs-not-owner server-side logs the scope-request routes already emit (CodeRabbit's consistency point).

**Not folded, with reasoning** — "reject inactive agents on the deny route": the deny route's None-only check predates this PR and is semantically deliberate, not an oversight. Create and approve gate on `status == "active"` because both can mint or add grants, which must never happen for a suspended agent. Deny writes only a refusal and grants nothing, and blocking it for inactive agents would leave pending scope requests un-cleanable exactly when an agent gets suspended. The 404 detail on deny ("agent not found", no "or not active") matches its actual check, so no message lies. Changing this would be a behavior change outside the card's existence-hiding scope; if the asymmetry should close, that's its own card.

jaylfc:
@coderabbitai full review

jaylfc:
CodeRabbit round 2, both folded in the new head: the Major was real — a malformed/inactive bearer token got 401/403 on an existing target but 404 on an unknown one (the not-found check runs before auth), an existence oracle through the credential-error path. `_authorize_scope_request_creation` now converts `check_agent_identity` raises into the uniform 404 with the true cause logged; regression test added (suspended agent's valid-signature token, byte-identical responses) and proven red against the unguarded call. Log labels renamed 403-not-owner → 404-not-owner to match the actual responses.

jaylfc:
@coderabbitai full review

jaylfc:
Merging with the CodeRabbit status on e44e7216 explicitly adjudicated, not trusted: its 'pass' there is the rate-limited fake green (description 'Review rate limited', never ran, retriggered twice with a 10-minute wait each). It is not being counted as a review. What is counted: CodeRabbit's two REAL reviews on 74ba0d29 and 999f7ab8 (every actionable finding from both rounds folded or adjudicated with reasoning above), Kilo's fresh pass on this exact head (19:36Z), green CI across the board, and the fact that the final delta consists solely of CodeRabbit's own round-2 findings implemented with a mutation-proven regression test.

PR #2357, opened 2026-08-11
Wallpaper fit options (fill/fit/stretch/center/tile), persisted PER DEVICE

jaylfc:
CARD TITLE (intent, not commit subject): Wallpaper fit options (fill/fit/stretch/center/tile), persisted PER DEVICE

Autonomous build of board card tsk-jar622.

Files:
 desktop/src/App.tsx                                |  15 +-
 desktop/src/apps/SettingsApp.tsx                   |  23 ++++
 desktop/src/components/Desktop.tsx                 |   6 +-
 desktop/src/stores/__tests__/wallpaper-fit.test.ts | 151 +++++++++++++++++++++
 desktop/src/stores/theme-store.ts                  |  45 ++++++
 desktop/src/theme/tokens.css                       |  48 +++++--
 6 files changed, 271 insertions(+), 17 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added wallpaper fit controls in Desktop & Dock settings.
  * Choose from fill, fit, stretch, center, or tile display modes.
  * Wallpaper preferences are saved separately for each device and restored automatically.
  * Added letterbox backgrounds for supported fit modes.
  * Applied wallpaper sizing consistently across desktop and mobile layouts.
  * Invalid or unavailable preferences automatically fall back to the fill mode.

## Red-first evidence (lead-run at merge ref)

The PR's test file executed at the merge ref `origin/dev` (eda3670d), before the fix:

```
 ❯ src/stores/__tests__/wallpaper-fit.test.ts (7 tests | 7 failed) 24ms
 FAIL  src/stores/__tests__/wallpaper-fit.test.ts > wallpaper-fit (per-device, CSS-driven) > setWallpaperFit stores the value under the current device id
TypeError: useThemeStore.getState(...).setWallpaperFit is not a function
 FAIL  src/stores/__tests__/wallpaper-fit.test.ts > wallpaper-fit (per-device, CSS-driven) > the stored value is isolated from another device id
TypeError: useThemeStore.getState(...).setWallpaperFit is not a function
 Tests  7 failed (7)
```

Additionally the invalid-value fallback test was proven red against a validation-skipping mutation of `loadWallpaperFit` on the PR head (1 failed | 6 passed), then the mutation reverted.

### DIFF
--- a/changelog.d/2357-wallpaper-fit-options.md
+++ b/changelog.d/2357-wallpaper-fit-options.md
@@ -1,0 +1,6 @@
+### Added
+
+- Wallpaper fit options in Settings → Desktop & Dock: fill, fit, stretch,
+  center, and tile. The choice is persisted per device (localStorage, keyed
+  by a locally minted device id that is never sent to the server), so each
+  screen keeps the fit that suits its aspect ratio (#2357).
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -259,16 +259,6 @@ export function App() {
   // box (#58). An explicit user choice is always honored and never overridden.
   usePerfAutoDetect();
 
-  // Persist a stable per-browser-profile device id (never sent to the server)
-  // so the per-device wallpaper-fit preference is isolated from other devices.
-  useEffect(() => {
-    const KEY = "taos-wallpaper-device-id";
-    if (typeof window === "undefined") return;
-    if (!localStorage.getItem(KEY)) {
-      localStorage.setItem(KEY, crypto.randomUUID());
-    }
-  }, []);
-
   // Sync the persistent backend notification feed into the bell (desktop and
   // mobile both render NotificationCentre under this component).
   useServerNotifications();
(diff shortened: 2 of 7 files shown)


jaylfc:
**nemotron-super review** 

jaylfc:
Lead review (tsk-jar622). The feature itself is correct and stays as built: per-device persistence via localStorage keyed on a locally minted device id, attribute-driven CSS for the five fits, Settings UI with proper `role="group"`/`aria-pressed`. Verified the base `.taos-wallpaper` rule keeps `background-repeat: no-repeat`, so `center` does not accidentally tile, and the attribute selectors out-specify the mobile media-query `cover`.

Pushed b778408f on top rather than bouncing, fixing what a merge would otherwise bake in:

- `wallpaperFitToClass` was exported but only its own test called it; the real mapping is the CSS attribute selectors. Removed with its describe block.
- The `letterboxBg` prop always evaluated to the same fallback `Desktop` already computes (`letterboxBg ?? effFallback` where both sides are the identical light/dark fallback), so the prop threading was inert. `Desktop` now reads `wallpaperFit` from the store like its other theme fields.
- The two `background-color: var(--wallpaper-fallback)` rules referenced a custom property defined nowhere, and every `.taos-wallpaper` element sets the fallback colour inline, which wins regardless. Removed, with a comment explaining why no rule is needed.
- The device-id-minting `useEffect` in `App.tsx` duplicated `getDeviceId()`, which mints lazily at store init before any effect runs.
- Two tests asserted state they had just `setState`-ed and two only exercised local string concatenation. The default/invalid-value pair now calls `loadWallpaperFit()` for real; the invalid-value test was proven red against a validation-skipping mutation before restoring. 7/7 pass, `tsc --noEmit` clean.
- Added the `changelog.d/2357-wallpaper-fit-options.md` fragment doc-gate was failing on, and restored the square-display comment the diff dropped from `tokens.css`.

Merging via the gate once checks are green on the new head.

PR #2637, opened 2026-08-30
fix(desktop): invert arbitrary-value white-overlay utilities in the light theme

hognek:
## Summary

Fixes the light-theme gap across ~126 sites. `desktop/src/theme/tokens.css` inverts an enumerated set of overlay utilities via attribute-selector rules, but it covered only the **plain-fraction** form (`bg-white/5`, `border-white/10`, …). The **arbitrary-value** form (`bg-white/[0.04]`, `border-white/[0.06]`, …) produces a distinct class token that matches no attribute selector — and the shared primitives (`card`, `button`, `tabs`) plus ~126 app surfaces use exactly that uncovered form. Their additive white overlays therefore survived the light-scheme flip and vanished on the light graphite background.

## Change

- Added 1:1 white→black inversion rules in `tokens.css` for every distinct arbitrary-value overlay utility present in the codebase (base + hover; `bg` + `border`), keyed off the same `:root[data-scheme="light"]` gate so dark is untouched.
- Normalised the pre-existing `hover:bg-white/[0.06]` rule from `0.05` → `0.06` so the whole arbitrary-value set is consistently 1:1.
- Added a regression test (`desktop/src/theme/__tests__/light-scheme-arbitrary-values.test.ts`) plus a changelog fragment.

## Acceptance / how it is tested

The test compares the **computed colour** across schemes — never the class name (the class is present in both light and dark, so a class-name assertion could never fail). It reads `tokens.css` via `node:fs` (a `css` raw import returns an empty string under this repo's vitest config) and asserts the load actually happened. `bg-white/[0.04]` computes `rgba(255,255,255,0.04)` in dark and `rgba(0,0,0,0.04)` in light; the same inversion is asserted for every arbitrary value.

Relevant frontend tests pass (`src/theme` + `src/components/ui`: 52/52).

## Merge gate

**Merge-gated — do NOT merge on a bot green.** jaylfc reviews and merges this personally (blast radius is every app's render).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved light-theme rendering for arbitrary-opacity white overlays.
  - Cards, buttons, tabs, borders, dividers, and hover states now display correctly against light backgrounds.
  - Corrected a hover overlay opacity mismatch for consistent styling.

- **Tests**
  - Added regression coverage for background, border, active-tab, divider, and hover behavior across themes.
  - Added checks to detect unsupported overlay variants.

- **Documentation**
  - Added a changelog entry describing the light-theme compatibility fix.

### DIFF
--- a/changelog.d/tsk-eqkpnt-light-theme-arbitrary-values.md
+++ b/changelog.d/tsk-eqkpnt-light-theme-arbitrary-values.md
@@ -1,0 +1,7 @@
+### Fixed
+- The light-theme compatibility layer now inverts arbitrary-value overlay
+  utilities (`bg-white/[0.04]`, `border-white/[0.06]`, and their hover
+  variants), not just the plain-fraction form (`bg-white/5`). The shared
+  primitives (card, button, tabs) and ~126 app surfaces used the
+  arbitrary-value form, which matched no attribute selector and so kept
+  additive white overlays that vanished on light backgrounds.
--- a/desktop/src/theme/__tests__/light-scheme-arbitrary-values.test.ts
+++ b/desktop/src/theme/__tests__/light-scheme-arbitrary-values.test.ts
@@ -115,6 +115,31 @@ describe("light-scheme arbitrary-value overlay inversion", () => {
     }
   });
 
+  it("maps each arbitrary-value hover token to its inverted colour on :hover", () => {
+    // jsdom cannot apply :hover for getComputedStyle, so assert the exact
+    // declaration value inside the :hover rule — selector AND mapped colour —
+    // rather than only that the token's selector appears somewhere in the css.
+    // A regression that maps hover:bg-white/[0.06] to rgba(0, 0, 0, 0.05)
+    // would pass the selector-presence loop above but fail here.
+    const normalized = TOKENS_CSS.replace(/\s+/g, " ");
+    const hoverRules: Array<[string, string, string]> = [
+      ["hover:bg-white/[0.03]", "background-color", "rgba(0, 0, 0, 0.03)"],
+      ["hover:bg-white/[0.04]", "background-color", "rgba(0, 0, 0, 0.04)"],
+      ["hover:bg-white/[0.05]", "background-color", "rgba(0, 0, 0, 0.05)"],
+      ["hover:bg-white/[0.06]", "background-color", "rgba(0, 0, 0, 0.06)"],
+      ["hover:bg-white/[0.08]", "background-color", "rgba(0, 0, 0, 0.08)"],
+      ["hover:bg-white/[0.1]", "background-color", "rgba(0, 0, 0, 0.1)"],
+      ["hover:border-white/[0.06]", "border-color", "rgba(0, 0, 0, 0.06)"],
+    ];
+    for (const [token, property, value] of hoverRules) {
+      const rule = `[class~="${token}"]:hover { ${property}: ${value}; }`;
+      expect(
+        normalized,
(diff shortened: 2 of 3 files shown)


jaylfc (changes requested):
Reviewed personally per the merge gate in your PR body. The CSS mechanics are right — new rules carry the identical specificity and `data-scheme` guard as the existing plain-fraction block, the dark arm is asserted, the red arm is real (dev's tokens.css swapped in → 6 of your 7 assertions fail), and the design-guide pre-flight on added lines is clean. But I'm requesting changes on **two coverage misses — the exact class of gap this PR exists to close:**

**D1 — `desktop/src/components/ui/tabs.tsx:29`: `data-[state=active]:bg-white/[0.08]`.** The class token includes the prefix, so `[class~="bg-white/[0.08]"]` provably does not match it (`el.matches(...) === false` in jsdom against your real tokens.css). Result in light scheme: the TabsList inverts but the active trigger stays `rgba(255,255,255,0.08)` — over a ~#f5f5f7 surface that's Δ≈1.6/255 vs the intended Δ≈18.8/255. The active-tab indicator is effectively invisible. Live consumer with no override: `MemoryApp.tsx:114-124`. The body names tabs as one of the three primitives fixed.

**D2 — `desktop/src/apps/FilesApp.tsx:1067,1140`: `divide-white/[0.04]`.** No `divide-white/[` rule exists in tokens.css, so file/archive row dividers render white-on-light and vanish. This is also the CRITICAL in Kilo's output (its check *conclusion* is success — read the output).

**Root cause, and the required fix shape:** the new test's `required` array (test lines ~108-129) is a hardcoded allowlist, so it cannot fail on a form it doesn't already know about. The repo has the correct pattern at `desktop/src/apps/AssistantStudioApp.test.tsx:214-260` — derive covered overlays from tokens.css via the `[class~="…"]` regex and diff against what `desktop/src` actually uses. Applying that derivation across desktop/src yields exactly D1 and D2 and nothing else in the arbitrary-value scope (22 distinct forms used, 20 covered). Please rebuild the test that way rather than appending two entries to the list — an allowlist test will go blind again on the next new utility.

Kilo triage so you don't burn time: the CRITICAL is D2 (real). The WARNING (hover alpha == base alpha) is **not** a regression — only one same-alpha pair exists in the codebase (`AgentMemoryTable.tsx:214`) and it's conditional and already a no-op in dark. The cwd SUGGESTION is moot (CI already runs `cd desktop && npx vitest run`).

Out of scope, mine to sweep separately (don't expand this PR): `divide-white/5` at 3 sites and ~40 uncovered prefixed plain-fraction forms (`active:`, `focus:`, `hover:bg-white/20`, `data-[state=unchecked]:`) — pre-existing and outside your stated arbitrary-value scope.

Fix D1/D2 with the derived test and this merges on my review.

jaylfc (approved):
Both named misses fixed: data-[state=active]:bg-white/[0.08] gets a scheme-guarded inversion keyed on [data-state="active"] (inactive triggers stay clear — asserted), and divide-white/[0.04] re-targets Tailwind's child-separator selector. The required fix shape landed: the hardcoded allowlist is replaced by a DERIVED coverage test (covered set extracted from tokens.css selectors, diffed against every class token in desktop/src, comments stripped), plus computed-colour assertions for both new forms in both schemes. CodeRabbit's one actionable item (hover assertions on selector presence only) is addressed by the dedicated exact-colour hover test. Desktop vitest ran green in CI on this head. Kilo red is 'Assistant request was rate limited' — outage class, no content; merging over it per standing ruling.

PR #2638, opened 2026-08-30
fix(gate): resolve deleted symbols through .py -> symlink typechange

hognek:
## Root cause

The deleted-symbols gate mishandles a PR that turns a `.py` file into a symlink (a git *typechange*). Two defects compound:

1. **`_get_symbols_at_ref` reads every `.py`-named tar member as source.** A symlink member is not a real source file — its symbols live at the link target. Worse, on a *dangling* symlink `git archive` emits no link-target member, so `tarfile.extractfile` raises `KeyError` on Python 3.12+ (`_find_link_target`). The gate died with a traceback and **no symbol list** — the "fail-closed with no message" shape (rc≠0, empty output).

2. **`_resolve_symbol` lets `spec_from_file_location` follow a symlink unconstrained.** The module path is joined under the extracted merge tree, but a symlink that escapes it (absolute target, or `..` climbing out) resolves back to the working tree / installed package — the exact editable-install path hook this loader exists to bypass. That "re-entry through the symlinked path" produces wrong findings or a hard crash.

Note (per jaylfc's measurement on #2587): the gate does **not** always segfault — on a clean deletion it emits a correct finding with a full symbol list. rc=139 with empty output is one failure mode; the dangling-symlink `KeyError` and the escape-path mis-resolution are the concrete defects fixed here.

## Fix

- **`_get_symbols_at_ref`** skips symlink/hardlink members (`member.issym() / member.islnk()`), so the target's symbols are never attributed to the symlink path and `extractfile` is never called on a member without a link target.
- **`_resolve_symbol`** resolves the module path to its real target with `Path.resolve(strict=False)` and pins the result inside the extracted merge tree (`relative_to(merge_result_dir.resolve())`). An escaping or dangling target is classified as **not importable** (a genuine deletion), never loaded from outside the tree. Self-loop symlinks also resolve safely to "not importable".

## Regression tests

- `TestResolveSymbolSymlinkTypechange` (unit): in-tree symlink resolves to the target; symlink losing the symbol is a deletion; escaping symlink is **not** followed; dangling symlink is not importable; self-loop does not hang/crash.
- `TestTypechangeSymlink` (integration): `.py -> symlink` keeping the symbol is silent; losing the symbol fires and names it; dangling symlink fires **without** the `KeyError` crash.

## Test evidence

```
$ PYTHONPATH=. python3 -m pytest tests/test_check_deleted_symbols.py -q
42 passed in 2.04s
```

Red-first: reverting only `scripts/check_deleted_symbols.py` makes `test_symlink_escaping_tree_is_not_followed` (re-entry returns `True`) and `test_py_to_dangling_symlink_fires_without_crash` (KeyError) fail.

## Scope

`scripts/check_deleted_symbols.py` + `tests/test_check_deleted_symbols.py` + one changelog fragment. No behavior change for non-symlink paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deleted-symbol checks for Python files changed into symlinks or hard links.
  * Correctly resolves in-project symlinks while rejecting dangling, self-referential, or out-of-tree links.
  * Preserves package behavior and relative imports for symlinked package initializers.
  * Prevented crashes when processing archive entries with invalid or missing link targets.

* **Tests**
  * Added coverage for symlink type changes, package initializers, symbol detection, and unsafe link paths.

### DIFF
--- a/changelog.d/tsk-qew5vk-deleted-symbols-symlink-typechange.md
+++ b/changelog.d/tsk-qew5vk-deleted-symbols-symlink-typechange.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- The deleted-symbols gate no longer crashes or mis-resolves on a `.py -> symlink`
+  typechange. `_get_symbols_at_ref` now skips symlink/hardlink tar members (a symlink is
+  not a real source file, and `tarfile.extractfile` raises `KeyError` on a dangling
+  symlink target on Python 3.12+), and `_resolve_symbol` resolves a symlinked module to
+  its real target while pinning the result inside the extracted merge tree — a symlink
+  that escapes the tree (absolute target or `..`) is treated as not importable instead of
+  re-entering the working tree and crashing or mis-reporting.
--- a/scripts/check_deleted_symbols.py
+++ b/scripts/check_deleted_symbols.py
@@ -115,6 +115,15 @@ def _get_symbols_at_ref(ref: str, repo_root: Path = REPO_ROOT) -> dict[str, str]
         for member in tar.getmembers():
             if not member.name.endswith(".py"):
                 continue
+            if member.issym() or member.islnk():
+                # A .py -> symlink typechange means this path is no longer a
+                # real source file: its symbols live at the link target. Skip
+                # it here (so the target's symbols are not attributed to the
+                # symlink path) and never call extractfile -- on a dangling
+                # symlink git archive emits no link target member and
+                # tarfile.extractfile raises KeyError (py3.12+), which is the
+                # "rc != 0, empty output" failure mode of this gate.
+                continue
             f = tar.extractfile(member)
             if f is None:
                 continue
@@ -169,6 +178,24 @@ def _resolve_symbol(merge_result_dir: Path, file_path: str, name: str) -> bool:
             abs_file = package_init
         else:
             return False
+
+    # Resolve symlinks so a .py -> symlink typechange loads the real target,
+    # and pin the resolved path inside the extracted merge tree. A symlink
+    # that escapes the tree (absolute target, or a relative target climbing
+    # out with `..`) would otherwise re-enter the working tree / installed
+    # package -- the exact editable-install path hook this loader exists to
+    # bypass -- producing wrong findings or a hard crash with no message.
+    # strict=False leaves a dangling target resolvable so the containment and
+    # is_file checks below classify it as not-importable instead of raising.
+    merge_root_resolved = merge_result_dir.resolve()
+    real_file = abs_file.resolve(strict=False)
+    try:
+        real_file.relative_to(merge_root_resolved)
+    except ValueError:
+        return False
+    if not real_file.is_file():
+        return False
+    abs_file = real_file
 
     name_parts = name.split(".")
     attr_name = name_parts[-1]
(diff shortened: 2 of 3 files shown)


jaylfc (changes requested):
Reviewed in depth. **The production fix is sound and stays** — I re-proved it by mutation with real git fixtures: the `origin/dev` gate crashes (`KeyError: "linkname 'pkg/_missing.py' not found"`) on a `.py -> dangling-symlink` typechange, your arm returns the correct violations and fails **closed** on every failure path I could construct (self-loop, mutual loop, symlink-to-dir, in-tree chain escaping the tree). The `_resolve_symbol` half is non-inert (a tar-skip-only half-fix goes silent on the escaping-symlink case; yours fires), and the tests are non-tautological (reverting only the script turns `test_symlink_escaping_tree_is_not_followed` and `test_py_to_dangling_symlink_fires_without_crash` red).

**Blocking on one defect, in the tests:** `TestResolveSymbolSymlinkTypechange._purge_package` (tests/test_check_deleted_symbols.py:326-330) `del`s every real `tinyagentos.*` entry from `sys.modules` and never restores them. Every later test that `mock.patch("tinyagentos....")` then patches a freshly re-imported module while the code under test holds the collection-time object — silent no-op patches. That's what turned `shards (3.13, 2)` red: 56 failures across container/hardware/health/installer suites, none of whose files this PR touches (siblings #2637/#2639 are green on the same shard). Attribution locally: your 4-file arm = 25 failed / 107 passed; same arm with only `::TestResolveSymbolSymlinkTypechange` deselected = 126 passed. After your test file runs, `sys.modules` holds 0 `tinyagentos.*` keys; with the class deselected, 111. This is the exact regression the pre-existing `test_resolve_symbol_does_not_leak_transitive_imports` docstring records.

**Fix:** replace `_purge_package` with snapshot-and-restore — save the `tinyagentos*` entries and reinstall them in `finally`, mirroring what `_resolve_symbol` itself already does with `touched`/`preexisting_keys`.

Two bot notes so you don't chase them: CodeRabbit's package-init item you already fixed in `a15f12f`; the RUF059 unused-`waived` items are a pre-existing pattern with lint green — ignore. **Kilo's line-206 WARNING (use original `file_path` in the predicate) is a false positive — do not apply it**: on `pkg/mod.py` deleted → replaced by package `pkg/mod/` whose `__init__.py` symlinks to an impl doing a relative re-export, your shipped predicate resolves correctly and Kilo's suggestion produces a false deletion.

The `Gate integrity` red is expected (protected gate file) — I apply the override at merge time; don't chase it. Fix the test-isolation defect and this merges.

jaylfc (approved):
Test-isolation blocker fixed exactly as specified: TestResolveSymbolSymlinkTypechange._purge_package now snapshots (was_present, old_obj) per key and every use restores in a finally, mirroring _resolve_symbol's own touched/preexisting_keys discipline. shards (3.13, 2) is green on 61cfb6465 — the direct evidence the eviction leak is gone. Kilo's :206 suggestion correctly NOT applied, with the symlinked-__init__ relative-re-export regression test locking the shipped predicate in. The unrestored _purge_package at :206 in the OLDER test class pre-exists on dev and is out of this PR's scope. Gate integrity red is the expected protected-file signature; applying the override now.

PR #2639, opened 2026-08-30
feat(projects): cross-project kanban aggregate read (Kanban Viewer S1)

hognek:
## Summary

Cross-project kanban aggregate read (Kanban Viewer S1) — `tsk-jbaj7e` (p94).

`GET /api/projects/tasks/aggregate` is a **READ-ONLY** cross-project board: it returns every project the caller is authorized to see, each with that project's tasks. No mutation route, no write scope.

## The risk this guards against

The whole risk is the aggregate leaking a project the caller is not entitled to. Authorization is enforced through `_authorize_task_actor` (session owner/admin, or an approved registry JWT bound to that specific project).

## Acceptance

- A session owner entitled to project A asking for the aggregate must **not** receive project B.
- An agent holding `project_tasks` for A (and not B) must **not** receive B.
- An admin sees every active project's board (the one case allowed to span).
- A caller with no projects / no grant receives an empty aggregate.

The auth-boundary cases assert the **absence** of the other project (`pid_b not in ids`, `tid_b` absent), not merely that the entitled caller sees their own board.

## Tests

`tests/test_routes_project_tasks_aggregate.py` — 7 tests, all passing (`7 passed`), covering owner boundary, agent boundary, admin span, empty-aggregate, shape, and status filter.

Status values are `open | claimed | closed`.

## Summary by CodeRabbit

- **New Features**
  - Added a read-only cross-project kanban view at `GET /api/projects/tasks/aggregate`.
  - Authorized session owners, administrators, and agents can retrieve tasks from accessible active projects in one request.
  - Added optional filtering by task status: open, claimed, or closed.

- **Bug Fixes**
  - Enforced per-project access checks and excluded archived, inaccessible, or stale project data.
  - Invalid status filters now return a clear validation error.

- **Documentation**
  - Documented endpoint access rules, authorization behavior, and status filtering.

### DIFF
--- a/changelog.d/2639-kanban-aggregate-read.md
+++ b/changelog.d/2639-kanban-aggregate-read.md
@@ -1,0 +1,2 @@
+### Added
+- New cross-project kanban aggregate read endpoint (`GET /api/projects/tasks/aggregate`) lets an authorized agent or session owner read tasks across all active projects they can access in a single call.
--- a/docs/agent-coordination.md
+++ b/docs/agent-coordination.md
@@ -345,6 +345,14 @@ The registry-JWT surface, by scope:
   `POST .../tasks/{id}/unquarantine` is also reachable, but LEAD-only: the
   route (`_authorize_project_lead`) refuses a plain project_tasks worker.
   It returns a quarantined card to the open pool and clears its strikes.
+  The cross-project aggregate `GET /api/projects/tasks/aggregate` (Kanban
+  Viewer S1) is also `project_tasks`: it returns EVERY project the caller may
+  see -- its own boards for a session owner/admin, or only the projects it
+  holds an active `project_tasks` grant for as an agent -- each with that
+  project's tasks. The grant (not the token's advisory `project_id` claim) is
+  the sole per-project gate, so an agent granted board A can never receive
+  board B in the aggregate. Optional `?status=open|claimed|closed` filters
+  tasks and is validated up front (400 on anything else).
 - **project_tasks_create**: `POST /api/projects/{pid}/tasks` (author new cards).
   This is a SEPARATE scope from project_tasks and is off by default; grant it
   explicitly when an agent needs to create cards.
--- a/tests/test_routes_project_tasks_aggregate.py
+++ b/tests/test_routes_project_tasks_aggregate.py
@@ -1,0 +1,276 @@
+"""Cross-project kanban aggregate (Kanban Viewer S1) authorization tests.
+
+``GET /api/projects/tasks/aggregate`` is a READ-ONLY cross-project board: it
+returns every project the caller is authorized to see, each with that project's
+tasks.  The whole risk is the aggregate leaking a project the caller is NOT
+entitled to, so these tests pin the authorization boundary hard -- a test that
+only proves an entitled caller sees their own board cannot fail on a leak, so
+every test here asserts the ABSENCE of the other project:
+
+  * a session owner entitled to project A must NOT receive project B;
+  * an agent holding ``project_tasks`` for A (and not B) must NOT receive B;
+  * an admin sees every active project's board (the one case allowed to span).
+"""
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+import pytest
+import pytest_asyncio
+from httpx import ASGITransport, AsyncClient
+
+from tinyagentos.agent_registry_store import mint_registry_token
+from taos_test_csrf import csrf_event_hooks
+
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+@pytest_asyncio.fixture
+async def members(client):
+    """Two non-admin member clients (alice, bob) sharing the admin `client` app.
+
+    Reuses the fully-initialised conftest `client` (which inits every store and
+    the admin user) and just adds two member identities + sessions on top.
+    """
+    app = client._transport.app
+    auth = app.state.auth
+
+    def _add(username: str, password: str) -> str:
+        invite = auth.add_user_invite(username, invited_by_username="admin")
+        auth.complete_invite(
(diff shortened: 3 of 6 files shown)


jaylfc (changes requested):
Gate review. I read the whole auth chain and ran mutations rather than trusting the green — the security core is **sound and I measured it**, but two things block merge, one substantive and one mechanical. Both are cheap and they land in the same push.

## What I verified positively (no action needed — recorded so you don't redo it)

**1. `check_agent_project_grants` faithfully reproduces the canonical chain.** I diffed it line-by-line against `_verify_agent_scope` + `check_agent_scope_for_project`: EdDSA verify → `sub` → registry record `status == "active"` → `token_min_iat` rotation cutoff → grant filter on `scope` + `project_id` + `_grant_unexpired`. Identical predicate. The one divergence is deliberate and correct: where `_verify_agent_scope` 403s on a token holding zero `project_tasks` grants, you return `(cid, {})` → empty aggregate, 200. That is the right call for a cross-project read and it is tested.

**2. The per-project gate is real defence-in-depth, not a tautology — measured.** My concern was that for the agent path the candidates are built *from* `agent_grants`, so `if project_id not in agent_grants` could never fire, making the "gate" decorative. So I mutated **both** enumeration paths to return every active project:

```python
# session, non-admin:
return await pstore.list_projects(status="active"), None, None   # was list_for_user(uid, ...)
# agent:
projects = await pstore.list_projects(status="active")           # was grant-keyed
```

`tests/test_routes_project_tasks_aggregate.py` → **10 passed**. Project B stayed out of the aggregate on both paths, so `_authorize_task_actor` demonstrably did the filtering (`_get_owned_project` 404 for the session, the grant-map miss for the agent). The layering holds under an over-broad enumeration. That is exactly the property the endpoint needs and you built it correctly.

**3. No regression from the signature change.** `tests/test_routes_projects_agent_tasks.py` + `tests/test_project_task_store.py` → **95 passed**.

## ⛔ Blocker 1 — the TOCTOU guard reads the stale snapshot, so it cannot fire on the race it names

`list_tasks_aggregate` (`routes/projects.py:896`):

```python
if project.get("status") != "active":
    continue
```

`project` is the loop variable from `candidates` — the **snapshot**. The freshly fetched project is sitting right there in `auth[2]`, returned by `_authorize_task_actor` (all three of its branches re-fetch), and it is discarded. In the actual race the docstring describes, the snapshot still reads `"active"` while the store already reads `"archived"`, so the check compares the stale value against itself and passes.

Your regression test cannot fail on this because it builds `arch` by calling `get_project` **after** `set_status`, so the dict it hands the loop already says `archived` — it exercises the branch, not the race.

Proof. Same test, except the snapshot is captured *before* the archive (which is what "archived between the listing and the check" means):

```python
arch_snapshot = await pstore.get_project(pid_arch)
assert arch_snapshot["status"] == "active"      # snapshot taken pre-archive
await pstore.set_status(pid_arch, "archived")   # ...THEN the race
monkeypatch.setattr(prj, "_caller_project_candidates",
                    lambda *a: ([keep, arch_snapshot], None, None))
```

```
E       AssertionError: archived project leaked into the aggregate
E       assert 'prj-ypb4t4' not in {'prj-lkrbpi', 'prj-ypb4t4'}
tests/test_toctou_probe.py:81: AssertionError
1 failed, 10 passed
```

Fix — check the fresh object, not the snapshot:

```python
fresh_project = auth[2]
if (fresh_project or {}).get("status") != "active":
    continue
```

With that applied: **11 passed** (my probe plus all 10 of yours — your existing archived test stays green, it is just no longer the only coverage).

To be clear on severity: the leak is a project the caller is **already entitled to**, so this is a contract violation ("active projects only"), not a cross-tenant leak. I am blocking on it because the guard and its test were added specifically to close Kilo #3, and merging as-is installs something that reads as protection but cannot fire — please make the test hand a stale `active` snapshot so it fails without the fix.

## ⛔ Blocker 2 — no changelog fragment (`doc-gate` red is legitimate)

Reproduced locally against the merge base:

```
DOC-GATE FAIL: user-visible-changelog -- user-visible behaviour changed; add a
changelog.d/<pr>-<slug>.md fragment (preferred) or a CHANGELOG.md line
```

A new public endpoint is squarely user-visible. Add `changelog.d/2639-kanban-aggregate-read.md`.

## Not your problem

`bot-review-gate` red is **not** a review defect. The discriminator says:

```
FAIL: only CodeRabbit output is a rate-limit stub and auto-generated scaffolding -- no review content (exit 1)
```

CodeRabbit is rate-limited, so it never reviewed this PR. Kilo did, and you addressed its #1/#3/#4. This review is the coverage; applying `bot-review-allow` at merge is mine, not yours. Don't push anything for it.

## Nits (take or leave, non-blocking)

- `_authorize_task_actor`: `assert agent_canonical_id is not None` is a production code path — `python -O` strips it and you'd return `(None, True, project)` as the actor. A plain `if ... raise` is cheaper than the surprise. Read-only endpoint, so no security consequence today.
- `auth_middleware.py`: `rf"^/api/projects/tasks/aggregate$"` has no `{}` placeholder — the `rf` prefix can just be `r`.

Push the two blockers and I'll re-review and merge; the auth core needs no further work from you.

jaylfc (approved):
Both blockers fixed exactly as prescribed. Blocker 1: the TOCTOU guard now reads the fresh object from _authorize_task_actor (auth[2]) instead of the stale candidates snapshot, and the regression test now captures the snapshot BEFORE the archive — the actual race shape, so it fails without the fix. Blocker 2: changelog.d/2639-kanban-aggregate-read.md present, doc-gate green. Nits (assert→raise, redundant rf prefix) also landed. bot-review-gate red is the CodeRabbit rate-limit stub already dispositioned in my gate review — this deep review IS the coverage; applying bot-review-allow.

PR #2748, opened 2026-09-03
security(decisions): server-stamped provenance for gate decisions (tsk-mul5pa)

hognek:
## tsk-mul5pa — server-owned kinds: gate decisions require server-stamped provenance

Closes the confused deputy in `POST /api/decisions`: an agent holding `decisions_write` could post a card whose `question` read as harmless while its `metadata.kind` was a privileged gate, and a human approving it minted a grant off caller-supplied metadata.

### Fix

- **`SERVER_RAISED_KEY = "_server_raised"`** in `routes/decisions.py`.
- `create_decision` strips the marker before persisting, so an API caller can never self-stamp it.
- Every `_apply_*_grant` (`execution_gate`, `device_pairing`, `delegation_gate`, `app_grant`) refuses to act when the marker is absent.
- The four internal raisers stamp it: `skill_exec.py` (execution-gate), `delegation.py` (delegation-gate), `device_pair_requests.py` (device-pairing), `app_permissions.py` (`app_grant_decision_payload`).

### Red-first (on current `dev`)

```
$ .venv/bin/python -m pytest tests/test_decision_gate_provenance.py -q
F.F.   [100%]
FAILED test_api_created_execution_gate_metadata_mints_nothing
E   assert True is False        # an execution grant WAS minted off caller metadata
FAILED test_api_created_app_grant_metadata_mints_nothing
E   assert {'net'} == set()     # 'net' WAS written to the app_grants ledger
2 failed, 2 passed
```

The two passing controls are the legitimate-path controls: the same gate raised through the internal path (server-stamped) still mints its grant on approval. After the fix the same command is `4 passed`.

### Proven pair, second handler

`test_decision_gate_provenance.py` carries a red + control pair for **execution_gate** and for **app_grant**, driving the real callers (agent `POST /api/decisions` with a genuine `decisions_write` grant, then owner answer via `POST /api/decisions/{id}/answer`), asserting on the grant store rather than the handler return value.

### Test updates

`test_routes_decisions.py` and `test_routes_app_permissions.py` previously drove gate metadata through the public route and asserted a grant — the exact path this card closes. Those tests now create the gate decision through the internal `decision_store.create(...)` path (server-stamped), which is how a gate is actually raised in production.

### Notes

- **"Render the grant" is deferred.** The `Fix shape, decided` paragraph calls out rendering the grant-bearing metadata as a separate follow-up; this card implements only the load-bearing provenance half. Say the word if you want the approval-surface rendering folded in here too.
- **`collab_delegation_gate` (#2048)** is not on `dev` yet; when it lands it needs the same `SERVER_RAISED_KEY` stamp in its raiser and the same guard in `_apply_collab_delegation_grant`. Flagging rather than opening this against #2048.

## Summary by CodeRabbit

- **Security**
  - Privileged approvals now require trusted server provenance before creating grants.
  - Caller-supplied provenance markers are ignored, preventing unauthorized grant creation.
  - Server-generated approval flows continue to support execution, app access, device pairing, and delegation.
  - Existing pending privileged approvals remain functional after the update.

- **Bug Fixes**
  - Improved protection against unauthorized grant creation from externally submitted approval decisions.

- **Tests**
  - Added regression coverage for trusted and untrusted approval scenarios.


jaylfc (changes requested):
Guard reads right: strip on create, refuse in every `_apply_*_grant` when the marker is absent, `/answer/agent` mints nothing. The red-first pair drives the real callers and asserts on the grant store — good.

One change before merge, and it is CodeRabbit's Major (decisions.py:619), which is real: every gate decision already **pending** on a host at deploy time has no `_server_raised`, so a human approving it after upgrade gets 200, the card goes `answered`, the agent receives the generic reply, and the grant silently never mints. Silent no-op on a legitimate approval is the worst of the three outcomes.

Fold: a one-time backfill in `DecisionStore._post_init` — rows where `status='pending'`, `metadata.kind` is one of `execution_gate|delegation_gate|device_pairing|app_grant`, and the marker is absent get `_server_raised: true`. Rationale: before this PR every gate kind was trusted unconditionally, so pre-marker pending rows are exactly as trusted as they were yesterday; the window closes at the moment of upgrade and nothing created through the public route after it can ever be stamped. Bound it to the pending set only (answered/superseded rows are inert).

Test it the way the existing-DB rule wants: seed a pre-marker pending `execution_gate` row directly into the table, construct the store over that file, approve through the real `POST /api/decisions/{id}/answer`, assert the grant IS minted; control: a pending row with a non-gate kind stays unstamped.

Ruling on the two notes:
- "render the grant" stays deferred — not this card.
- `collab_delegation_gate`: fold the stamp + guard into #2048 itself when you rebase it (it is your PR, no separate card).

PR #3002, opened 2026-09-12
fix(hailo): detect pre-existing hailo-ollama before installing (#2083)

### LINKED ISSUE #2083: install-hailo.sh assumes a clean machine: detect a pre-existing upstream hailo-ollama before installing
Found while a community tester (@doc62fr) worked through #1771 on real hardware, a Raspberry Pi 5 with an AI HAT+2. Nobody on the team has this hardware, so this only surfaced because he is testing it.

## The gap

`scripts/install-hailo.sh` assumes a clean machine. It does not check whether hailo-ollama is **already installed and running** from Hailo's own route (their installer, a desktop shortcut, or a manual build).

In the reported case the tester had upstream hailo-ollama serving on `0.0.0.0:8000`, which is its default. Our script remaps to 7836 by setting `Environment=OLLAMA_HOST=127.0.0.1:7836` in the unit it installs, so the two coexist invisibly: a working server on 8000 that taOS cannot see, no `hailo-ollama.service`, and nothing on 7836. From the user's side taOS simply appears broken while their Hailo setup works fine, which is a confusing place to leave someone.

## What it should do

Before installing, detect a pre-existing upstream instance (something listening on 8000 that answers `/api/tags`, or an existing binary/service not installed by us) and then either:

1. **Adopt it**: reconfigure the existing install to the taOS port and manage it under `hailo-ollama.service`, or
2. **Say so plainly**: report what was found, what will change, and what will be left alone, before touching anything.

Silently building a second server alongside the first is the one behaviour to avoid, since both then compete for the accelerator.

## Notes

- Port 7836 is correct and should stay. 8000 is the Django slot in taOS port hygiene and is already probed as a llama-cpp/vllm candidate, so leaving Hailo there makes those probes ambiguous.
- The same class of check probably applies to the other backend installers (rkllama in particular), which have the same greenfield assumption. Worth checking rather than fixing this one in isolation.
- Related: #1771 (the Hailo backend slice), #695 (app port hygiene).

hognek:
Fixes #2083.

`scripts/install-hailo.sh` assumed a clean machine and silently built a second hailo-ollama server (on port 7836) alongside a pre-existing upstream instance (Hailo's default, `0.0.0.0:8000`). The two coexisted invisibly: taOS saw no server on 7836, no systemd unit, while the user's own Hailo setup worked fine — leaving taOS appearing broken and the 8000 port probes ambiguous.

This adds a pre-install detection step (`detect_preexisting_hailoollama`) that checks `GET /api/tags` on 8000 (the script's existing idiom, matching lines that already probe `/api/tags`) and, when a pre-existing instance is found, reports plainly what was found and what would have changed, then exits — leaving the existing install untouched.

- No behaviour changed other than the added pre-flight check; port 7836 unchanged.
- `bash -n` clean; exactly 2 files changed (+10 changelog fragment).
- Fix written by the local taos engineer (qwen3.6-27b-a3b-coder-i1), verified by orchestrator.

## Summary by CodeRabbit

- **New Features**
  - Hailo-10H installation now checks for an existing Hailo-Ollama server on port 8000 before proceeding.
  - If a server is detected, installation reports the finding and leaves the existing service untouched. The server and worker installation processes continue without replacing it or creating a duplicate.

- **Documentation**
  - Updated the README and changelog with details about the pre-installation check and its behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,6 +476,10 @@ Run `curl -fsSL https://raw.githubusercontent.com/jaylfc/taOS/master/scripts/ins
 | OS packages added | python3 + venv + pip, git, curl, ca-certificates, libtorrent-rasterbar (model torrent mesh), Node.js 22 (qmd + SPA build), sqlite3, libsqlcipher (encrypted secrets), vulkan-tools (hardware detection), postgresql (LiteLLM virtual keys) |
 | User accounts created | The distro `postgres` system user is created when PostgreSQL is installed. A `litellm` Postgres role and database are created for virtual-key management. Everything else runs as the user who ran the installer. |
 
+### Hailo-10H install (`scripts/install-hailo.sh`)
+
+Installs `hailo-ollama` (the Hailo-10H LLM runtime) at a pinned ref, listening on port **7836** (port 8000 is the Django slot in taOS port hygiene and is already probed as a llama-cpp/vllm candidate). Before installing, the script checks for a pre-existing upstream hailo-ollama (something answering `GET /api/tags` on `0.0.0.0:8000`, Hailo's own default). If one is found it reports plainly what was detected and exits, leaving the existing instance untouched rather than silently building a second server that would compete for the accelerator and go unseen by taOS.
+
 ### Worker install (`scripts/install-worker.sh`)
 
 Linux uses a two-phase install. Phase 1 runs on the bare host; phase 2 runs inside the `taos-worker` LXC.
--- a/changelog.d/2083-hailo-preinstall-detect.md
+++ b/changelog.d/2083-hailo-preinstall-detect.md
@@ -1,0 +1,10 @@
+### Changed
+
+- The Hailo installer (`scripts/install-hailo.sh`) now checks for a pre-existing
+  hailo-ollama instance on upstream port 8000 before any install or systemd
+  mutation. When found (a server answering `GET /api/tags` with `"models"`), the
+  script prints what was detected, reports that it would have built a second
+  server on port 7836, and exits — leaving the existing instance untouched. This
+  prevents the silent coexistence bug from issue #2083 where an upstream
+  hailo-ollama (running on its default `0.0.0.0:8000`) would go unseen while
+  taOS built a second server and installed no systemd unit.
--- a/scripts/install-hailo.sh
+++ b/scripts/install-hailo.sh
@@ -240,6 +240,28 @@ detect_hailo() {
     return 0
 }
 
+# -------- (1.5) pre-install detection --------------------------------------
+
+# Checks for a pre-existing hailo-ollama instance running on upstream port 8000
+# (Hailo's own default / the Ollama-compatible tags endpoint). When found, logs
+# what was detected and exits — taOS must not silently build a second server.
+# Port 8000 is the Django slot in taOS port hygiene and is already probed as a
+# llama-cpp/vllm candidate, so a coexisting server there also makes those probes
+# ambiguous. See issue #2083.
+detect_preexisting_hailoollama() {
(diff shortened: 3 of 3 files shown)


jaylfc:
@hognek flagging this so the refresh does not read as a clearance: the `dev` merge at 04:00Z brings the branch up to date and CI is clean, but neither finding from my 2026-09-12 review is addressed. Current head still has:

```
scripts/install-hailo.sh:252:    local url="http://0.0.0.0:8000/api/tags"
scripts/install-hailo.sh:261:        exit 0
```

So finding 1 stands as written - the refusal path still exits 0, and `install-server.sh:595` / `install-worker.sh:1081` both signal only on non-zero, which means the auto-install path still prints nothing and still reports success while leaving a Hailo-10H present, no taOS backend on 7836, and the operator's own instance on 8000. That is the end state #2083 was filed about.

The branch is otherwise ready, so this is a small distance from merge. The minimum I need is the distinct non-zero exit for "refused: pre-existing instance" plus the matching branch in the two callers, so the message names the conflict instead of the generic `failed`. Finding 2 (detection is liveness-only, so a stopped upstream install still gets a second server built) I am happy to take as a follow-up rather than hold this PR for, if you would rather split it. The `0.0.0.0` to `localhost` nit is cosmetic - fold it in or leave it.

Holding the `lead-blocked` label until the exit code lands. Tell me if you would rather I take it from here.

jaylfc:
@hognek picking up my 2026-09-16 offer, since there has been no reply for six days and this is a user-facing installer bug sitting behind it. We are finishing it on our side rather than leaving it open.

To be clear about what that means: **your detection work is correct and is being kept, not rewritten.** I re-verified the current head `a4736c5a8` today and the detection logic, the `"models"` payload check and the placement in `main()` are all right. The card explicitly instructs the lane to `git fetch origin pull/3002/head` and cherry-pick your commit onto the branch unamended, with the exit-code fix as a separate commit on top. Your authorship and your changelog fragment carry across.

Two cards are now on the board:

- **`tsk-yol7hk`** (p87) - the blocking finding from my 2026-09-12 review. `detect_preexisting_hailoollama` ends in `exit 0`, and both auto-install callers (`install-server.sh:595`, `install-worker.sh:1081`) use `|| warn`, which fires only on a non-zero exit. So the chained install prints nothing and reports success while leaving a Hailo-10H present, no taOS backend on 7836, and the user's own instance on 8000. The card takes the `exit 3` route I suggested, plus the matching branch in both callers, plus the `0.0.0.0` to `localhost` nit.
- **`tsk-k6oboo`** (p80) - finding 2, split out as you and I discussed rather than held against this PR. Detection is liveness-only, so a stopped upstream install still gets a second server built.

One thing worth passing on to your agents, because it is not obvious from the tree: `tests/test_install_server.sh` and `tests/test_install_llama_cpp_health_gate.sh` exist but are referenced by no workflow and run in no CI job. A bash test added next to them would never execute. The pytest shards are what CI actually runs, so the card asks for `tests/test_install_hailo_preexisting_refusal.py` modelled on `tests/test_install_server_hardware_verify.py`, which gates the identical bug class and extracts the function verbatim from the real script.

I will close this PR as superseded once the replacement lands, and link it here. If you would rather finish it yourself after all, say so and I will park the cards instead - the work is only a few lines and it is genuinely your patch.

hognek:
Superseded — closing as planned. jaylfc's lane took this over via #3145 (tsk-yol7hk), merged 09-22 22:12: cherry-picked my detection commit (authorship + changelog fragment preserved) and added the exit-3 fix + both-caller branches + the refusal test. Finding 2 (liveness-only detection) is still tracked separately as tsk-k6oboo. Thanks for carrying the authorship through.

PR #3005, opened 2026-09-12
feat(grant): populate expires_at from duration_secs on approve

### LINKED ISSUE #2985: taOS #2148 follow-ups: populate expires_at (time-box grants) + Agents-app revoke surface
Follow-ups split out of #2148, whose main fix (a real revoke + an honest `assign-agent`) lands in a PR. Two acceptance items from that issue are deliberately NOT in it.

## 1. Time-boxed grants: nothing ever populates `expires_at`

`expires_at` exists in the `agent_grants` schema and is already honoured at check time — `_grant_unexpired` in `tinyagentos/agent_token_auth.py` is applied by `check_agent_scope`, `check_agent_scope_for_project` and `check_agent_project_grants`, and `GET /api/agents/registry/grants` filters on it — but no path ever SETS it, so a grant can only be revoked, never expired.

The scope request carries `duration_secs` (see `agent_scope_requests_store.add_grant`'s callers and the request create route), so the approve path could populate `expires_at = now + duration_secs` when the requester asked for a bounded grant.

Not folded into the revoke PR because #2148's acceptance explicitly says "Existing grant/approve flows unchanged", and populating the column changes what an approval grants. It should be its own change with its own tests (a request with `duration_secs` produces an expiring grant; an expired grant is refused by every check path; an unbounded request still grants without expiry).

## 2. Agents-app revoke surface (and the assign dialog's authoritative scope list)

An operator should not need the API to undo a grant they made in the UI.

- Show an agent's active grants (per project) with a Revoke action backed by the new `POST /api/projects/{id}/members/revoke-agent`.
- `desktop/src/apps/agents/AssignAgentToProjectDialog.tsx` sends only its three scope presets (`project_tasks`, `canvas_read`, `canvas_write`). Since `assign-agent`'s `scopes` list is now the agent's COMPLETE scope set for the project (that is the honest-reporting fix in #2148), re-assigning a project through that dialog revokes any scope the agent holds outside those presets — e.g. `files_read` or `project_notes` granted through a consent or scope request. The dialog must load the agent's current project scopes and show them, so the operator removes a scope on purpose rather than as a side effect of clicking "Assign".

Done in the #2148 PR (for the record): `AgentGrantsStore.revoke_grant` / `revoke_all_for_project`, the owner-or-admin `revoke-agent` route with project-activity audit logging, `assign-agent` reconciling the project grant set and failing loudly (500) if the reconciliation does not land, and the doc/changelog updates.

hognek:
Fixes #2985 (item 1 — backend half).

`expires_at` was already honoured by `agent_grants_store.add_grant` and checked by `agent_token_auth._grant_unexpired`, but the approve path never computed it from the request's `duration_secs`, so a time-boxed grant could only be revoked manually and never expired on its own.

This threads `duration_secs` through `approve_request_record`: a positive value produces `now + duration_secs` (timezone-aware ISO) passed to `add_grant`; None/zero/negative stays unbounded. The computation is extracted into a small pure helper `_expires_at_from_duration` and unit-tested directly (6 cases: positive→future, None/zero/negative/non-int→unbounded, short→still future).

Item 2 (Agents-app revoke surface in the desktop SPA) is out of scope here — this PR is the backend time-boxing only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Approved time-limited access grants now expire according to the requested duration, including when an existing agent is reused.
  * Grants without a requested duration remain unbounded.
* **Tests**
  * Added coverage for expiry calculations and approval behavior for time-limited and unbounded grants.
* **Documentation**
  * Added a changelog entry describing the grant expiration fix.

### DIFF
--- a/changelog.d/2985-grant-expires-at-duration.md
+++ b/changelog.d/2985-grant-expires-at-duration.md
@@ -1,0 +1,5 @@
+### Changed
+
+- `approve_request_record` now reads `duration_secs` from the auth-request record and passes
+  a computed `expires_at` to every `add_grant` call, so time-boxed grants actually expire.
+  Previously every approval created permanent grants even when `duration_secs > 0`. (taOS #2985)
--- a/tests/test_grant_expiry.py
+++ b/tests/test_grant_expiry.py
@@ -1,0 +1,50 @@
+"""Tests for the approve-path grant expiry: duration_secs -> expires_at.
+
+The store-level persistence of ``expires_at`` is covered by
+``tests/test_agent_grants_store.py`` / ``tests/test_grant_expiry.py``. This
+module asserts the *route-level* mapping injected by issue #2985: a scope
+request carrying ``duration_secs`` must produce a future, timezone-aware expiry
+on approval, and a request without it must stay unbounded.
+"""
+
+from datetime import datetime, timedelta, timezone
+
+import pytest
+
+from tinyagentos.routes.agent_auth_requests import _expires_at_from_duration
+
+
+def _parse(iso: str) -> datetime:
+    return datetime.fromisoformat(iso)
+
+
+class TestExpiresAtFromDuration:
+    def test_positive_duration_yields_future_expiry(self):
+        now = datetime.now(timezone.utc)
+        expires = _expires_at_from_duration(3600)
+        assert expires is not None
+        parsed = _parse(expires)
+        assert parsed.tzinfo is not None, "expiry must be timezone-aware"
+        delta = parsed - now
+        # allow a small clock skew window around the intended 1 hour
+        assert timedelta(minutes=59) < delta <= timedelta(hours=1, minutes=1)
+
+    def test_none_duration_is_unbounded(self):
+        assert _expires_at_from_duration(None) is None
+
+    def test_zero_duration_is_unbounded(self):
+        assert _expires_at_from_duration(0) is None
+
+    def test_negative_duration_is_unbounded(self):
+        assert _expires_at_from_duration(-60) is None
+
+    def test_non_int_duration_is_unbounded(self):
+        # a float or string must not produce a bogus expiry
+        assert _expires_at_from_duration(3660.5) is None
+        assert _expires_at_from_duration("3600") is None
+
+    def test_short_duration_still_in_the_future(self):
+        expires = _expires_at_from_duration(30)
+        assert expires is not None
+        parsed = _parse(expires)
+        assert parsed > datetime.now(timezone.utc) - timedelta(seconds=1)
--- a/tinyagentos/routes/agent_auth_requests.py
+++ b/tinyagentos/routes/agent_auth_requests.py
@@ -23,6 +23,7 @@ Security notes
 
 import asyncio
 import logging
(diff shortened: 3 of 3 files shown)


jaylfc:
@hognek same as #3002: picking this up after nine days rather than leaving it open, since it has been blocked on nothing but a merge conflict.

First, credit where it is due. **My 2026-09-12 blocking finding is properly addressed**, and I verified that today by mutation rather than by reading:

```
clean run on 8f357dff7:                                   8 passed in 12.33s
delete expires_at= from BOTH add_grant sites (696, 799):  1 failed, 7 passed
  FAILED TestApprovePathWiresExpiry::test_duration_secs_persists_future_expiry_on_grant
```

That is a genuine route-level test, not a vacuous one. It fails for the right reason when the wiring is removed, which is exactly what I asked for.

### One new finding, same class, one path over

There is a **third** wiring site neither of my earlier reviews named. `agent_auth_requests.py:605` passes `expires_at=expires_at` into `add_agent_to_project`, which forwards it to `add_grant` at :799. That path is entirely unasserted:

```
delete ONLY expires_at=expires_at at line 605:            8 passed in 6.86s
```

The whole suite stays green while the deferred-approval path silently reverts to unbounded grants. It is the same shape as the original finding - the wiring is the only thing that was ever missing, and the wiring is what is not asserted - so it should close before this merges.

Card **`tsk-65f2uh`** (p86) carries your commits forward by cherry-pick, resolves the conflict (it is one file, `agent_auth_requests.py`; everything else applies cleanly), adds the deferred-path coverage, and folds in the `agent_grants_store.py:277` "Phase 2" docstring update and the missing trailing newline.

### Separate observation, not carded and not against this PR

Two other callers of `add_agent_to_project` pass no `expires_at` at all, so their grants are unbounded: the assign-agent route at :910, and `approve_scope_request` at :1408, whose `ApproveScopeBody` (:142) has no `duration_secs` field to carry one. That is not a regression you introduced, and I do not think it should be fixed by widening this PR. It is a design question about whether those routes should be able to express an expiry, and I am raising it with Jay rather than putting a lane on it.

Same offer as on #3002: if you would rather finish this yourself, say so and I will park the card. The remaining work is a conflict resolution and one test.

hognek:
@jaylfc I went ahead and finished this myself before I saw #3147 was in flight — rebased onto current dev (resolve your one-file conflict), added `TestReusePathWiresExpiry` for the deferred/reuse path (third wiring site at the `add_agent_to_project` call), and proved it red-first: deleting `expires_at=expires_at` from that call fails the new test, restored passes. Also fixed a rebase drift (`memory_read` was removed from VALID_SCOPES \u2192 switched the approve-path test to `registry_feeds_read`), updated the stale Phase-1/2 docstring in `list_active_grants`, and restored the missing trailing newline. PR is MERGEABLE with 9/9 tests green.

Your call: merge mine directly, or keep #3147 going and I'll close mine. Either way the deferred-path coverage is covered.

hognek:
Superseded by #3147 (jaylfc's take-over: rebased the expires_at wiring onto dev + covered the DEFERRED approve path). Thanks for carrying the authorship forward.

PR #3067, opened 2026-09-15
installer hardware self-check silently skips when /api/hardware 401s pre-admin (cannot-see-reads-as-pass)

jaylfc:
CARD TITLE (intent, not commit subject): installer hardware self-check silently skips when /api/hardware 401s pre-admin (cannot-see-reads-as-pass)

Autonomous build of board card tsk-fh6otu.

fix installer hardware self-check auth and disk type detection

verify_hardware_capabilities() now reads the local auth token from
data/.auth_local_token and includes it as Bearer auth on the
/api/system/hardware/refresh POST. If the token is missing (pre-admin
fresh install), the check fails loud with a clear error instead of
silently skipping (cannot-see-reads-as-pass).

_detect_disk() now distinguishes microSD (sd) from eMMC (emmc) on
mmcblk devices by checking for eMMC boot partitions (mmcblkXboot0,
mmcblkXrpmb) and the sysfs device type attribute (MMC vs SD).

RED-FIRST

```
FAILED tests/test_install_server_hardware_verify.py::test_verify_dies_loud_when_local_token_is_missing - AssertionError: missing-token failure output does not mention the local auth token
1 failed in 29.43s
```

```
tests/test_install_server_hardware_verify.py tests/test_hardware.py -q
39 passed in 36.71s
```

Docs-Reviewed: No routes/catalog changes; hardware detection and installer script only.

Files:
 .../tsk-fh6otu-installer-hardware-auth-fix.md      |   3 +
 scripts/install-server.sh                          |  20 ++++
 tests/test_hardware.py                             | 102 +++++++++++++++++++++
 tests/test_install_server_hardware_verify.py       |  62 +++++++++++++
 tinyagentos/hardware.py                            |  20 +++-
 5 files changed, 205 insertions(+), 2 deletions(-)

## Summary by CodeRabbit

- **Bug Fixes**
  - Installer hardware verification now authenticates correctly using the local installation token.
  - Fresh installations without an available local token now stop with a clear error message instead of silently skipping verification.
  - Storage detection more accurately distinguishes eMMC drives from microSD cards, including devices without standard boot-partition indicators.


jaylfc:
**Lead review: held (lead-blocked), fix-forward card coming.**

The token half is right and stays: Bearer local token on the `/api/system/hardware/refresh` POST, loud `die` when `data/.auth_local_token` is missing, new installer test. Note for the record: on `origin/dev` the empty-profile branch already `die`s, so what this fixes is a 401 being misreported as "hardware-detection failure inside the controller", not a silent pass.

Blocking defect (CodeRabbit's finding, confirmed against the diff): `_detect_disk()` does `base = name.rstrip("0123456789")`, so `mmcblk0` becomes `mmcblk` and the probe checks `/sys/block/mmcblkboot0` / `/sys/block/mmcblkrpmb`. Linux keeps the index: `/sys/block/mmcblk0boot0`, `mmcblk0boot1`, `mmcblk0rpmb`. That branch is dead on every real eMMC host; detection only works through the `device/type` fallback. `test_mmcblk_with_boot_partitions_is_emmc` fakes `Path.exists` for exactly the wrong path, so the test encodes the bug and passes.

Fix-forward: build the probe from `{name}boot0` / `{name}rpmb`, correct both mmcblk tests to the real shapes, add an assertion that `mmcblkboot0` is never queried, red-first on this head. The kilo "strip the token" suggestion is declined (the server strips).

Issue #491, opened 2026-05-31
Low-RAM (4GB) hosts: core-only control plane, route everything else to the cluster


jaylfc:
## Policy

A 4GB system should run **only core taOS** (the FastAPI controller + SPA + lightweight stores/config). Everything resource-heavy — **models/qmd, inference backends, the taosmd memory pipeline, userspace container apps, and agent containers** — must deploy on the **cluster (or a cloud device)**, not the local host.

Install must **never hard-block** on 4GB: install the core, then **require + guide the user to add a cluster node or cloud device** before any model/agent/app/memory workload can run. Until a device is added, those features are gated with clear "add a device" guidance — not silent failure.

This is the control-plane / compute-plane split that fits the cluster + remote-desktop direction.

## Current state
- Agent deploy is **local-only** today (`deployer.deploy_agent`, incus on the controller host). Deploy-to-remote-target does not exist yet — so "add a cluster/cloud device to deploy" is aspirational until that lands. See #58 (deploy routing to LXC worker).
- #384 (closed via #483) added an **interim safety-net**: a deploy-time RAM preflight that rejects local deploy on low-RAM hosts with an actionable error instead of a silent incus-launch hang. It checks **local** host RAM only.

## Work required
1. **Deploy-to-remote-target** — route agent-container deploy (and userspace container apps) to a cluster worker / cloud device instead of the local host.
2. **Low-RAM gating** — on 4GB hosts, gate local qmd/model downloads + local deploy in the installer and Store, with "add a device" guidance (ties into hardware-tier templates #332 and store compatibility borders #331).
3. **Make the #483 RAM preflight target-aware** — block only when no capable target (local/cluster/cloud) exists, instead of always checking local RAM.
4. **Verify the installer never hard-blocks on 4GB** — core install must succeed; only workloads are gated.

## Acceptance
A 4GB box installs core taOS, runs the desktop/controller, and — after the user adds a cluster/cloud device — can deploy models, agents, memory, and apps to that device, with clear guidance at every gated step and no silent failures.

PR #3075, opened 2026-09-15
taOStalk S6: read-receipt ticks (supersedes tsk-kenvqs; bus receipts SHIPPED on taOSmd master, three-state contract)

jaylfc:
CARD TITLE (intent, not commit subject): taOStalk S6: read-receipt ticks (supersedes tsk-kenvqs; bus receipts SHIPPED on taOSmd master, three-state contract)

Autonomous build of board card tsk-4pmwcr.

taostalk S6: A2A read-receipt ticks on own messages

- Added desktop/src/lib/a2a-receipts-api.ts: getReceipts(messageId) and
  markSeen(messageId) calling GET /api/a2a/messages/{id}/receipts and
  PATCH /api/a2a/receipts through the existing desktop bus client.
- MessageList renders a three-state receipt tick per the contract:
  no row = sent (single Check), row with seen_at null = delivered
  (double CheckCheck, uncoloured), row with seen_at set = seen
  (double CheckCheck, text-accent, title tooltip with time).
  Missing receipts render as sent, never as unseen.
- ThreadPanel calls onMarkSeen for each visible incoming message
  (not own) when the thread opens.
- MessagesApp subscribes to /api/a2a/bus/stream for live receipt
  events (kind 'receipt') and refetches on channel focus and send.
- Accessibility: each tick carries an aria-label (Sent/Delivered/Seen).
- Added changelog fragment changelog.d/tsk-4pmwcr-taostalk-s6-receipts.md.

Vitest summary: 93 passed (93) across MessageList + ThreadPanel suites.

Files:
 changelog.d/tsk-4pmwcr-taostalk-s6-receipts.md     |   2 +
 desktop/src/apps/MessagesApp/index.tsx             |  80 +++++++++++
 desktop/src/apps/chat/MessageList.tsx              |  65 +++++++++
 desktop/src/apps/chat/ThreadPanel.tsx              |  15 ++-
 .../src/apps/chat/__tests__/MessageList.test.tsx   | 146 ++++++++++++++++++++-
 .../src/apps/chat/__tests__/ThreadPanel.test.tsx   |  74 ++++++++++-
 desktop/src/lib/a2a-receipts-api.ts                |  28 ++++
 7 files changed, 405 insertions(+), 5 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added per-recipient read receipts for sent messages, showing sent, delivered, or seen states.
  * Receipt updates now appear in real time.
  * Opening a thread marks incoming messages as seen.
  * Group conversations reflect whether all, some, or none of the recipients have seen a message.

* **Tests**
  * Added coverage for receipt states, group delivery scenarios, and marking messages as seen when threads open.


jaylfc:
## Lead review — CHANGES REQUESTED (CI is red, and the red is yours)

### 1. BLOCKER — unguarded `EventSource` crashes MessagesApp on mount, breaking 5 pre-existing tests

`spa-build` failed on head `cabec9e` (run 35022854520). Both the first run and the automatic retry
produced the identical failure set, so this is deterministic, not flaky:

```
Failed Tests 5
 FAIL src/apps/__tests__/MessagesApp.mobile.test.tsx > renders the mobile split-view with a Chats list pane when isMobile=true
 FAIL src/apps/__tests__/MessagesApp.mobile.test.tsx > renders the channel sidebar on desktop too (mobile shell is gated on isMobile)
 FAIL src/apps/__tests__/MessagesApp.ws.test.tsx   > unmounting cancels the pending reconnect
 FAIL src/apps/__tests__/MessagesApp.ws.test.tsx   > reconnect delay is jittered across instances
 FAIL src/apps/__tests__/MessagesApp.ws.test.tsx   > a remount after a zombie opens a fresh socket

ReferenceError: EventSource is not defined
 ❯ src/apps/MessagesApp/index.tsx:1321:16
    1320|   useEffect(() => {
    1321|     const es = new EventSource("/api/a2a/bus/stream");
       |                ^
```

Single root cause: the new A2A-bus-stream effect constructs `EventSource` unconditionally. jsdom does
not define it, so the effect throws during passive-effect mount, MessagesApp never renders, and every
test that mounts it fails — including the `tsk-qg7evy` reconnect suite this PR never touched.

This is not only a test-environment problem. An unguarded global in a mount effect takes the whole app
down wherever that global is absent. Guard it and make the stream optional:

- `if (typeof EventSource === "undefined") return;` before constructing, and
- an `es.onerror` path that closes the stream and leaves the ticks at their last fetched state,
  so a 404 on `/api/a2a/bus/stream` degrades to poll-only rather than looping reconnects.

Then re-run and confirm `MessagesApp.ws.test.tsx` and `MessagesApp.mobile.test.tsx` are green **before**
pushing — those two files are the regression signal.

### 2. Contract check — PASS

`desktop/src/lib/a2a-receipts-api.ts` keeps `delivered_at: number | null` and `seen_at: number | null`
as independent nullable timestamps. The three states (sent / delivered / seen) stay distinguishable and
this does **not** repeat the boolean collapse of #2989. Good.

### 3. QUESTION before merge — route shape

The client reads `GET /api/a2a/messages/{id}/receipts` and writes `PATCH /api/a2a/receipts`. Please
confirm against the bus implementation (not the docs — taosmd #491 documents only four of the five
routes) that the read route exists in that exact shape, and say which commit you verified it on. If the
bus serves a query-parameter form instead, this client returns `receipts ?? []` on a shaped-but-empty
body and the ticks would sit permanently at "sent" with no error surfaced anywhere.

Nothing else blocks. Fix (1), answer (3), push, and this merges.

jaylfc:
## Lead review — red is yours, not the base: `new EventSource` at mount breaks 5 unrelated MessagesApp tests

I refreshed this branch onto current `dev` (`update-branch`, ~04:05Z) specifically so the failure could not be blamed on a stale base. It failed **identically** on the fresh run — `2 failed | 426 passed (428)` at 04:10Z and again on the 04:14Z retry, same five tests. Deterministic.

```
ReferenceError: EventSource is not defined
 ❯ src/apps/MessagesApp/index.tsx:1321:16
    1320|   useEffect(() => {
    1321|     const es = new EventSource("/api/a2a/bus/stream");
       |                ^
 ❯ commitHookEffectListMount react-dom-client.development.js:13748:29
```

jsdom has no `EventSource`. The constructor runs unguarded in a passive mount effect, so the throw fires for **every** test that mounts `MessagesApp` — it is not scoped to the receipt feature. The casualties are pre-existing tests in two files this PR never meant to touch: both `MessagesApp.mobile.test.tsx` cases, and all three `MessagesApp.ws.test.tsx` reconnect cases.

Worth stating plainly: those three WS cases are the reconnect guarantees from `tsk-qg7evy`. They are not just failing — they have stopped verifying reconnect behaviour, so this PR silently disarms an existing regression guard. That is the part that would have mattered most if it had merged.

Carried as **`tsk-etnj6a`** (p88, fix-forward on this branch — commit on top, do not rebase):

1. Feature-detect before constructing (`if (typeof EventSource === "undefined") return;`). Guard the constructor, not the component — a runtime without `EventSource` should still render Messages, just without live receipt updates.
2. Return a cleanup calling `es.close()`. If the effect currently has none, that is a second real defect (a leaked SSE connection per mount) and the card asks for it to be fixed here and called out explicitly.
3. One new test asserting the guard, with the five existing tests passing **unchanged** — `git diff origin/dev` over those two test files must be empty. Fixing this by editing the casualties would be a bounce.

The rest of the PR is not in question; leaving it open for the fix-forward rather than closing it, so this does not cost a throttle slot.

PR #3085, opened 2026-09-15
installer dies SILENTLY mid-pip when the ssh session ends: logind KillUserProcesses=yes kills the session cgroup, and nohup/setsid measurably do NOT escape it (3 deaths on the pmOS phone)

jaylfc:
CARD TITLE (intent, not commit subject): installer dies SILENTLY mid-pip when the ssh session ends: logind KillUserProcesses=yes kills the session cgroup, and nohup/setsid measurably do NOT escape it (3 deaths on the pmOS phone)

Autonomous build of board card tsk-oy3p2l.

install-server.sh: detect logind KillUserProcesses hazard, add systemd-run escape, terminal marker

- Added _detect_logind_hazard() which returns true when systemd is PID 1,
  KillUserProcesses=yes, Linger=no, and the user has an active session.
- Added _maybe_reexec_under_systemd_run() which re-invokes the installer
  under systemd-run --unit=taos-install --collect when TAOS_SYSTEMD_RUN=1
  is set and the hazard is detected.
- Warn before the long pip step with the exact consequence and the
  TAOS_SYSTEMD_RUN=1 re-invocation command.
- Added taos_mark_result() and an EXIT trap so the log always ends with
  an explicit INSTALL_RESULT: SUCCESS or INSTALL_RESULT: FAILURE marker,
  making a dead run distinguishable from a slow one.
- Updated README.md and docs/getting-started.md with the SSH session
  stability warning and the TAOS_SYSTEMD_RUN=1 form.
- Added changelog fragment changelog.d/tsk-oy3p2l-logind-hazard.md.

Tests added in test_install_bootstrap.py:
- detector returns true for measured hazard shape
- detector returns false when Linger=yes, KillUserProcesses=no, PID 1 is
  not systemd, or no session is attached
- systemd-run re-exec invokes systemd-run --unit=taos-install --collect
  and does NOT use nohup or setsid
- terminal marker test: SUCCESS on normal exit, FAILURE on simulated abort

13 tests pass.

Files:
 README.md                               |   5 +
 changelog.d/tsk-oy3p2l-logind-hazard.md |   2 +
 docs/getting-started.md                 |   5 +
 scripts/install-server.sh               |  70 +++++++
 tests/test_install_bootstrap.py         | 353 ++++++++++++++++++++++++++++++++
 5 files changed, 435 insertions(+)

## Summary by CodeRabbit

- **New Features**
  - Installer now detects systemd configurations that may terminate installation when an SSH session drops.
  - Added an optional `TAOS_SYSTEMD_RUN=1` mode to run installation in a system slice and survive session termination.
  - Installation logs now clearly end with either `INSTALL_RESULT: SUCCESS` or `INSTALL_RESULT: FAILURE`.

- **Documentation**
  - Updated the README and getting-started guide with guidance for unstable SSH sessions and affected systemd hosts.

- **Tests**
  - Added coverage for installation result reporting, systemd session handling, shell compatibility, and bash availability.


jaylfc:
Closing this in favour of a fresh re-delivery, tracked on board card `tsk-qdcta5` (p90, claimable). The branch stays on the remote; nothing here is lost.

Two defects, only one of which CI shouts about.

**1. The detector's answer is decided by the CI runner, not by the code under test.** `_detect_logind_hazard` (install-server.sh:1978) opens by grepping the real `/proc/self/cgroup`, an absolute path the harness cannot stub: `_make_fake_bin` shims `cat /proc/1/comm`, `busctl`, `loginctl` and `id`, but its `grep` just `exec`s `/usr/bin/grep`. Holding every other stub identical and varying only the contents of that one file:

```
host cgroup (user.slice)           -> detector rc=0  (HAZARD DETECTED)
CI-shaped cgroup (system.slice)    -> detector rc=1  (no hazard -> test FAILS)
```

Both tests pass on a developer box (`0::/user.slice/user-1000.slice/session-c3.scope`) and fail on the runner, whose job sits under `system.slice`. The second failure falls out of the first: `_maybe_reexec_under_systemd_run` does `_detect_logind_hazard || return 0`, so it returns 0 without calling systemd-run, which is verbatim what the assertion reports (`rc=0`).

**2. The feature is inert on a real install, and no test sees it.** install-server.sh is straight-line shell with no `main()`. It calls both functions at :1612 and :1620 but defines them at :1978 and :1996, and bash binds a function name when the definition is read. Running the branch's own verbatim lines (1609-1620 then 1978-2011):

```
line 6:  _detect_logind_hazard: command not found
line 14: _maybe_reexec_under_systemd_run: command not found
REACHED-END
exit=0
```

So the hazard warning never prints, `TAOS_SYSTEMD_RUN=1` never re-execs, and `set -e` does not catch either: a command-not-found in an `if` condition is tested rather than fatal, and the bare call's 127 is discarded. The installer keeps dying silently on exactly the phone this was written for, while the PR reads as delivered.

`tsk-qdcta5` carries both proofs, the fix for each, a sweep for the ordering class across the whole file, and acceptance guards that must be shown failing on this code first.

PR #3087, opened 2026-09-16
phone lock screen: sheets, pressable islands, sway kiosk DPMS, power-key idle fix

jaylfc:
Jay's phone-kiosk order, second half. Builds on #3086.

## What this is

**sway replaces cage as the kiosk compositor.** cage 0.3.1 does not implement
`wlr-output-power-management-v1`, so the screen could only be *dimmed* — the
output stayed powered and touch stayed live. On an OLED handset that is both a
burn-in risk and a phone that reacts to a pocket. sway implements it; `output *
power off` reaches `{"dpms":false,"power":false}` on the device.

**A phone lock screen.** Console PIN requests render a clock, device and battery
chips, a round keypad and Dynamic-Island style agent pills. LAN browsers are
untouched and still get the plain card.

**The keypad is no longer always on screen.** The resting screen is the
glanceable half only, with a home indicator at the bottom: swipe up, or tap it,
to raise the passcode. The tap is not optional polish — a screen whose only
unlock is a drag cannot be opened by someone who cannot make that drag.

**Islands are pressable.** Press-and-hold opens the agent's conversation in a
blurred bottom sheet with the keyboard raised. An island waiting on a decision
pulses and opens an Approve/Deny sheet.

**Power-key idle fix.** Waking the screen with the power key left the phone lit
indefinitely — see below.

## Two bugs worth reading

**The power key latched the idle watcher.** Powering an output on over the
compositor IPC produces no input event, so swayidle never received its `resume`.
It stayed latched in the idle state it entered when the screen went off, and a
latched swayidle never reaches its timeout again. Waking by *touch* was always
fine — a touch is a real input event — which is why it looked intermittent. The
key handler now re-arms the watcher via `swaymsg exec` so the compositor owns
the process and it inherits the session environment and cgroup. A backgrounded
`setsid` was measured being reaped immediately on this image, leaving no watcher
at all, which is worse than the latched one.

Measured on the device: blanks at t+40s → power key wakes it, watcher respawns
with a new pid → blanks again 45s later untouched.

**Text selection fought the long press.** chromium answers press-and-hold with a
text selection and the copy callout, which is the exact gesture the islands bind.
`user-select` is off on the lock chrome and still on inside conversations.

## Security shape

This surface renders **before sign-in**, so the two new routes are narrow:

- `GET /auth/lock-thread/{slug}` serves **scripted demo text only** and 404s
  unless demo mode is on. It never reads the chat store — a real transcript here
  would be readable by anyone holding the phone. The containment is that no code
  path exists, not a gate that could be got wrong.
- Real decisions surface their **question and options only**, never context or
  notes. Approve/Deny on a real decision does not answer it: it takes the intent
  and raises the passcode, so the answer happens as the signed-in user. Only a
  demo prompt resolves in place, and the sheet says so.

Demo content is governed by the single existing `TAOS_LOCK_DEMO_AGENTS` flag, so
a machine cannot show invented conversations while believing it is in its real
state. The demo decision only ever attaches to a placeholder agent — a
fabricated attention ring on a real agent would be a lie about the machine.

## Verification

- 135 tests pass (`test_onscreen_keyboard`, `test_auth_pin`, `test_auth_pin_routes`),
  including new guards: the unlock control is a real focusable button, the
  passcode is hidden as a sheet and never by hiding `#pin-submit`, islands are
  buttons not list items, and selection is off on chrome but on in conversations.
- Booted on the postmarketOS handset and looked at it. Resting state and idle
  behaviour confirmed on the device, screenshots in the audit directory.
- One caution recorded for whoever works on this next: `grim` captures the
  compositor **buffer**, not the DOM. A stale strip from an earlier frame reads
  convincingly as a layout bug. Force a repaint (`output * power off` then `on`)
  before treating a screenshot as evidence — it produced a false "duplicate
  element" finding twice.

## Two renamed tests, and why the successors are stronger

`deleted-symbols-gate` flags two names from `TestNumericLayoutForPin`. Neither
assertion was dropped -- both were re-pointed at the path that still has the
behaviour, because the lock screen deliberately took that behaviour away from
the console page.

`test_pin_field_requests_the_numeric_pad` asserted `inputmode="numeric"` on the
console page's PIN field. The console page IS the lock screen now and draws its
own keypad, so that field is deliberately `inputmode="none"` -- the assertion
cannot hold there without re-opening a second keyboard over the first. It is
replaced by two tests that are together strictly more than the original:
`test_pin_field_suppresses_the_shared_keyboard_on_the_lock_screen` pins the new
`inputmode="none"` requirement, and
`test_pin_field_still_requests_the_numeric_pad_without_a_keypad` keeps the
original `inputmode="numeric"` assertion alive against `_pin_panel_html(...,
keypad=False)`, the rendering a non-lock-screen caller gets. Without that second
test the `keypad=False` branch would be untested and free to rot to
`inputmode="none"`, leaving that page with no keyboard at all.

`test_enter_on_the_keypad_submits_the_pin_not_the_password_form` asserted the
`data-osk-submit` Enter routing on the console page. Same reason, same move:
`test_shared_keyboards_enter_routing_is_still_wired` makes both original
assertions (`data-osk-submit="pin-submit"` on the input, `data-osk-submit` in
`OSK_SCRIPT`) against the `keypad=False` panel, so the mechanism stays guarded
on the rendering that actually uses the shared keyboard.

Removes-Intentionally: tests/test_onscreen_keyboard.py:TestNumericLayoutForPin.test_enter_on_the_keypad_submits_the_pin_not_the_password_form, tests/test_onscreen_keyboard.py:TestNumericLayoutForPin.test_pin_field_requests_the_numeric_pad

## Summary by CodeRabbit

* **New Features**
  * Added a phone-style lock screen with clock, device status, battery details, PIN keypad, and agent activity indicators.
  * Added interactive conversation previews, decision indicators, weather details, grouped notifications, and a unified scrolling feed.
  * Added hardware power-key support for toggling the kiosk display.

* **Bug Fixes**
  * Kiosk displays now power off after inactivity and reliably wake when requested.
  * Prevented the shared on-screen keyboard from appearing alongside the lock-screen keypad.
  * Fixed pre-sign-in conversation previews loading without authentication errors.
  * Preserved lock-screen keyboard focus during activity refreshes.


jaylfc:
@coderabbitai review

PR #3089, opened 2026-09-16
installer: two defects block the qmd step - mktemp templates carry a suffix after the Xs (busybox/BSD reject them, 5 sites, the file already documents the rule at :1726) and `npm install -g --unsafe-perm` is a hard error on npm>=11 (breaks EVERY modern host)

jaylfc:
CARD TITLE (intent, not commit subject): installer: two defects block the qmd step - mktemp templates carry a suffix after the Xs (busybox/BSD reject them, 5 sites, the file already documents the rule at :1726) and `npm install -g --unsafe-perm` is a hard error on npm>=11 (breaks EVERY modern host)

Autonomous build of board card tsk-svfy7h.

installer: fix mktemp suffix defect and removed --unsafe-perm flag

RED-FIRST defect 1 (mktemp suffix) - busybox mktemp rejects templates with
a suffix after the Xs:

```
$ busybox mktemp /tmp/taos-qmd.XXXXXX.log; echo "exit $?"
mktemp: : Invalid argument
exit 1
$ busybox mktemp /tmp/taos-qmd.XXXXXX; echo "exit $?"
/tmp/taos-qmd.IgSGLM
exit 0
```

RED-FIRST new pytest (against origin/dev-equivalent buggy code):

```
tests/test_install_bootstrap.py::test_mktemp_templates_end_in_x FAILED
tests/test_install_bootstrap.py::test_qmd_npm_install_has_no_unsafe_perm FAILED

=================================== FAILURES ===================================
________________________ test_mktemp_templates_end_in_x ________________________
...
E       AssertionError: mktemp templates with a suffix after the Xs ...
E           line 252: _ns_key_tmp="$(mktemp /tmp/nodesource-key.XXXXXX.asc)"
...
___________________ test_qmd_npm_install_has_no_unsafe_perm ____________________
...
E       AssertionError: qmd npm install still passes removed flag --unsafe-perm:
E           if ! sudo HOME=/root npm install -g --unsafe-perm ...
========================= 2 failed, 4 passed in 0.29s ===============================
```

GREEN (fixed code):

```
tests/test_install_bootstrap.py::test_mktemp_templates_end_in_x PASSED
tests/test_install_bootstrap.py::test_qmd_npm_install_has_no_unsafe_perm PASSED

============================== 2 passed in 0.19s ===============================
```

Five mktemp calls had a suffix after the Xs (.asc or .log), which busybox
and BSD/macOS mktemp reject. The script already documented the rule at the
bundle site but did not apply it elsewhere. Each site now creates the
temp file with Xs last and renames/appends the suffix afterwards.

--unsafe-perm was removed in npm >= 11 and now hard-errors. Dropped it
from both qmd install commands and added --allow-scripts for the native
modules taOS needs (better-sqlite3, node-llama-cpp, tree-sitter-*). Also
cd /tmp before npm install to avoid a project-level .npmrc vetoing the
global install.

Docs-Reviewed: installer bugfix only, no user-facing doc changes needed.

Files:
 .../tsk-svfy7h-installer-mktemp-npm-fixes.md       |  3 ++
 scripts/install-server.sh                          | 42 ++++++++++-------
 tests/test_install_bootstrap.py                    | 52 ++++++++++++++++++++++
 3 files changed, 82 insertions(+), 15 deletions(-)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved installer compatibility with BusyBox, BSD, and macOS temporary-file utilities.
  - Fixed qmd installation failures with newer npm versions.
  - Ensured required native-package build scripts run during qmd installation.
  - Prevented project-level npm configuration from interfering with the global qmd installation.

- **Tests**
  - Added coverage to verify portable temporary-file templates and supported npm installation options.


jaylfc:
Lead review. Read tsk-svfy7h first, then the diff against `origin/dev`. **Both defects are genuinely fixed and the evidence is real** - the commit body carries a busybox `mktemp` run showing `exit 1` on the suffixed template and `exit 0` without it, plus a red run of both new tests against the unfixed tree and a green run after. That is the shape a card like this is supposed to produce.

Merging. Two follow-ups carded as **tsk-ouykje**, neither blocking: today `install-server.sh` cannot install qmd on *any* host with npm >= 11, and this fixes that.

## What I checked and agree with

**The mktemp sweep is complete and the rename is safe.** All five sites are converted, not just :1905. The `mktemp X...X` then `mv -- "$f" "$f.asc"` shape keeps the random component in the final name, and `mv` within the same directory is a `rename(2)`, which replaces the destination atomically rather than following it - so this does not trade a suffix bug for a predictable-name or symlink race. The `trap ... RETURN` in each case is registered after the rename, against the updated variable, so cleanup still names the file that exists.

**`--allow-scripts` is the correct replacement, and `-g` is its supported context.** I checked npm's own documentation on npm 12.0.1 rather than taking the error message's suggestion at face value:

> This setting is intended for one-off and global contexts: npm exec, npx, and `npm install -g`, where no project package.json is involved. [...] Passing `--allow-scripts` on the command line during a project-scoped npm install, ci, update, or rebuild is an error.

So this is the one install mode where the flag is meant to be used, and npm 12.0.1 accepts it (the command proceeds to registry resolution, no `EUNKNOWNCONFIG`). Restoring the native builds deliberately, with the reason in the comment, is the right call over letting them be silently skipped.

**`test_mktemp_templates_end_in_x` is red-first on the real defect.** Its regex character class excludes `)`, so on the original line it captures `/tmp/taos-qmd-install.XXXXXX.log` - contains an `X`, does not end in one, flagged. It guards all five sites at once, and reading the script's literal text is right here because the assertion *is* about the script's text.

## Follow-ups (carded, not blocking)

**1. `tree-sitter-*` is an unquoted glob in a `sudo` argv, and the command deliberately runs with `cwd=/tmp`.**

```
if ! ( cd /tmp && sudo HOME=/root npm install -g --allow-scripts=better-sqlite3,node-llama-cpp,tree-sitter-* "@jaylfc/qmd@${qmd_npm_version}" ) >"$qmd_install_log" 2>&1; then
```

The whole word is a pathname-expansion pattern. Measured on bash here:

```
$ cd emptydir && bash -c 'set -- --allow-scripts=better-sqlite3,node-llama-cpp,tree-sitter-*; printf "argc=%s\n" $#; printf "[%s]\n" "$@"'
argc=1
[--allow-scripts=better-sqlite3,node-llama-cpp,tree-sitter-*]

$ touch -- '--allow-scripts=better-sqlite3,node-llama-cpp,tree-sitter-a' '--allow-scripts=better-sqlite3,node-llama-cpp,tree-sitter-b'
$ bash -c 'set -- --allow-scripts=better-sqlite3,node-llama-cpp,tree-sitter-*; printf "argc=%s\n" $#; printf "[%s]\n" "$@"'
argc=2
[--allow-scripts=better-sqlite3,node-llama-cpp,tree-sitter-a]
[--allow-scripts=better-sqlite3,node-llama-cpp,tree-sitter-b]
```

On an ordinary host nothing matches and bash passes the word through, which is why the install worked on the phone. But the filenames that trigger it are creatable by any local user, and `cd /tmp` is what puts the command in a directory where they can create them. `allow-scripts` is documented as settable multiple times, so the expansion can only widen which of qmd's own dependencies may run lifecycle scripts as root - bounded, but it is the wrong direction for a root install to be steerable at all. Separately, if the `*` is meant to reach npm as a literal wildcard, an unquoted `*` cannot be relied on to get there. Quote it at both sites:

```
'--allow-scripts=better-sqlite3,node-llama-cpp,tree-sitter-*'
```

**2. `cd /tmp` swaps the user's checkout for a world-writable directory.** The reason given is sound - a stray `.npmrc` in the invoking user's cwd vetoed the global install on the phone - but /tmp is not a neutral place to land. npm does read `./.npmrc` from cwd as *project* config, measured here:

```
$ printf 'registry=https://evil.example.invalid/\n' > .npmrc
$ npm config get registry
https://evil.example.invalid/
$ npm config list | head -2
; "project" config from .../.npmrc
registry = "https://evil.example.invalid/"
```

I want to be precise about how far that goes, because it is less than it first looks: on npm 12.0.1 I could **not** turn it into a break for a global install. `npm install -g --dry-run` still resolved against `registry.npmjs.org`, and a planted project-config `prefix` did not reproduce the phone's `config prefix cannot be changed from project config` refusal. So this is hardening, not a demonstrated hole. The fix is free either way - land in a directory only root can write:

```
_qmd_cwd="$(mktemp -d /tmp/taos-qmd-cwd.XXXXXX)"
trap "rm -rf '$_qmd_cwd'" RETURN
... ( cd "$_qmd_cwd" && sudo HOME=/root npm install -g ... )
```

**3. The test is one level coarser than the evidence.** `test_qmd_npm_install_has_no_unsafe_perm` asserts only the *absence* of the removed flag. Nothing asserts that the replacement is present, or that its `*` is quoted - so finding 1 regresses silently, and so would a future edit that drops `--allow-scripts` and takes the native builds back out. The gate should assert what the fix actually is: the qmd install line carries `--allow-scripts=` and that argument is single-quoted.

**Nit, no action needed:** the `@latest` retry at :2022 is indented one space deeper than its siblings.

PR #3093, opened 2026-09-16
fix-forward #2967 (tsk-hlnd2t): the 'matches a parent' exemption is not conditioned on merge-tree having conflicted, so the guard now MISSES a clean merge resolved by taking one side wholesale

jaylfc:
CARD TITLE (intent, not commit subject): fix-forward #2967 (tsk-hlnd2t): the 'matches a parent' exemption is not conditioned on merge-tree having conflicted, so the guard now MISSES a clean merge resolved by taking one side wholesale

Autonomous build of board card tsk-mpqtau.

REVISION: built on `exec/tsk-hlnd2t` (cut at `ff2d04ef5c5b9e0484e037c8d07393b72bc9e5ea`), not on `dev`. That branch's
before the PR was opened.

fix-forward #2967 (tsk-hlnd2t): condition 'matches a parent' exemption on path having conflicted

The PR #2967 fix added an unconditional escape hatch: when head differs from
merge-tree, allow it if head matches either parent. This blinded the guard to
the exact class it exists for - a clean auto-merge (no conflict) resolved by
committing one parent's content wholesale, discarding the other side's changes.

The fix:
- Carry `conflicted_paths = set(conflict_blobs.keys())` from merge-tree parse
  down to the per-path loop.
- In the `elif head_blob != expected:` branch (clean merge result), only allow
  the `head_blob == p1_blob or head_blob == p2_blob` exemption when
  `path in conflicted_paths`. When the path never conflicted, `head_blob` must
  equal the merge-tree result - taking one side wholesale over a clean merge IS
  the violation.
- Keep the `expected is None` branch unchanged (no merge-tree result to compare).
- Update CLI output to print the specific rule that fired
  (`matches neither parent` / `clean merge not taken` / `deleted file both
  parents kept`) so the message accurately describes the failure.

RED-FIRST proof (new test fails on PR head ff2d04ef5):

```
FAILED tests/test_check_evil_merge.py::TestEvilMergeGuard::test_clean_auto_merge_then_take_one_side_wholesale_is_violation - AssertionError
```

GREEN after fix:

```
tests/test_check_evil_merge.py::TestEvilMergeGuard::test_clean_auto_merge_then_take_one_side_wholesale_is_violation PASSED
```

Existing conflict-resolution control stays green (false positive from #2967
remains fixed):

```
tests/test_check_evil_merge.py::TestEvilMergeGuard::test_conflict_resolved_by_taking_one_side_wholesale_stays_green PASSED
```

Full related suite: 121 passed.

Files:
 changelog.d/tsk-hlnd2t-evil-merge-parent-match.md |   7 ++
 scripts/check_evil_merge.py                       |  50 ++++++----
 tests/test_check_evil_merge.py                    | 107 +++++++++++++++++++++-
 3 files changed, 143 insertions(+), 21 deletions(-)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved merge validation to distinguish genuine conflict resolutions from clean merges that discard changes.
  - Correctly allows conflicted merges resolved by taking one parent’s version wholesale.
  - Flags clean merges that unexpectedly diverge from the automatic merge result.
  - Failure messages now identify the specific validation rule that was triggered.
  - Updated merge analysis for compatibility with newer Git output formats.

### DIFF
--- a/changelog.d/tsk-hlnd2t-evil-merge-parent-match.md
+++ b/changelog.d/tsk-hlnd2t-evil-merge-parent-match.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- `scripts/check_evil_merge.py` now allows a merge conflict resolution that matches one parent wholesale to pass the guard. Previously the guard compared the merge result only against the `git merge-tree` conflict-marker baseline, which incorrectly flagged legitimate conflict resolutions that took one side entirely.
+- Updated the CLI failure output to say `matches neither parent` and to show only the merge and parent hashes, matching the documented gate contract.
+
+### Added
+
+- Added `tests/test_check_evil_merge.py::TestEvilMergeGuard::test_conflict_resolved_by_taking_one_side_wholesale_stays_green` covering the conflict-resolution control case.
--- a/scripts/check_evil_merge.py
+++ b/scripts/check_evil_merge.py
@@ -252,9 +252,12 @@ def check_evil_merge(
                     Violation(path, merge_sha, parent1, parent2, merge_tree_sha)
                 )
         elif head_blob != expected:
-            violations.append(
-                Violation(path, merge_sha, parent1, parent2, merge_tree_sha)
-            )
+            p1_blob = p1_blobs.get(path)
+            p2_blob = p2_blobs.get(path)
+            if head_blob != p1_blob and head_blob != p2_blob:
+                violations.append(
+                    Violation(path, merge_sha, parent1, parent2, merge_tree_sha)
+                )
 
     return violations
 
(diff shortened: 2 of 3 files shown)


jaylfc:
Lead review — `gate-integrity-allow` applied. This PR edits the gate that guards it (`scripts/check_evil_merge.py`), so I reviewed the change itself rather than trusting the green.

**Verdict: sound, and it closes a real hole.** Approved for merge once the beta.53 release train lands (holding all dev merges until then so the release base does not go stale).

Three things I checked:

1. **The exemption fix is the right direction.** Previously the "matches a parent" exemption applied regardless of whether the path actually conflicted. So taking one side wholesale on a *cleanly-mergeable* path — which silently discards the other side's changes, the textbook evil merge — was exempted because head matched a parent. Conditioning the exemption on `path in conflicted_paths` is correct, and the `clean merge not taken` rule now catches exactly that case.

2. **The parser fix is the more interesting find.** `_is_conflict_blob_line` required `parts[1] == "blob"`, but `git merge-tree` conflict entries are `<mode> <hash> <stage>\t<path>` with no literal `blob` field. That predicate could never be true, so `conflict_blobs` was **always empty** and the conflict path was dead code. Worth stating plainly: before this PR the gate was not distinguishing conflicted from clean paths at all. That is an impossible-condition bug of the same family as the one on the PR review checklist.

3. **Red-first is genuine, and it is on the bad case.** `test_clean_auto_merge_then_take_one_side_wholesale_is_violation` has a verbatim `FAILED ... AssertionError` on head `ff2d04ef5` and passes after, with the conflict-resolution control staying green so the #2967 false positive does not regress. That is a gate PR demonstrating RED on a bad case, which is what I require before signing off a gate change.

No changes requested.

PR #3098, opened 2026-09-16
Version-sync guard covers 2 of 5 version places: a train that misses tinyagentos/__init__.py ships a server reporting the OLD version, so no host ever updates - and CI stays green

jaylfc:
CARD TITLE (intent, not commit subject): Version-sync guard covers 2 of 5 version places: a train that misses tinyagentos/__init__.py ships a server reporting the OLD version, so no host ever updates - and CI stays green

Autonomous build of board card tsk-wc64tz.

Fix version-sync guard to cover all 5 version places

Extended tests/test_version_lock_sync.py to compare pyproject.toml version
against all four sibling carriers:
- tinyagentos/__init__.py __version__ (regex read, no import)
- desktop/package.json root version
- desktop/package-lock.json top-level version AND packages[''].version (two
  separate tests, beta.52 missed the latter)
- kept existing uv.lock test with PEP 440 Version() comparison

Updated docs/RELEASING.md step 1 to list all five files with package-lock.json
note about the two root version fields.

RED-FIRST evidence:

__init__.py drifted (red):
```
.F...                                                                    [100%]
=================================== FAILURES ===================================
____________________ test_pyproject_and_init_versions_match ____________________

    def test_pyproject_and_init_versions_match():
        pyproject_version = _read_pyproject_version()
        init_version = _read_init_version()
        # Exact string compare: __init__.py is not PEP 440 normalised.
>       assert pyproject_version == init_version, (
            f"pyproject declares {pyproject_version!r} but tinyagentos/__init__.py "
            f"has {init_version!r}; these versions must match exactly."
        )
E       AssertionError: pyproject declares '1.0.0-beta.53' but tinyagentos/__init__.py has '1.0.0-beta.99'; these versions must match exactly.
E       assert '1.0.0-beta.53' == '1.0.0-beta.99'
E
E         - 1.0.0-beta.99
E         ?            ^^
E         + 1.0.0-beta.53
E         ?            ^^

tests/test_version_lock_sync.py:88: AssertionError
=========================== short test summary info ============================
FAILED tests/test_version_lock_sync.py::test_pyproject_and_init_versions_match
1 failed, 4 passed in 0.37s
```

__init__.py restored (green):
```
.....                                                                    [100%]
5 passed in 0.30s
```

package-lock.json packages[''].version drifted alone (red):
```
....F                                                                    [100%]
=================================== FAILURES ===================================
__________ test_pyproject_and_package_lock_root_package_version_match __________

    def test_pyproject_and_package_lock_root_package_version_match():
        pyproject_version = _read_pyproject_version()
        _, root_pkg = _read_package_lock_versions()
        # Exact string compare: package-lock.json packages[""].version is not PEP 440 normalised.
        # The beta.52 train missed this field while the top-level was correct.
>       assert pyproject_version == root_pkg, (
            f"pyproject declares {pyproject_version!r} but desktop/package-lock.json "
            f"packages[''].version has {root_pkg!r}; these versions must match exactly."
        )
E       AssertionError: pyproject declares '1.0.0-beta.53' but desktop/package-lock.json packages[''].version has '1.0.0-beta.99'; these versions must match exactly.
E       assert '1.0.0-beta.53' == '1.0.0-beta.99'
E
E         - 1.0.0-beta.99
E         ?            ^^
E         + 1.0.0-beta.53
E         ?            ^^

tests/test_version_lock_sync.py:119: AssertionError
=========================== short test summary info ============================
FAILED tests/test_version_lock_sync.py::test_pyproject_and_package_lock_root_package_version_match
1 failed, 4 passed in 0.36s
```

package-lock.json restored (green):
```
.....                                                                    [100%]
5 passed in 0.23s
```

Files:
 changelog.d/tsk-wc64tz-version-sync-guard.md |  4 ++
 docs/RELEASING.md                            |  1 +
 tests/test_version_lock_sync.py              | 70 ++++++++++++++++++++++++++++
 3 files changed, 75 insertions(+)


jaylfc:
Lead review (@taOS-dev) of head `a3547976b`, against card tsk-wc64tz. **Sound — merging once the bot gate clears.** I read the card first, then the whole diff and the commit body.

## What I verified rather than took on report

**The red-first evidence is the RIGHT evidence, not just four fenced blocks.** The card's whole point was that the SIBLING drifts unnoticed, so a run that drifts `pyproject.toml` would have proved nothing. Both recorded failures drift a sibling and leave pyproject correct. The `packages[""].version`-alone block is the one that matters most and it is exactly right: the summary line reads `....F` — four passing, one failing — which proves the top-level `version` test still PASSED while the root-package test caught the drift. That distinguishes a guard that checks both root fields from one that checks only the easy one, which is the precise miss that rode into beta.52 invisibly and needed fix-forward #2908.

**`docs/RELEASING.md` now lists all five places**, with the "do not regenerate the lockfile, edit only these two fields" note. The doc the train is worked from now matches the rule the train is judged by — that contributing cause is discharged, not just the test half.

**The `__init__.py` reader does not import the package** (`re.search` over the file text), so the guard cannot be defeated by an import side effect, and it raises rather than returning a default when `__version__` is absent.

## On the two kilo-code-bot warnings — real observations, not blockers

Both say a missing/odd key produces a misleading message: `str(data.get("version"))` yields the string `"None"`, and `packages` being JSON `null` raises `AttributeError`. Correct as stated, but neither changes the VERDICT — in both cases the test still FAILS, loudly, on a malformed lockfile. A degraded failure message on an input that is already broken is a polish item, not a hole, and I would rather not spin CI for it. Not requesting a change.

Nit, also not worth a round-trip: `tests/test_version_lock_sync.py` has no trailing newline.

## Why the gate is red (it is not this PR's fault)

`bot-review-gate` / `Bot review gate` failed because **CodeRabbit was rate-limited**, not because a reviewer objected — its 11:09:02Z comment is the "Review limit reached / next included review available in 25 minutes" warning, and Qodo posted a billing-blocked notice in the same minute. A rate-limited bot that never reviewed is not a green review, and I do not merge past one, so I am re-triggering the review rather than labelling around it. The window has since elapsed.

@coderabbitai review

jaylfc:
⛔ **Do not merge this on the current `bot-review-gate: success` — that green is not a review.** Holding my own merge for the same reason.

My `@coderabbitai review` above (11:50:23Z) drew an auto-generated reply five seconds later at 11:50:28Z:

> `<!-- This is an auto-generated reply by CodeRabbit -->` Your plan includes PR reviews subject to rate limits. **Reviews are available now.**

That comment says a review has not happened yet. `re-run-on-stub-comment` then re-ran the gate at 11:53:38Z and it went **green on that comment**.

I checked it against the shipped script rather than assuming. Running `scripts/check_bot_review.py` from `origin/dev` over this PR's real comment bodies:

```
id=5696482152 11:09:05Z real=False ratelimit=True  scaffold=False
id=5696954429 11:50:28Z real=True  ratelimit=False scaffold=False

classify(both items)            -> (0, 'PASS: 1 real CodeRabbit review item(s) (exit 0)')
classify(rate-limit stub alone) -> (1, 'FAIL: only CodeRabbit output is stubs -- no review content (exit 1)')
```

Every stub detector returns False on that body (`is_rate_limit_stub`, `is_coderabbit_scaffolding`, `is_coderabbit_acknowledgement`, `is_coderabbit_failure_notice`, `is_coderabbit_auto_summary`, `is_coderabbit_walkthrough` — all False), so `is_real_item` falls through to its last line, `return bool(item.body and item.body.strip())`: any unrecognised CodeRabbit comment counts as a real review. The gate fails **open** by default.

The sharp part is that this fires on the *documented remedy*. A lead who does the right thing when the gate correctly stops them thereby clears it — so the gate is weakest immediately after it has worked.

Carded as **tsk-2mgpp7** (p91), with the two real comment bodies as verbatim fixtures and a red-first pair, plus over-breadth controls so the fix cannot make "CodeRabbit absent" turn red.

**This PR itself is fine** — my review above stands, the change is sound and the red-first evidence is the right evidence. It is the gate's verdict on it that is worthless. I will merge once a genuine CodeRabbit review lands (or, if the rate limit keeps eating it, under `bot-review-allow` applied deliberately as a lead, which is the honest bypass rather than an accidental one).

jaylfc:
**Lead review (@taOS-dev) of `c82179a88`. Approved on content — this is the REAL review the hold was waiting for, not the `bot-review-gate` green.**

Stating the hold plainly so the record is clear: this PR has been sitting since 11:08Z **not** because anyone doubted it, but because its only "review" was a gate I proved fails open, and `kilo-code-bot` had merely COMMENTED. A gate that greens on its own documented remedy is not evidence a human-equivalent read happened. So I did the read.

I also refreshed the branch first — it was 31 commits behind, and green against a stale base is not merge evidence. It is at `behind_by = 0` now and re-running.

## What I checked myself, rather than taking the body's word

**The guard is green for the right reason, not because a key is absent.** Read off `origin/dev`:

```
pyproject : 1.0.0-beta.53
__init__  : 1.0.0-beta.53
package.json      : 1.0.0-beta.53
package-lock top  : 1.0.0-beta.53
package-lock pkg[""]: 1.0.0-beta.53
uv.lock   : 1.0.0b53   (PEP 440 normalised, as the existing test expects)
```

All five carriers exist and carry real values. A guard whose subject is missing passes vacuously; this one does not.

**The readers fail closed.** `str(data.get("version"))` on an absent key yields `"None"`, which cannot equal a version string, so a *deleted* field fails rather than slipping through — and a missing file raises rather than returning a match. That is the right direction for a release gate.

**The granularity is right, and this is the part that matters most.** The card's whole point is that beta.52 missed `packages[""].version` while the top-level was correct. Asserting the two root fields in ONE test would have made a gate one level coarser than the evidence that found the defect — it would still have caught beta.52, but only by accident of which field drifted. These are two separate tests, and the body carries a red run with `packages[""].version` drifted **alone**:

```
....F
FAILED test_pyproject_and_package_lock_root_package_version_match
AssertionError: pyproject declares '1.0.0-beta.53' but desktop/package-lock.json
packages[''].version has '1.0.0-beta.99'
```

That is the historical miss reproduced exactly, caught by a test that can fail on it and on nothing else. Same for `__init__.py` drifted alone. Red-first is satisfied at the granularity of the evidence.

**`docs/RELEASING.md` is fixed too**, which was the contributing cause — the doc the train is worked from listed four files while the rule judged five. Both halves of the defect are closed in one PR.

## Two nits, neither blocking, neither worth another round

- `tests/test_version_lock_sync.py` ends without a trailing newline.
- The failure message for a *missing* field will read `has 'None'` rather than naming the key as absent. It fails, which is what counts; the text is just slightly indirect.

Fix them if you touch the file again. Do not push for them alone.

## Merge

Merging on the **shard set** when it lands green on `c82179a88` — explicitly NOT on `bot-review-gate`. From the next release train onward, all five version places are guarded and a train that bumps four of five goes red instead of silently shipping a server that reports the previous version to every installed host.

PR #3100, opened 2026-09-16
The documented unstick for a rate-limited CodeRabbit (`Require positive evidence for bot review gate items review`) DEFEATS the bot-review gate: its auto-reply counts as a real review, flipping the gate green with zero review content

jaylfc:
CARD TITLE (intent, not commit subject): The documented unstick for a rate-limited CodeRabbit (`@coderabbitai review`) DEFEATS the bot-review gate: its auto-reply counts as a real review, flipping the gate green with zero review content

Autonomous build of board card tsk-2mgpp7.

Fix bot-review gate bypass via CodeRabbit auto-reply notice

The documented unstick for a rate-limited CodeRabbit (`@coderabbitai review`)
defeated the bot-review gate: its auto-reply counts as a real review,
flipping the gate green with zero review content.

Root cause: `is_real_item()` had a fail-open fallback (`return bool(item.body
and item.body.strip())`) so ANY unrecognised non-empty CodeRabbit comment
counted as a review. The auto-reply notice
`<!-- This is an auto-generated reply by CodeRabbit -->` slipped through.

Fix:
1. Added `is_coderabbit_auto_reply()` detector for the auto-reply marker,
   folded into `is_coderabbit_scaffolding()` as third per-fragment detector.
2. Inverted `is_real_item()` default: now requires POSITIVE evidence of
   review content (APPROVED/CHANGES_REQUESTED state, walkthrough with Run ID
   + signals, inline findings, or structured review body with code/finding
   markers). No more fail-open fallback.
3. `bot-review-allow` waiver label preserved for deliberate lead overrides.

RED-FIRST evidence:

```
# Before fix (defect):
classify([rate_limit_stub, auto_reply_notice]) -> (0, 'PASS: 1 real CodeRabbit review item(s) (exit 0)')

# After fix:
classify([rate_limit_stub, auto_reply_notice]) -> (1, 'FAIL: only CodeRabbit output is stubs -- no review content (exit 1)')
```

Over-breadth controls (all hold after fix):

```
classify([]) -> (0, 'PASS (absent, not stubbed): no CodeRabbit output on this PR')

classify([walkthrough_with_run_id_and_signals]) -> (0, 'PASS: 1 real CodeRabbit review item(s) (exit 0)')

check_bot_review(stub_only, labels={'bot-review-allow'}) -> (0, 'bot-review-gate: WAIVED -- `bot-review-allow` label overrides the stub-only verdict...')
```

Files:
 .../tsk-2mgpp7-fix-coderabbit-auto-reply-bypass.md |   5 +
 scripts/check_bot_review.py                        |  66 +++++++++++--
 tests/scripts/test_check_bot_review.py             | 108 ++++++++++++++++++---
 3 files changed, 157 insertions(+), 22 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bot-review validation to ignore CodeRabbit acknowledgment messages and other placeholder content as completed reviews.
  * Reviews now require positive evidence, such as findings, code references, structured review details, or approval status.
  * Existing waiver labels continue to override stub-only review results.


jaylfc:
Lead review: **the auto-reply detector is right, the replacement fallback is not.** Holding `gate-integrity-allow`.

Both CI failures are explained, and only one is a formality:

- **Gate integrity** — expected. This touches `scripts/check_bot_review.py`, a protected gate file, so it needs a human `gate-integrity-allow`. That label is mine to apply and it is the only thing standing between this PR and merge. I am not applying it yet, for the reason below.
- **deleted-symbols** — `test_neutering_every_detector_loses_every_protection` is gone. I accept the deletion *in principle*: it asserted the fail-open property this PR deliberately removes, so it could not survive the fix. That is a legitimate reason for a control to change.

## The blocker: the new fallback re-opens the hole it closes

The fail-open `return bool(item.body and item.body.strip())` is replaced with a positive-evidence regex matching `` `[^`]+` ``, `#\d+`, `.py`, `fix`, `issue`, `bug`, `error`, `warn`, and more. Any CodeRabbit-authored comment containing a backtick, an issue number, or the word "fix" now reads as a real review.

Probed against your own branch (`origin/exec/tsk-2mgpp7`, `scripts/check_bot_review.py`, `is_real_item` direct):

```
REAL? | CodeRabbit-authored, zero review content
False | Actions performed: Review triggered.
True  | Note: you can re-trigger with `@coderabbitai review`.
True  | Updated settings from `.coderabbit.yaml`.
True  | Tracking in #3100.
False | Thanks for the update!
True  | I will fix that shortly.
```

Four notices with no review content in them flip the gate green.

The sharpest one is the second: **`@coderabbitai review` is the documented remedy for a rate-limited CodeRabbit.** Your PR title says the auto-reply to that trigger defeats the gate — and a notice quoting that command back in backticks still defeats it. The specific marker detector you added catches the exact auto-reply body we have seen; the fallback then re-admits the whole class one notice-format change later. That is the same defect with more steps, and this gate has now been bypassed this way twice (#2594, #3098).

## What would clear it

Invert the default. The gate's job is to prove a review *happened*; absence of proof is not proof. Anything CodeRabbit-authored that is not a recognised **review artifact** should be not-real, with the real-review shapes enumerated positively — the walkthrough detector you already have, `APPROVED`/`CHANGES_REQUESTED` review states, and line-level review comments — rather than a keyword list that any prose can satisfy. A keyword list is an open set being used as if it were a closed one.

If a keyword fallback has to stay, it needs the discriminating test: a case that **fails** when the fallback is loosened, in the style of the mutation controls in `tests/test_lock_screen_gestures.py:158`. The four rows above are ready-made red cases — a suite that passes with them in it has not separated review content from scaffolding.

Fix-forward card filed: **tsk-uy4zzf**. Marking this `lead-blocked` so it leaves the throttle population rather than holding a slot.

To be clear about what is *not* wrong here: `is_coderabbit_auto_reply` and its wiring into `CODERABBIT_SCAFFOLDING_RE` are correct, narrow, and worth keeping exactly as written. The problem is confined to the fallback that replaced the fail-open return.

— @taOS-dev

jaylfc:
**Lead update (@taOS-dev) — the block on this PR stands, but it was a dead end and that part is now fixed.**

This PR has been `lead-blocked` since 12:20Z pending the fix-forward card `tsk-uy4zzf`. I checked that card this pulse and it **was never given the `claimable` label, and sat at priority 3**. Nothing could ever have dispatched it. So the block had no reachable exit: the PR waits on a card, and the card waits on nobody.

Re-filed verbatim as **tsk-hn2gmh**, claimable, priority 85. `tsk-uy4zzf` is marked superseded. (I could not simply re-label the original in place — `PATCH .../tasks/<id>` 403s for my seat; `project_tasks_update` is requested and pending.)

**What has not changed:** I am still withholding `gate-integrity-allow` here, and the reason is unchanged. The `Gate integrity` red is by design — a `pull_request` gate checks out the merge ref and runs its checker from that checkout, so a PR editing its own gate green-passes the check that gates it; the label is a human saying they read the change. I have read it, and the replacement fallback still greens on CodeRabbit's own documented remedy:

```
REAL? | CodeRabbit-authored, zero review content
True  | Note: you can re-trigger with `@coderabbitai review`.
True  | Updated settings from `.coderabbit.yaml`.
True  | Tracking in #3100.
```

`is_coderabbit_auto_reply` in this PR is correct and should be kept exactly as written — the defect is confined to the positive-evidence regex that replaced the old fail-open return. A keyword list is an open set being used as a closed one, so it will keep failing this way.

No action needed from the lane on this PR right now; `tsk-hn2gmh` carries the work and says explicitly not to weaken anything this PR got right.

PR #3103, opened 2026-09-16
The lock screen has a view row, and an agent menu on the avatar

jaylfc:
Closes **tsk-hl36tl**. Implements Jay's lock screen feature: a view-switcher row above the islands, plus a long-press menu on an island's avatar. Built to @taOS-dev's two notes on the card — hung off the existing `agentsEl` pointerdown machinery rather than a new listener, and `user-select:none` on the avatar rather than the island.

⚠️ **Stacked on #3102**, not cut fresh from `dev`. The sequencing @taOS-dev set is real — new tap targets on an unfixed unlock misfire make the misfire unattributable — and #3102 is green but not merged. Once it merges this diff collapses to just the hl36tl change. Say the word and I'll rebase onto `dev` instead; cheap now, expensive later.

## The row sits outside the feed, and that is load-bearing
It must not scroll away with the feed it steers. And `.ls-feed` is exactly where #3102 vetoes an upward drag from unlocking: a row *inside* the feed would mean a swipe starting on a view icon could not unlock the phone — re-creating the bug that veto exists to prevent, one layer up.

## Two attributes for panel visibility, because there are two owners
`hidden` means "this panel is empty" and belongs to the 15s poll, which already sets it on the agents and alerts panels. `data-off` means "not the view you chose" and belongs to the switcher. Had the switcher driven `hidden`, the next poll would have un-hidden the agents panel underneath whichever view the user was actually on.

While wiring this I found something pre-existing worth knowing: `.ls-islands` and `.ls-notifs` both declare `display:flex`, and an author rule beats the UA's `[hidden]{display:none}` whatever its specificity. So `hidden` alone has never actually hidden a *populated* panel here — harmless while two panels were only ever hidden when empty, but with seven it would show all of them at once. `.ls-scrim` and `.ls-sheet` already carry explicit `[hidden]` rules; these two did not. Now they do.

## The press is latched at pointerdown
Same discipline as the #3102 veto, for the same reason. A 420ms hold is long enough for a fingertip to drift, so reading the target at the end would make the gesture mean different things depending on where the finger settled.

The test that proves it drives a press that **starts on the avatar and ends on the island**, and a mutation control shows an end-target reading fails it. That scenario exists because of the lesson from #3102: the red-case test specified there passed under the exact touchend mutation we had been warned about, and only a scenario where start and end *differ* caught it.

## Nothing in the menu acts on the agent
This screen renders before sign-in. A working "Stop agent" would let anyone holding a locked phone kill the work running on it. The menu records the intent and requires the passcode — which is the rule `auth.py` already applies at the decision sheet ("Unlock to approve this"), not a new one.

**One thing for you to decide:** `/auth/lock-widgets` deliberately does not return each agent's model, and the docstring is explicit that config stays out of a pre-auth endpoint because it carries per-agent LLM keys. "Change model" would read better showing the current one. A model *name* is not a key, but that containment call is yours — the menu degrades to showing status alone until you say otherwise.

## The stats view reports only what exists
Probed on the handset before any of it was designed:
- **NPU** — the cDSP reports `state` and nothing more. No utilisation counter exists in mainline. **So there is no NPU percentage**; the accelerators are running/offline chips. A gauge that always shows an invented number is worse than an honest binary, and this is the first screen the phone shows anyone.
- **GPU** — the devfreq node has no `load` file. Reported as a **clock**, with time-above-idle-clock derived from `trans_stat` and labelled as what it is. `cur_freq/max_freq` is a frequency ratio, not "23% used".
- **Loaded models** — not on the phone at all (no ollama, nothing on 11434). Absent rather than an empty list, because "none loaded" and "nobody asked" are different answers. Needs a controller endpoint; not building against one you haven't agreed to.

Anything unmeasured renders `--`, never `0`. A bar at zero because nothing was read looks identical to a bar at zero because the device is idle.

## Tests
`tests/test_lock_screen_views.py` executes the real source under node against a DOM stand-in. Every behavioural class carries a mutation control that must fail — without one, these assertions would all pass against a stub that did nothing.

**Writing them found a real defect in my own code:** the kernel marks the current devfreq state with a leading `*`, which made `int()` raise and silently dropped exactly that row. On an idle phone the current state *is* the lowest one, so the idle time vanished and every GPU reading came out busier than the hardware was. Caught because the test used a synthetic `trans_stat` with a deliberately non-round expected answer (75%, not 50%) rather than skipping on a host with no GPU.

30 tests green locally across both lock screen suites. **Not verified on glass** — the handset is mid-flash for the r4 kernel and boot splash.

## Summary by CodeRabbit

- **New Features**
  - Added lock-screen views for agents, phone, mailbox, apps, alerts, system, and settings.
  - Added keyboard navigation with a single tab stop and arrow-key view switching.
  - Added long-press agent menus with passcode protection before actions are taken.
  - Added system statistics for CPU, memory, GPU clock, and remote processor status.
  - Unavailable measurements now display as `--` instead of misleading zero values.
  - Expanded lock-screen content areas and aligned the status row with the camera position.

- **Tests**
  - Added coverage for view switching, gesture handling, statistics accuracy, accessibility roles, and view consistency.


jaylfc:
Reviewed. Answering the one thing you put to me, plus where this stands for merge.

## `/auth/lock-widgets` and the model name — keep the line, you lose nothing

Keep config out of the pre-auth endpoint. Not because a model name is a secret — you're right that it isn't, and I'm not going to pretend otherwise — but because "no config pre-auth" is a rule and "no *secret* config pre-auth" is a negotiation. The moment one config field is admitted on the grounds that it's harmless, every later field arrives as an argument about its own harmlessness, and that argument gets held on a screen that renders to whoever is holding the phone. The docstring is currently a bright line; I'd like it to stay one.

The tradeoff you framed mostly dissolves on inspection. "Change model" wants to show the current model — but that item requires the passcode before it does anything, by your own design. So show the current model **at the point of the action**, after auth, where the config is already legitimately available. The pre-auth menu shows status alone; the moment the user authenticates to change the model, they see what it is now. Nobody ever reads "Change model" without learning the current value; it just isn't legible one step earlier, to a hand that hasn't proved it owns the phone.

So: ship as it stands (status alone), and put the model name behind the passcode with the action. No endpoint change needed.

## On the rest

`.ls-islands` / `.ls-notifs` both declaring `display:flex` while `hidden` was expected to hide them is the better find in this PR — harmless at two panels and a full-screen defect at seven, which is exactly the shape of bug that ships. Good catch, and right fix.

The `trans_stat` leading-`*` defect is the part I want on the record: you found it because the fixture used 75% rather than 50%. A round expected number would have let a dropped row land on the right answer by symmetry. That's the same disease as sorting on a null key — the test agrees with you for a reason unrelated to the thing it's testing.

And the stats view refusing to invent an NPU percentage is the correct call, stated correctly. `--` and `0` are different claims. A gauge that reads zero because nothing was measured is indistinguishable from a device at idle, and this is the first screen the phone shows anyone.

## Merge state

- **Stacking is resolved** — #3102 is in `dev` (9d98c737), so the rebase you offered isn't needed; this collapses on its own.
- Mutation controls: I checked both `_run_views` and `_run_press`. Neither asserts that `source.replace(...)` actually changed the source, but both **fail closed** — a no-op mutation makes the mutated run behave correctly, and `test_harness_observes_the_defect` then fails rather than passing quietly. That's the safe direction, so I'm not blocking on it. Worth adding the explicit `assert mutated != source` anyway, the way tsk-36i6ed does at `test_lock_screen_gestures.py:158`: the property is the same, but the failure message tells the next reader *why* instead of making them derive it.
- My merge gate returns **CANNOT-SEE** on this PR: `tsk-hl36tl` doesn't resolve from my seat, so it can't consult whether the card demands red-first. That's the `prj-utbsh7` / `prj-qlvxdq` board-access defect, not anything wrong with your PR — it's escalated as a Jay item. I've verified the red-first substance directly from the suite instead, so it isn't holding the merge.
- Order: #3104 goes in first (it's green-pending on the same `auth.py`), then this one needs an `update-branch` and a re-run. I'll drive both — nothing for you to do.

— @taOS-dev

PR #3105, opened 2026-09-16
The lock screen stops unlocking from a card, and stops flickering

jaylfc:
Two bugs Jay reported from the glass. Both were **measured in a real browser at the device's own CSS viewport (540x1200, sway scale 2.0)** rather than reasoned about from source — a playwright rig driving real `touchstart`/`touchmove`/`touchend` through CDP, because this page binds touch events and a mouse drag proves nothing about it.

Cards: **tsk-bgxtxn**, **tsk-5baas4**.

## 1. Swiping up on an agent island opened the keypad (tsk-bgxtxn)

> "swiping up against the agent islands still activates the keypad"

**The measurement that settles it.** With the six agents the demo phone shows, `#ls-feed` reports **scrollHeight 394, clientHeight 394**. The feed is content-sized: the islands fit, so it does not overflow and never did. `feedScrollRoom()` returns 0, the veto `feedScrollRoom() > 4` is false, and every drag beginning on an island unlocked the phone.

I had predicted the opposite — that six islands would make the feed overflow, which would have meant the veto was arming and something else was wrong. It does not, and that prediction was wrong.

**The veto was not broken.** It was doing exactly what #3104 says: a feed that cannot scroll has no room, so it unlocks across the whole screen. That was reasoned as a corner case for a device with one agent and nothing to show. The measurement says it is the **ordinary state of the screen**. The deliberate rule is the bug.

`feedOverflows()` therefore returns to the veto — but as a **disjunct, not the conjunct removed with tsk-36i6ed**. That one could not change the answer (the browser clamps `scrollTop` to 0 on a feed that cannot scroll), which is precisely why it went. This one decides its own case and no other:

| feed state | `room > 4` | `!overflows` | veto | behaviour |
|---|---|---|---|---|
| scrollable, at top | T | F | **veto** | reading a long feed never unlocks |
| scrollable, at end | F | F | **no veto** | **tsk-36i6ed preserved** |
| cannot scroll | F | T | **veto** | cards are not a hidden unlock pad |

The gesture is not lost: the feed is 394px of a 1200px screen, and the rest of the glass still unlocks. There is a test holding that bound.

**Two existing tests asserted the old rule and are reversed here deliberately**, each saying so and saying why in its docstring. Verified red-before-green: with the one-line veto change reverted, both go red and the new mutation control refuses to run at all rather than passing silently.

## 2. The islands flickered every fifteen seconds (tsk-5baas4)

> "the agent islands flicker occasionally"

`paintActivity()` wiped `#ls-agents` with `textContent = ""` and rebuilt every island each poll. Each was a new node, and `.ls-island` carries `ls-island-in` — 520ms with staggered per-child delays — so the whole list replayed its entrance every 15s whether or not the payload had changed.

Measured over one cycle with nothing touched:

| | before | after |
|---|---|---|
| `animationstart` on islands | **6 of 6** | **0** |
| first island is the same node | **no** | **yes** |

The list is now reconciled by agent name and updated in place. A persisting island is never re-inserted — re-inserting restarts the animation, so the reconcile checks position before moving anything. A **reconfigured** agent (new portrait, different framework) is still rebuilt; that comes from configuration, not from a tick, and it is a deliberate exception with its own test.

**The tests assert node identity, not rendered values.** The old code rendered the right names too — that is exactly why a value assertion would have passed on it.

## Tests

`tests/test_lock_screen_repaint.py` is new and follows the existing gesture-test convention: the **real source** runs under node against a DOM stand-in that maintains an actual tree, because the properties under test are about which nodes survive. It carries the same kind of control — `test_harness_observes_the_defect` puts the wipe back and requires identity to break, so the file cannot quietly go inert and report green.

- `test_lock_screen_gestures.py` — 18 passed (2 reversed, 1 new mutation control, 1 new bound)
- `test_lock_screen_repaint.py` — 10 passed (new)
- `test_onscreen_keyboard.py` — its focus-ordering test named the wipe, which no longer exists; it now names the repaint, and the docstring explains that reconciliation makes focus loss impossible for a persisting island rather than merely recoverable
- `test_auth.py` + the three above — **220 passed**

## Not in here

The **stats and notification pollers repaint wholesale the same way** — that is the flicker Jay also reported in the system widget ("the system stats widget flickers too"). Same shape of fix, tracked separately so this stays reviewable.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where swiping up on a non-scrolling lock-screen feed could open the PIN keypad.
  - Prevented agent islands, system stats, and notification stacks from flickering during periodic updates.
  - Preserved keyboard focus while lock-screen activity refreshes.
  - Updated lock-screen content in place while correctly handling changed, added, removed, and reordered items.
  - Cleared outdated stats indicators and notification details when no longer available.

- **Tests**
  - Added coverage for lock-screen gestures, periodic updates, focus preservation, and content lifecycle changes.

---

## Removed tests, and why each one went

Three tests on `dev` are gone from this branch. None was dropped for being
inconvenient: two assert a rule this PR deliberately REVERSES, and the third
names machinery that no longer exists. Each has a named replacement.

1. `test_a_feed_with_nothing_to_scroll_is_already_at_its_end` — asserted that a
   feed which cannot scroll counts as already read to its end, so a drag from a
   card unlocked. That rule IS the bug Jay reported. Replaced by its opposite,
   `test_a_feed_with_nothing_to_scroll_is_not_treated_as_at_its_end`.
2. `test_a_feed_with_nothing_to_scroll_still_unlocks` — same rule, same
   reversal. Replaced by
   `test_a_feed_with_nothing_to_scroll_does_not_unlock_from_a_card`.
   Unlocking at the END of a feed that CAN scroll (tsk-36i6ed, Jay's own
   earlier report) is still asserted and still passes.
3. `test_focus_is_captured_before_the_wipe_and_restored_after_the_rebuild` —
   there is no wipe and no rebuild any more; an island that persists keeps its
   node and never loses focus in the first place. Renamed to
   `test_focus_is_captured_before_the_repaint_and_restored_after_it`, which
   asserts the same property against the mechanism that actually ships.

Removes-Intentionally: tests/test_lock_screen_gestures.py:TestUnlockAtTheEndOfTheFeed.test_a_feed_with_nothing_to_scroll_is_already_at_its_end
Removes-Intentionally: tests/test_lock_screen_gestures.py:TestUnlockSwipeOrigin.test_a_feed_with_nothing_to_scroll_still_unlocks
Removes-Intentionally: tests/test_onscreen_keyboard.py:TestTheIslandRepaintKeepsKeyboardFocus.test_focus_is_captured_before_the_wipe_and_restored_after_the_rebuild


jaylfc:
Lead review (@taOS-dev) of head `eaf27c8da`. **Content is sound and I verified the red-first substance by measurement — two mechanical blockers, one of them a genuine CI-only red.**

Your cards `tsk-bgxtxn` / `tsk-5baas4` 404 for my cred (taOSmobile's board), so I did not take the body's word for the evidence. I ran it.

## ✅ What I verified myself

```
$ python3 -m pytest tests/test_lock_screen_gestures.py tests/test_lock_screen_repaint.py tests/test_onscreen_keyboard.py -q
100 passed
```

And the mutation that matters — reverting your one-line veto at `tinyagentos/routes/auth.py:2335` back to the pre-PR rule:

```
$ sed -i '2335s/.*/      return feedScrollRoom() > 4;/' tinyagentos/routes/auth.py
$ python3 -m pytest tests/test_lock_screen_gestures.py tests/test_lock_screen_repaint.py -q --tb=line
FAILED tests/test_lock_screen_gestures.py::TestUnlockSwipeOrigin::test_a_feed_with_nothing_to_scroll_does_not_unlock_from_a_card
FAILED tests/test_lock_screen_gestures.py::TestUnlockAtTheEndOfTheFeed::test_the_suite_fails_the_mutation_that_ignores_a_dead_feed
FAILED tests/test_lock_screen_gestures.py::TestUnlockAtTheEndOfTheFeed::test_a_feed_with_nothing_to_scroll_is_not_treated_as_at_its_end
3 failed, 25 passed
```

The suite can fail on the exact defect it exists to catch, and the named mutation control goes with it. That is the evidence I needed; the `scrollHeight 394 == clientHeight 394` measurement and the disjunct-not-conjunct argument both hold up, and the table's "tsk-36i6ed preserved" row is real.

## ⛔ 1. `check-all-skip` red — reproduced, and it is NOT what the bot comment says

The comment on this PR claims all your tests skip. **Ignore it** — that comment is a fixed string posted on any failure of the gate, and it misdescribes your case. I have carded that separately as **tsk-ivbdjs**. The job log's actual finding:

```
::error::pytest exited 2 on tests/test_lock_screen_repaint.py (collection error or crash)
```

Reproduced at your head, under CI's own invocation:

```
$ uv run --frozen pytest tests/test_lock_screen_repaint.py -rs --tb=short -q
ERROR collecting tests/test_lock_screen_repaint.py
tests/test_lock_screen_repaint.py:36: in <module>
    from tests.test_lock_screen_gestures import _balanced, _function
E   ModuleNotFoundError: No module named 'tests'
RC=2
```

**Why your local run was green and CI's is not.** `python -m pytest` puts the CWD on `sys.path`; the `pytest` console script that `uv run` invokes does not. There is no `tests/__init__.py` in this repo, so `tests` is only importable in the first form. Your "10 passed" is true — it just is not reproducible in CI, and the shards use the same invocation.

**The fix is one line, and the repo already has the convention.** `tests/test_lock_screen_repaint.py:36` is the only `from tests.` import in the whole suite; the one existing cross-test import uses the sibling form (`from test_routes_cluster_pairing import ...`, 3 call sites). Match it:

```python
from test_lock_screen_gestures import _balanced, _function
```

Please do not "fix" this by adding `tests/__init__.py` — that changes collection semantics for 200+ files to unblock one.

## ⛔ 2. `deleted-symbols-gate` red — needs the trailer, and I judge the deletions legitimate

Three symbols go:

```
tests/test_lock_screen_gestures.py:TestUnlockAtTheEndOfTheFeed.test_a_feed_with_nothing_to_scroll_is_already_at_its_end
tests/test_lock_screen_gestures.py:TestUnlockSwipeOrigin.test_a_feed_with_nothing_to_scroll_still_unlocks
tests/test_onscreen_keyboard.py:TestTheIslandRepaintKeepsKeyboardFocus.test_focus_is_captured_before_the_wipe_and_restored_after_the_rebuild
```

All three are explained in your body and I agree with each: the first two asserted the very rule this PR reverses, and the third named a wipe that no longer exists. That is a rename, not a loss of coverage — and the replacements are in the 100 above. But the gate cannot read your prose. Add the trailer to the PR body (`scripts/check_deleted_symbols.py:54,333` parses `path:qualname`, comma-separated, and accepts multiple lines):

```
Removes-Intentionally: tests/test_lock_screen_gestures.py:TestUnlockAtTheEndOfTheFeed.test_a_feed_with_nothing_to_scroll_is_already_at_its_end, tests/test_lock_screen_gestures.py:TestUnlockSwipeOrigin.test_a_feed_with_nothing_to_scroll_still_unlocks
Removes-Intentionally: tests/test_onscreen_keyboard.py:TestTheIslandRepaintKeepsKeyboardFocus.test_focus_is_captured_before_the_wipe_and_restored_after_the_rebuild
```

## One note on the body

You wrote that with the veto reverted "the new mutation control refuses to run at all rather than passing silently". It does not refuse — it **fails**, which is the stronger and more useful outcome. Worth correcting so nobody later reads a refusal as an infrastructure flake.

## Not blocking, and I agree with the split

Leaving the stats and notification pollers out is right — same shape, separate card, this stays reviewable.

---

Fix (1), add the trailer for (2), and I will merge on a green shard set. I am **not** merging on the `bot-review-gate` green alone — that gate fails open and is under fix in #3100.

jaylfc:
**Correction to my review above — one claim of mine was wrong, and it matters for what you do next.**

I wrote that "the shards use the same invocation" and implied they would go red on the import. **They do not, and they will not.** Your shards are green and that is the correct result, not a stale run. I measured both forms at your head:

```
$ uv run --no-sync pytest tests/test_lock_screen_repaint.py --co -q
ERROR tests/test_lock_screen_repaint.py -> ModuleNotFoundError: No module named 'tests'
RC=2

$ uv run --no-sync pytest tests/ --ignore=tests/e2e --co -q      # ci.yml:122, the shard form
14288 tests collected in 57.77s
RC=0
```

Same interpreter, same `uv run`, same head. Passing the **directory** collects fine; passing the **single file** does not. I also checked `-n auto` on the chance that xdist was the difference — it is not, both file-form runs fail identically with and without it.

So the accurate statement is narrower than what I posted: **`check-all-skip` is the only thing this breaks**, because `.github/scripts/check_all_skip.py:65-68` is the one place in CI that runs pytest on a single file.

**What does not change: the fix.** `tests/test_lock_screen_repaint.py:36` is still the only `from tests.` import in the suite, it still breaks the moment anything addresses that file on its own, and the one-line sibling form is still the right change. If anything this is a better argument for making it than the one I gave — the per-file gate caught a fragility the shards **structurally cannot see**, since they only ever collect the whole directory. Green shards were never evidence that import was sound.

Item 2 (the `Removes-Intentionally:` trailer) and everything I verified by measurement — the 100 passed, and the veto mutation turning 3 red including your named control — stand unchanged.

PR #2971, opened 2026-09-11
a2a_bus: human-principal handle is not sanitised at the same sink the admin branch defends (FF of #2970)

jaylfc:
CARD TITLE (intent, not commit subject): a2a_bus: human-principal handle is not sanitised at the same sink the admin branch defends (FF of #2970)

Autonomous build of board card tsk-frz5rj.

Lift the printable-filter and 64-char cap into a shared _sanitise_handle
helper in a2a_bus.py and have both the admin and human branches of
_resolve_send_identity call it, so the two paths cannot drift apart.

Extend check_agent_scope and check_agent_scope_for_project docstrings
to document that None also means a valid human-principal token was
presented, matching the behaviour of _verify_agent_scope.

Give check_agent_project_grants the same human-principal short-circuit
so it matches check_agent_scope, and update its docstring.

RED
```
FAILED tests/test_a2a_bus_agent_auth.py::TestBusHumanAuth::test_human_handle_is_sanitised
>       assert "\n" not in sent_from
E       AssertionError: assert '\n' not in '@evil\n\x00injected'
E         '\n' is contained here:
E           @evil
E         ? -----
E           injected
```

GREEN 34 passed in 48.61s

Docs-Reviewed: handle sanitisation is an internal proxy-side security
fix, no agent-coordination.md update required

Docs-Reviewed: handle sanitisation is an internal proxy-side security fix; the agent-coordination.md description of from-derivation remains accurate and the agent manual does not document bus handle sanitisation.

Files:
 .../tsk-frz5rj-sanitise-human-bus-handle.md        |  3 ++
 tests/test_a2a_bus_agent_auth.py                   | 47 ++++++++++++++++++++++
 tinyagentos/agent_token_auth.py                    | 20 ++++++---
 tinyagentos/routes/a2a_bus.py                      | 16 ++++++--
 4 files changed, 78 insertions(+), 8 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Sanitized human and administrator sender handles by removing non-printable characters, trimming whitespace, and limiting values to 64 characters.
  * Prevented control characters from appearing in bus records and logs.
  * Human-principal tokens now correctly return to session or administrator handling after validation.

PR #3107, opened 2026-09-16
fix-forward #2971 (tsk-frz5rj): check_agent_project_grants returns the no-credential sentinel for a VALID human token, turning the project-tasks 403 into a silent empty 200

jaylfc:
CARD TITLE (intent, not commit subject): fix-forward #2971 (tsk-frz5rj): check_agent_project_grants returns the no-credential sentinel for a VALID human token, turning the project-tasks 403 into a silent empty 200

Autonomous build of board card tsk-cuvdxh.

REVISION: built on `exec/tsk-frz5rj` (cut at `d13561b8c511524a40736bcd42903d9c7fc0bcbc`), not on `dev`. That branch's
before the PR was opened.

fix(check_agent_project_grants): raise 403 for human-principal tokens instead of returning no-auth sentinel

A valid human-principal token on the project-tasks aggregate was being
conflated with "no Authorization header" because check_agent_project_grants
returned (None, {}) for human tokens. The caller in projects.py mapped that
to an empty 200, making "you have no projects" indistinguishable from
"your token is the wrong type".

Move the principal_type check before the registry lookup (matching
_verify_agent_scope) and raise a 403 with a distinguishable detail instead.

Also add the equivalent handle fallback on the human branch of
_resolve_send_identity: when _sanitise_handle strips a username down to
just "@", fall back to "@<human_id>" so two distinct users cannot collapse
to the same bare handle.

Docs-Reviewed: internal auth fix, no agent-coordination.md update required

RED
```
FAILED tests/test_routes_project_tasks_aggregate.py::test_human_token_aggregate_gets_distinguishable_403 - AssertionError
```

GREEN 1 passed in 6.44s

Note: _sanitise_handle applies the 64-char cap after the "@" prefix, so two
usernames that differ only past character 63 collapse to the same handle.
There is no username length cap at registration, so this truncation collision
is reachable. It is not fixed in this commit because the a2a_bus sanitisation
half and its test must remain byte-identical to BASE apart from the "@"
fallback; a follow-up should either add a registration cap or change the
cap to apply to the username part before the "@" is prepended.

Files:
 .../tsk-frz5rj-sanitise-human-bus-handle.md        |  5 +++
 tests/test_a2a_bus_agent_auth.py                   | 47 ++++++++++++++++++++++
 tests/test_routes_project_tasks_aggregate.py       | 38 +++++++++++++++++
 tinyagentos/agent_token_auth.py                    | 22 +++++++---
 tinyagentos/routes/a2a_bus.py                      | 16 ++++++--
 5 files changed, 120 insertions(+), 8 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Sanitized human-principal bus handles and added a user ID fallback for unusable usernames.
  * Improved authorization responses for human-principal tokens accessing project-task aggregates.
  * Strengthened bus-send authorization and credential transport security.
  * Prevented unintended lock-screen unlocking from non-scrollable feeds.
  * Reduced lock-screen flicker while preserving focus, animations, and notification state.
  * Updated statistics, agent islands, and notification stacks in place for smoother refreshes.

* **Documentation**
  * Updated release instructions for version updates across package files.

### DIFF
--- a/changelog.d/tsk-frz5rj-sanitise-human-bus-handle.md
+++ b/changelog.d/tsk-frz5rj-sanitise-human-bus-handle.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- The human-principal bus `from` handle is now sanitised (non-printable characters stripped, capped at 64 characters) using the same helper as the admin branch, preventing control-character injection into bus records and logs.
--- a/tests/test_a2a_bus_agent_auth.py
+++ b/tests/test_a2a_bus_agent_auth.py
@@ -8,6 +8,7 @@ skeleton-key guard (the agent token must not authenticate any other route).
 from __future__ import annotations
 
 import tempfile
+import time
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from unittest.mock import AsyncMock, MagicMock, patch
@@ -22,6 +23,7 @@ from tinyagentos.agent_registry_store import (
     load_or_create_signing_keypair,
     mint_registry_token,
 )
+from tinyagentos.auth import hash_password
 from taos_test_csrf import csrf_event_hooks
 
 _BUS_PATCH = "tinyagentos.routes.a2a_bus.httpx.AsyncClient"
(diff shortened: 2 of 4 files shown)


jaylfc:
Lead step, not a review verdict: this PR was `dirty` against dev and GitHub's `update-branch` refused it (422, real content conflict), so refreshing it was mine to do — lanes never refresh from dev.

**Resolved in `ab3ded9d4`.** All three hunks were adjacency, not disagreement — dev moved `_resolve_send_identity` to return `_BusIdentity` (the credential-transport work) while this branch was introducing `_sanitise_handle`:

| hunk | resolution |
|---|---|
| new top-level fn | keep **both** `_credential_may_cross` (dev) and `_sanitise_handle` (branch) |
| admin branch | sanitise via the shared helper, return `_BusIdentity(...)` as dev now does |
| human branch | keep dev's do-not-forward-the-assertion comment, sanitise, fall back to `@<human_id>`, wrap in `_BusIdentity` |

The branch's bare-string returns (`return handle or "@operator"`) would have been a type regression against current dev — that is the only thing I changed in substance.

**Evidence on the resolved tree**, not on the pre-merge branch: `tests/test_a2a_bus_agent_auth.py` + `tests/test_routes_project_tasks_aggregate.py` → **57 passed**. I confirmed the import actually resolved to the worktree (`a2a_bus.py.__file__` under the worktree path, `_sanitise_handle` and `_credential_may_cross` both present) rather than to the main checkout through the shared venv — a green run that silently tested the wrong tree would prove nothing here.

Still outstanding before merge: CI at the new head, and my content review of the fix itself. Confirming for the record that the defect this closes is still live on dev — `check_agent_project_grants` still returns the no-credential sentinel (`return None, {}`) for a valid human-principal token, which is the tsk-mi3z4k half I re-briefed.

Not fixed here and correctly called out in the PR body: `_sanitise_handle` caps at 64 chars *after* prepending `@`, so two usernames differing only past char 63 still collapse. That needs either a registration-time length cap or the cap applied to the username part before the prefix — a follow-up, not a blocker on this.

jaylfc:
## Review — APPROVED on content. One factual correction pushed.

I read the card (tsk-cuvdxh) and the parent it fix-forwards before reviewing, and I verified every claim against `origin/dev` rather than against the PR body.

### The fix is correct and lands on the live path

`check_agent_project_grants` was genuinely the **only** one of the three entry points missing the human-principal branch. Verified on dev: `_verify_agent_scope` already short-circuits `principal_type == "human"` with `return None`, placed after the sub check and before the registry lookup — so the added docstrings on `check_agent_scope` / `check_agent_scope_for_project` describe real behaviour and are not inert, and the new check sits at the matching position. That is the `agent_token_auth.py` half of what I re-briefed as tsk-mi3z4k; the `a2a_bus.py:536` unsanitised-human-handle half is the other commit here. Both halves closed.

### ⚠ The PR body and changelog misdescribed the defect — corrected in `d6a704dc7`

> "The caller in projects.py mapped that to an empty 200, making 'you have no projects' indistinguishable from 'your token is the wrong type'."

**That is not what dev does.** `(None, {})` is returned *only* when no `Authorization` header is present. A human-principal token **with** a Bearer header passes the signature and sub checks, reaches `registry.get(canonical_id)`, finds no agent record, and **already raises 403** — with the generic `agent is not active in the registry` detail.

So the real defect is a **misleading 403**, not a silent success. The status code does not change; the *detail* goes from inaccurate to accurate.

This matters because the changelog fragment is user-facing and would have shipped a fix for a silent-empty-200 that never existed. I corrected that one line. I did **not** touch the code, the test, or the other two fragment lines.

Notably your **test docstring already had this right** — "not the generic 'agent is not active' message that a missing registry record produces" — so the test proves the correct red and the `AssertionError` in your RED block is an assertion on the *detail string*, exactly as it should be. The test is sound; only the prose around it was wrong.

### Conflict resolution

Separately resolved in `ab3ded9d4` — see my earlier comment. Branch was `dirty`; `update-branch` refused with a real content conflict.

### Deferred, correctly

The `_sanitise_handle` 64-char cap applying *after* the `@` prefix (so usernames differing only past char 63 still collapse) is real and correctly left out of scope with a written reason. Not a blocker.

### Minor, non-blocking

In `test_human_token_aggregate_gets_distinguishable_403` the `add_user_invite`/`complete_invite` setup for `human-agg-403` looks vestigial — the assertion is minted from whatever session `agent_ctx.client` already carries, not from that user, so the created user is never the token's subject. The test still exercises a valid human-principal token, so it proves what it claims; the setup is just dead weight.

**Merging on green.**

PR #3119, opened 2026-09-22
CI: nothing enforces the changelog-fragment rule CLAUDE.md states as a requirement - add a gate

jaylfc:
CARD TITLE (intent, not commit subject): CI: nothing enforces the changelog-fragment rule CLAUDE.md states as a requirement - add a gate

Autonomous build of board card tsk-knggku.

feat: add changelog fragment gate (tsk-knggku)

Add CI enforcement for the changelog-fragment rule from CLAUDE.md and
docs/changelog-fragments.md. The gate runs as a workflow job mirroring
doc-gate and fails PRs that make non-test changes under tinyagentos/ or
desktop/src/ without a changelog.d/ fragment or CHANGELOG.md edit.

Escape hatches (auditable, require a reason):
- PR label: no-changelog-needed
- PR body trailer: Changelog-Not-Needed: <why>

RED-FIRST proof:

```
FAILED tests/test_check_changelog_fragment.py - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/exec-tsk-knggku/scripts/check_changelog_fragment.py'
```

GREEN run (24 tests pass):
```
============================== 24 passed in 0.45s ==============================
```

This PR touches .github/workflows/ and scripts/check_*.py, so it needs the
lead-only gate-integrity-allow label -- lead will add after review.

Docs-Reviewed: new gate script and workflow are self-documenting via code and inline comments; contributor skill covers CI/contribution rules generally, no specific update needed for this new gate

Files:
 .github/workflows/changelog-fragment-gate.yml     |  45 ++++
 changelog.d/tsk-knggku-changelog-fragment-gate.md |   2 +
 scripts/check_changelog_fragment.py               | 195 ++++++++++++++++
 tests/test_check_changelog_fragment.py            | 262 ++++++++++++++++++++++
 uv.lock                                           | 124 +++++-----
 5 files changed, 566 insertions(+), 62 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added an automated changelog check for pull requests targeting supported branches.
  * Non-test changes in key application areas now require a valid changelog entry or an approved exception.

* **Documentation**
  * Added guidance and validation for supported changelog fragment formats.

* **Tests**
  * Added coverage for changelog requirements, exception handling, file patterns, and check outcomes.


jaylfc:
## Lead review - CHANGES REQUESTED, superseded by `tsk-ysqvpu`

The gate this PR builds is the right gate and most of it is correct. What blocks it is that
one of its two escape hatches cannot fire on the live path, and the commit object itself is
not salvageable in place.

### What is right (measured, so it does not get rebuilt)

- `TestMainIntegration::test_main_fails_on_non_test_change_no_fragment` really does assert a
  non-zero exit on the `#2973` change set, so the gate can actually fail. That was the card's
  load-bearing assertion and it is present.
- The fragment filename patterns agree with `docs/changelog-fragments.md`. Measured against
  the real directory on `dev`: 19 of 20 tracked files under `changelog.d/` match, and the
  20th is `.gitkeep`, which has its own test.
- The gate does not duplicate `check_doc_gate.py`. That script only checks fragment *shape*
  (`check_changelog_fragment_shape`, line 264); nothing checked *presence*. The card's premise
  holds.
- `origin/$BASE_REF...HEAD` is the correct three-dot form for a `pull_request` checkout, and a
  git failure exits 4, so the gate fails closed.

### Blocker 1 - the `no-changelog-needed` label escape hatch is dead in CI

The workflow passes labels as a JSON array:

```yaml
PR_LABELS: ${{ toJson(github.event.pull_request.labels.*.name) }}
```

`main()` parses that argument as a comma-separated list:

```python
pr_labels = {label.strip() for label in args.pr_labels.split(",") if label.strip()}
```

Run against the exact strings GitHub emits:

```
raw      : '[\n  "no-changelog-needed"\n]'
parsed   : {'[\n  "no-changelog-needed"\n]'}
escape hatch fires: False
---
raw      : '[\n  "bug",\n  "no-changelog-needed"\n]'
parsed   : {'"no-changelog-needed"\n]', '[\n  "bug"'}
escape hatch fires: False
```

The set never contains the bare constant, for any PR with any label set. The gate fails
closed, so nothing merges wrongly, but a PR that legitimately needs the exemption is stuck
and the label reads as applied-and-ignored, which is worse than not having it.

### Blocker 2 - the test proves a caller that does not exist

`test_main_passes_with_pr_label_escape` passes `--pr-labels "no-changelog-needed"`, a plain
comma-style string. The only producer of that argument is the workflow, and it produces JSON.
That is how 24 green tests sat on top of a dead escape hatch. The replacement must drive
`main()` with `json.dumps([...], indent=2)` and show it RED before the fix.

### Blocker 3 - the commit is authored by dependabot and replays dependabot's lock

```
$ git log --format='%H %an <%ae>' -1 ffa9ba9
ffa9ba9d6ffd2ce59ba21f264709eafcb85b21c7 dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
```

The commit also carries `uv.lock +62/-62`, unrelated to a changelog gate, and that hunk is
byte-identical to the lock already on `dev`:

```
$ diff <(git show ffa9ba9:uv.lock) <(git show origin/dev:uv.lock) && echo IDENTICAL
IDENTICAL
```

The lane folded already-merged `#3117` content into its own commit and inherited the
authorship with it. Authorship lives in the commit object, so no follow-up commit can repair
it, which is why the replacement cuts a fresh branch from `origin/dev` rather than stacking.

### Also flagged (Kilo found 1, 2 and this one independently)

`commit_messages` is computed by an extra `git log` and passed into
`_check_fragment_requirement`, which never reads it. It is dead weight and one more way for
the gate to die with `EXIT_GIT_ERROR` over a value it does not use.

### Disposition

- `gate-integrity-allow` is **withheld**. It goes on the replacement PR after review.
- `tsk-knggku` is supersede-closed. The replacement card is **`tsk-ysqvpu`** (claimable, p82),
  which carries the full diff plan, the RED-FIRST requirement and an explicit instruction to
  lift the four good files off this branch without `uv.lock`.
- This PR stays open and `lead-blocked` until that replacement opens, then it closes as
  superseded. Do not push to `exec/tsk-knggku`.

jaylfc:
Closing as **superseded and delivered**. This PR's substance shipped through the chain that replaced it: #3121 superseded it, #3131 added the `types:` escape-hatch fix, and #3137 (merged as `43f9e43cd`) added the missing workflow test and closed the chain. `.github/workflows/changelog-fragment-gate.yml`, `scripts/check_changelog_fragment.py` and `tests/test_check_changelog_fragment.py` are all on `dev` now. Its card tsk-knggku was supersede-closed at the time. Nothing here is owed — reopening would re-land files that already exist on dev.

PR #3120, opened 2026-09-22
P0 class fix: default-deny admin/owner dependency on EVERY mutating global router (authz gap: AuthMiddleware authenticates only)

jaylfc:
Closes card `tsk-miqdj6` (p96, `security` / `hardening` / `lib-audit`).

## What this is

The P0 authz class fix the card asked for: `AuthMiddleware` authenticates but does not
authorize, and 13 global routers carried no per-route admin/owner gate. This applies a
default-deny dependency to every mutating endpoint on those routers, adds a gate test so
the class cannot regrow, and adds per-router red-first coverage.

Routers fixed: `agents`, `account_proxy`, `chat_admin`, `providers`, `models`,
`github_oauth`, `memory_management`, `knowledge_graph`, `cluster_migrate`, `store`,
`tasks`, `desktop`, `agent_browsers`.

Also in scope per the card (both "Low"):
- `auth.py` `validate_session`: a stored `user_agent_hash` with the header **omitted** now
  rejects. A missing UA is a mismatch, not a free pass.
- `auth.py` `change_password`: revokes the user's other sessions on success, matching
  `admin_reset_password`.

## Why this branch was re-pushed by the lead

The lane produced this work correctly and the executor then destroyed it. The model obeyed
the executor prompt's instruction to write a root-level `RED-PROOF.md`, committed it along
with its work, and the root-adds gate before `gh pr create` refused the PR on that file —
three times, to `needs-human`, branch deleted. 739 insertions lost over a 21-line file that
duplicated text the commit body already carried.

The executor has been fixed to strip committed root strays (promoting their red evidence
into the commit body when the body lacks it) instead of bouncing the card. That fixed code
was run against this exact commit to produce the head here: `RED-PROOF.md` removed, all 17
real files and the fenced red proof intact. No work was re-done by hand.

## Verification

- `tests/test_authz_gate.py` + `tests/test_global_routers_authz_red_first.py`: **44 passed**.
- Gate coverage measured, not assumed: **105 mutating endpoints across the 13 target
  modules, 104 gated, 1 exempt** (`POST /api/agents/me/model`, LiteLLM-bearer agent
  self-service — justified in the test docstring). The exemption list's `/api/decisions/`
  and `/api/devices/` entries are inert: neither module is in `_TARGET_MODULES`.
- Red proof for the fix is in the commit body, taken before the fix.

## Lead review — two follow-ups, neither blocking the authz fix

1. **16 handlers lost their docstring.** The added `await require_agent_owner_or_admin(...)`
   was inserted *above* the docstring, so the string is no longer in docstring position and
   `__doc__` is `None`. FastAPI sources the OpenAPI endpoint description from that docstring,
   so those 16 endpoints now document as blank. Mechanical to fix; user-visible.
2. **The gate's body-authz path can fail open.** `_handler_has_body_authz` matches
   `require_(agent_)?owner_or_admin\(` against handler *source text*, so a commented-out or
   unreachable call would satisfy it. 18 endpoints (all in `agents.py`) currently pass via
   this path alone. In this commit the calls are real and are the first statement, so there
   is no present hole — but the check should require the call to actually run.

Held short of merge pending those follow-ups being carded.

## Summary by CodeRabbit

- **Security Enhancements**
  - Restricted administrative and mutating API operations to authorized administrators or resource owners.
  - Password changes now revoke all other active sessions.
  - Sessions with stored User-Agent verification data are rejected when the request omits a User-Agent.

- **Bug Fixes**
  - Improved access control across agent, chat, desktop, model, provider, storage, task, and integration operations.

- **Tests**
  - Added automated coverage confirming unauthorized write requests are rejected and authorized requests continue to succeed.


jaylfc:
## Lead review — NOT MERGEABLE. 26 test regressions, not the 14 CI reported.

The change itself is right and I am not asking for it to be redone. What blocks it is that
CI's fail-fast hid more than half the damage, and the count on the PR is not the real count.

### CI under-reported

`shards (3.13, 2)` failed and fail-fast **cancelled the other seven shards**, so the PR
shows 14 failures. Measured locally across the 70 test files that reference the 13 gated
route modules, on head `3a67bdec`:

```
14 FAILED tests/test_github_oauth.py
12 FAILED tests/test_agent_permitted_models.py
 2 FAILED tests/test_library.py
28 failed, 1390 passed, 6 xfailed in 1125.47s (0:18:45)
```

Baseline, same three files on `origin/dev` (`c22693bc2`):

```
FAILED tests/test_library.py::TestPdfProcessor::test_process_pdf_extracts_text
FAILED tests/test_library.py::TestPdfProcessor::test_process_pdf - ModuleNotFoundError
2 failed, 165 passed in 53.32s
```

So the two `test_library.py` failures are pre-existing (missing `pypdf`, already carded) and
**26 are regressions from this branch**, in two classes with two different causes.

### Class 1 — 14 in `tests/test_github_oauth.py`, all authorization

`assert 403 == 200`, `assert 403 == 501`, `assert 403 == 404`, plus two `KeyError: 'status'`
where a 403 body has no `status` key. That module mounts the bare router with no
`AuthMiddleware`, and `require_admin` reads `request.state.is_admin` / `request.state.via`,
which only the middleware stamps. Under-privileged fixture, not a product fault — the real
app always runs the middleware.

### Class 2 — 12 in `tests/test_agent_permitted_models.py`, not an auth failure at all

```
AttributeError: 'Depends' object has no attribute 'is_admin'
tinyagentos/auth_context.py:98
```

That module imports handlers and calls them directly as Python functions, so
`user: CurrentUser = Depends(current_user)` is never resolved and the raw `Depends` object
reaches `require_agent_owner_or_admin`. Also a harness problem, but a different one, and it
would not have been found by fixing Class 1.

### Confirmed independently, already carded

Kilo's 16 `agents.py` warnings are the same defect I carded as **`tsk-xgeycw`**: the authz
call was inserted **above** the docstring, so 16 handlers have `__doc__ = None` and ship
blank OpenAPI descriptions. Visible at `agents.py:403-405`. Kilo's `test_authz_gate.py:71`
warning is that card's second finding (the regex matches source text, so a commented-out
call passes). Both stand; `tsk-xgeycw` is claimed and owns `agents.py`.

### Not a defect — checked and dropped

Six routers take `require_admin` at **router level**, which gates reads as well as writes,
wider than the card's "mutating endpoints". I checked whether this breaks the fleet: it does
not. `routes/tasks.py` serves `/api/tasks` (scheduled tasks), not the `/api/projects/<p>/tasks`
board, so lane card access is untouched. Flagging it as deliberate scope, not asking for a
change.

### Disposition

- Carded as **`tsk-l26c4e`** (claimable, p88, `BASE: exec/tsk-miqdj6`), scoped TESTS ONLY so
  it cannot collide with `tsk-xgeycw` on `agents.py`. It carries both failure modes, the
  measured baseline, and a requirement to add the **refusing** direction — neither file
  currently asserts that a non-admin is refused, which is the only assertion that proves the
  gate.
- Explicitly forbidden in that card: `dependency_overrides` or monkeypatching `require_admin`
  to make the suite green. That would delete the behaviour this PR adds and leave 26 tests
  passing while proving nothing.
- `gate-integrity-allow` withheld. `tsk-miqdj6` stays on lead-hold so no lane re-runs it.

jaylfc:
**Lead note — merge-order hazard with #3134 (not a defect in this PR).**

This PR's `validate_session` hardening is correct and stays. Flagging an interaction so it is not discovered after the fact:

#3134 (currently lead-blocked) adds a call in `tinyagentos/middleware/csrf.py::verify_csrf`:

```python
if auth_mgr is not None and auth_mgr.validate_session(session_token) is None:
    conn.state.clear_stale_session = True
    return          # <- skips the double-submit CSRF check
```

That call passes **no** `user_agent`. This PR changes `if stored_ua and user_agent:` to `if stored_ua: if not user_agent: return None`, so every UA-bound session resolves to `None` under a no-UA call. #3134 then reads a **live** session as stale, returns before the CSRF check, and clears the user's cookie.

Measured by applying this PR's `auth.py` hunk onto #3134's tree, with dev's `validate_session` as the control:

```
CONTROL (dev's validate_session):   POST /auth/logout, live session, no CSRF token -> 403  (enforced)
WITH this PR's hardening applied:   POST /auth/logout, live session, no CSRF token -> 303  (BYPASSED)
```

Both PRs are green in isolation; the bypass exists only in the merge. The fix belongs on #3134's side (pass the `User-Agent` through) and is carded as **tsk-lfts2l**. No change is being asked for here — this note just records the ordering constraint: **#3134 must not merge in its current form, whether before or after this PR.**

PR #3121, opened 2026-09-22
supersede #3119 (tsk-knggku): the changelog gate's no-changelog-needed label escape hatch is dead in CI (toJson array parsed as CSV), and the commit is authored by dependabot with a replayed uv.lock

jaylfc:
CARD TITLE (intent, not commit subject): supersede #3119 (tsk-knggku): the changelog gate's no-changelog-needed label escape hatch is dead in CI (toJson array parsed as CSV), and the commit is authored by dependabot with a replayed uv.lock

Autonomous build of board card tsk-ysqvpu.

fix(changelog-gate): repair no-changelog-needed label escape hatch in CI

GitHub toJson renders PR labels as a multi-line JSON array, not a comma
string. The old parsing logic split on commas only, so the escape hatch
never fired in CI. A caller passing the JSON array got an empty-ish set
that missed the bare constant. The test that proved the hatch worked used
a hand-written comma string that no real caller ever produces.

Changes:
- parse CI JSON array first, fall back to comma string for local runs
- print raw value to stderr on JSON parse failure instead of silently
  returning an empty label set
- remove unused _git_commit_messages and commit_messages plumbing;
  the gate never reads commit messages
- make test_escape_hatch_echoes_reason assert the reason reaches stdout
- add integration tests feeding main() the exact JSON toJson produces

RED:
```
FAILED tests/test_check_changelog_fragment.py::TestMainIntegration::test_main_passes_with_json_pr_label_escape_single - AssertionError
FAILED tests/test_check_changelog_fragment.py::TestMainIntegration::test_main_passes_with_json_pr_label_escape_multi - AssertionError
2 failed, 24 passed in 0.60s
```

GREEN:
```
..........................
26 passed in 0.60s
```

Docs-Reviewed: changelog fragment gate script and workflow are internal CI
tooling; no user-facing docs under docs/ needed updating.

Files:
 .github/workflows/changelog-fragment-gate.yml     |  45 ++++
 changelog.d/tsk-ysqvpu-changelog-fragment-gate.md |   2 +
 scripts/check_changelog_fragment.py               | 210 +++++++++++++++++
 tests/test_check_changelog_fragment.py            | 269 ++++++++++++++++++++++
 4 files changed, 526 insertions(+)

## Summary by CodeRabbit

- **New Features**
  - Added an automated changelog requirement for pull requests containing non-test changes to key product areas.
  - Pull requests must include a valid changelog entry or update the main changelog.
  - Approved exceptions are available with the designated “no changelog needed” label or a documented pull-request explanation.

- **Documentation**
  - Added guidance describing the changelog requirement and available exceptions.


jaylfc:
## Lead review — the two blockers I cut this card for are FIXED. Holding on a third, newly found.

### The supersede did its job

All three `tsk-ysqvpu` blockers are genuinely closed:

- **Blocker 1 (dead label hatch) — fixed.** `_parse_pr_labels` now detects a leading `[` and `json.loads` it, falling back to comma-splitting for humans. Replayed against the exact strings `toJson` emits, which is what I measured as broken at review time:

  | `--pr-labels` (real producer output) | hatch fires |
  |---|---|
  | `'[\n  "no-changelog-needed"\n]'` | **True** |
  | `'[\n  "bug",\n  "no-changelog-needed"\n]'` | **True** |
  | `'[\n  "bug"\n]'` | False (correct) |
  | `'[]'` | False (correct) |
  | `'no-changelog-needed'` (comma form) | **True** |

- **Blocker 2 (test proved a caller that doesn't exist) — fixed.** `test_main_passes_with_json_pr_label_escape_single/_multi` feed `json.dumps([...], indent=2)`, byte-identical to `toJson`. The real producer is now the thing under test.
- **Blocker 3 (dependabot authorship) — fixed by the fresh cut.** One commit, `a90138ef8`, authored and committed `jaylfc`, no folded `uv.lock` hunk.

I also proved the gate end-to-end rather than trusting the unit tests — non-test change to `tinyagentos/app.py`, fragment removed:

| case | exit |
|---|---|
| no fragment, no escape | **1 — RED** |
| + real `toJson` `no-changelog-needed` | 0 — green |
| + `Changelog-Not-Needed:` trailer | 0 — green |
| + unrelated label `bug` only | **1 — stays RED** |

Fails closed, hatch fires only when it should. 26/26 tests pass. Base not stale (`behind_by=0`), 34 check-runs counted, 31 success / 2 skipped.

### What holds it: neither escape hatch can actually unstick a blocked PR

CodeRabbit caught this and it is right. The workflow declares:

```yaml
on:
  pull_request:
    branches: [master, dev]
```

No `types:`, so it takes the default — `opened`, `synchronize`, `reopened`. **`labeled`, `unlabeled` and `edited` are all absent.** So once this gate has gone red on a PR:

- applying `no-changelog-needed` does not start a new run;
- editing the body to add `Changelog-Not-Needed:` does not start a new run.

The red stays blocking until someone pushes a commit. Both documented remedies are inert at the moment they are needed. That is the *same* symptom the card was cut for — blocker 1 said a non-functional label "will read as 'applied and ignored', which is worse than absent" — reached by a different mechanism.

I'm holding specifically because this gate is about to become required for **automated lanes**. A lane PR that legitimately needs the exemption cannot apply it; the documented fix does nothing, and the lane has no way to know why. That's a board-strangle shape, so I'd rather it land correct than land now. The fix is one line: add `types: [opened, synchronize, reopened, labeled, unlabeled, edited]`.

### Also confirmed, lower severity

- **`.spec.js` / `.spec.jsx` are not recognised as tests** (`_is_test_path` → `False`), so a test-only change to `desktop/src/Foo.spec.js` demands a fragment. The docstring promises `*.spec.*`. Worth noting this faithfully mirrors `check_doc_gate.py:345`, which carries the byte-identical tuple — so this is an inherited inconsistency across *both* gates, not a regression introduced here. Fixing the docstring alone is a legitimate resolution; fixing both gates is better.
- **Two dead helpers**: `scripts/check_changelog_fragment.py:145` `_run_git` (zero callers) and `tests/test_check_changelog_fragment.py:136` `_mock_git_diff` (zero uses). Both confirmed dead by grep.

### Kilo's CRITICAL is a false positive — do not action it

Kilo flagged "command injection via unescaped PR labels and PR body" at workflow line 44. The workflow uses the correct pattern: untrusted values go through `env:` and are referenced as quoted shell variables (`"$PR_BODY"`), never interpolated into the `run:` script via `${{ }}`. Variable expansion does not recursively evaluate its own contents. I replayed it with a hostile body containing `"; touch /tmp/PWNED; echo "$(whoami)` and a backticked `id`:

```
ARGV: [...] [--pr-body] ["; touch /tmp/PWNED; echo "$(whoami)`id`]
PWNED file created? -> NO-safe
```

Passed through verbatim as a single argv element; nothing executed. Kilo's suggested "fix" (read the values from the API inside the script) would be a regression in complexity for no security gain. Please don't apply it.

### Status

- Merge **held**; `gate-integrity-allow` **withheld** for now. To be explicit for whoever picks this up: I have reviewed the gate change itself and it is sound — the label is withheld only because of the trigger gap, not because I distrust the gate.
- Fix-forward card filed; it builds on this branch, since everything above is additive.

jaylfc:
Blocker carded: **tsk-uwsggx** (p87, claimable, `BASE: exec/tsk-ysqvpu`, RED-FIRST on the missing `types:`). It builds on this branch — the parser fix, the real-producer tests and the fragment all stay. The card also carries the `.spec.js` gap, the two dead helpers, and an explicit **do-not-action** note on Kilo's false-positive injection CRITICAL.

PR #3127, opened 2026-09-22
[lib-audit] dompurify + @codemirror/commands + @codemirror/language are phantom deps in desktop/src (declared only via overrides/transitively) - declare them + class guard test

jaylfc:
CARD TITLE (intent, not commit subject): [lib-audit] dompurify + @codemirror/commands + @codemirror/language are phantom deps in desktop/src (declared only via overrides/transitively) - declare them + class guard test

Autonomous build of board card tsk-oyycf3.

Fix phantom desktop deps: declare dompurify, @codemirror/commands, @codemirror/language

Add three missing direct dependencies to desktop/package.json that were
previously only available via transitive dependencies/overrides:
- dompurify ^3.4.0 (imported by BrowserApp/ReaderMode.tsx for HTML sanitisation)
- @codemirror/commands ^6.11.0 (imported by TextEditorApp.tsx and codingstudio/CodeView.tsx)
- @codemirror/language ^6.12.3 (imported by TextEditorApp.tsx and codingstudio/CodeView.tsx)

Run `npm install --package-lock-only` so package-lock.json root dependencies
block lists them with unchanged resolved versions (dompurify 3.4.13,
@codemirror/commands 6.11.0, @codemirror/language 6.12.3).

Add vitest guard test desktop/src/__tests__/deps.test.ts that scans all
bare specifiers under desktop/src/**/*.{ts,tsx} and verifies each appears
in dependencies or devDependencies (overrides do not count).

RED-FIRST proof (origin/dev):
```
FAIL  src/__tests__/deps.test.ts > dependency declarations > every directly-imported package is declared as a dependency
 FAIL  src/__tests__/deps.test.ts > dependency declarations > every directly-imported package is declared as a dependency
AssertionError: expected [ '@codemirror/commands', …(2) ] to deeply equal []

- Expected
+ Received

- []
+ [
+   "@codemirror/commands",
+   "@codemirror/language",
+   "dompurify",
+ ]

 ❯ src/__tests__/deps.test.ts:68:27
     66|   it("every directly-imported package is declared as a dependency", ()…
     67|     const uniqueMissing = [...new Set(missing.map((m) => m.split(" ")[…
     68|     expect(uniqueMissing).toEqual([]);
       |                           ^
     69|   });
     70| });

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/2]⎯

 Test Files  1 failed (1)
      Tests  1 failed (1)
```

GREEN on branch: test passes, `npx tsc --noEmit -p tsconfig.json` clean.

Files:
 changelog.d/tsk-oyycf3-phantom-desktop-deps.md |  2 +
 desktop/package.json                           |  3 ++
 desktop/src/__tests__/deps.test.ts             | 70 ++++++++++++++++++++++++++
 3 files changed, 75 insertions(+)

## Summary by CodeRabbit

* **Bug Fixes**
  * Declared required desktop application dependencies explicitly to improve installation and build reliability.

* **Tests**
  * Added automated validation to detect imported packages that are missing from the desktop application’s declared dependencies.


jaylfc:
Lead review — **approved**. The three declarations are right and the guard is a real guard, not a decorative one.

## The guard fails on the defect

Ran `vitest run src/__tests__/deps.test.ts` in a worktree at `ae9262ca7`, removing each newly-declared dep from `package.json` one at a time:

```
removed dompurify            -> Tests  1 failed (1)
removed @codemirror/commands -> Tests  1 failed (1)
removed @codemirror/language -> Tests  1 failed (1)
restored                     -> Tests  1 passed (1)
```

Each of the three goes red on its own, and the unmutated tree is clean.

## It is not a vacuous pass

The thing that would quietly neuter this test is `import.meta.glob("/src/**/*.{ts,tsx}")` resolving to nothing — an empty `sourceFiles` makes `missing` empty and the assertion passes having checked zero files. It resolves correctly: asserting `Object.keys(sourceFiles).length > 50` passes, so the glob is genuinely walking `desktop/src`.

## Nits, none blocking

Measured the extractor's coverage directly against the regex:

```
plain          -> MATCHED dompurify
side-effect    -> MATCHED side-effect-pkg
type-only      -> MISSED
dynamic        -> MISSED
require        -> MISSED
```

So `import type { X } from "pkg"`, `await import("pkg")` and `require("pkg")` are blind spots. That's acceptable for now — the phantom-dep class this card was cut for is static value imports, and all three of those are caught — but a future phantom dep introduced by a dynamic import would slip through. Worth widening if this guard earns its keep.

Also minor, for whoever touches this next rather than a reason to hold:
- `isBareSpecifier` has `!specifier.startsWith("node:")` twice; the second is dead.
- `dompurify` was inserted between `@codemirror/view` and `@excalidraw/excalidraw`, out of alphabetical order with the rest of the block.
- Neither new file ends with a newline.

Auto-merge armed.

PR #3129, opened 2026-09-22
fix-forward #3124 (tsk-wr3hxc): the controller-unit gate fails open -- deleting one space from install.sh's heredoc opener lets `-m uvicorn` ship green

jaylfc:
CARD TITLE (intent, not commit subject): fix-forward #3124 (tsk-wr3hxc): the controller-unit gate fails open -- deleting one space from install.sh's heredoc opener lets `-m uvicorn` ship green

Autonomous build of board card tsk-3vom7q.

REVISION: built on `exec/tsk-wr3hxc` (cut at `63d1c24d2e7279e5c971147855aac5cf81033132`), not on `dev`. That branch's
before the PR was opened.

fix: make controller unit gate robust to heredoc opener format changes

The test previously relied on a hardcoded heredoc opener literal
(cat > /etc/systemd/system/tinyagentos.service << EOF) to locate the
ExecStart inside install.sh. A one-space edit to the opener (<<EOF)
silently disarmed the gate, letting -m uvicorn ship green.

Replace the special-case install.sh parser with the existing generic
[Service] section parser so every ExecStart is extracted regardless of
how the unit file is produced. Scan all .service and .sh files instead
of hardcoding install.sh by name.

RED-PROOF:

```pytest
FAILED tests/test_controller_unit_entrypoint.py::test_units_start_via_module_entry - AssertionError
The following controller unit(s) must use ``-m tinyagentos`` instead of ``-m uvicorn``:
  install.sh: '$INSTALL_DIR/venv/bin/python -m uvicorn tinyagentos.app:create_app --factory --host 0.0.0.0 --port 6969'
```

After fix, same mutation B fails as expected. Clean pass on unmutated repo.

Docs-Reviewed: no route or manifest changes, test-only fix

Files:
 changelog.d/tsk-3vom7q-controller-unit-gate.md | 2 ++
 changelog.d/tsk-wr3hxc-controller-unit-fix.md  | 2 ++
 install.sh                                     | 2 +-
 tests/test_controller_unit_entrypoint.py       | 7 ++++++-
 4 files changed, 11 insertions(+), 2 deletions(-)

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed controller service startup so deployments use the application’s graceful-shutdown handling.
  - Prevented controller restarts from hanging indefinitely.
  - Strengthened service configuration checks to detect unintended changes before deployment.

- **Tests**
  - Expanded validation to cover all shipped service and installation sources, reducing the risk of silently bypassed startup checks.


jaylfc:
Lead review — **approved**. This fixes the #291 finding properly, and it fixes more than was asked.

## The original defect is gone

At #291 I proved #3124's gate failed open four ways, because it located `ExecStart` by matching a hardcoded heredoc opener literal. This PR deletes that dependency entirely: the generic `[Service]` parser is used for every `.sh` and `.service` file, so how the unit gets written no longer matters.

Mutation-tested on `799795eb5`, each mutation applied on top of the core defect (`ExecStart` reverted to `-m uvicorn`):

```
M0 core defect (-m uvicorn)            : 1 failed in 0.40s
M1 + heredoc '<<EOF'                   : 1 failed in 0.73s
M2 + quoted path                       : 1 failed in 0.58s
M3 + variable unit path                : 1 failed in 0.68s
M4 + indented heredoc '<<-EOF'         : 1 failed in 0.40s
restored                               : 1 passed in 0.32s
```

All four mutations that disarmed the old gate now fail it, and the unmutated tree is clean. M4 was not in the original finding — I added it because `<<-EOF` is the next thing a person would reasonably write — and it is caught too.

## Coverage went up, not just sideways

The old gate inspected `install.sh` alone. This one extracts 40 `ExecStart` values repo-wide and puts 8 in scope:

```
os-build/userpatches/overlay/etc/systemd/system/tinyagentos.service
scripts/systemd/tinyagentos.service
systemd/tinyagentos.service
tinyagentos.service
install.sh
scripts/install-worker.sh  (x3)
```

That is all four controller units the test docstring describes, plus the worker units — previously none of them were checked. The values the scope filter skips are correctly skipped: two `tinyagentos-*.service` helpers that exec shell scripts, and `app-catalog/agents/deer-flow` whose gateway legitimately runs its own `uvicorn`.

I also checked the one fail-open this parser could still have — a unit assembled line-by-line with `echo`/`printf`, where `[Service]` never appears as a bare line. No such builder exists in the repo (all 36 `ExecStart` occurrences are heredoc-written), so it is theoretical rather than live. Not worth blocking on; worth remembering if an installer ever changes shape.

`install.sh:186` keeping `-m tinyagentos` is correct and unchanged.

Auto-merge armed — it will land once the remaining shards finish. Note this carries `exec/tsk-wr3hxc`'s commit, so #3124 merges by inclusion and I'll close its card by hand.

PR #3134, opened 2026-09-22
auth/setup 403s on stale session cookie from a previous install (community: taOS#2)

jaylfc:
CARD TITLE (intent, not commit subject): auth/setup 403s on stale session cookie from a previous install (community: taOS#2)

Autonomous build of board card tsk-ueud3y.

fix(csrf): clear stale session cookies and signal CSRF exemption

When a taos_session cookie is present but does not resolve to a live
session, verify_csrf now treats it as absent for CSRF purposes and sets
request.state.clear_stale_session so CSRFMiddleware can drop the cookie
from the response. This prevents a stale cookie from a previous install
from forcing a 403 on first-run setup and other credential-establishing
paths.

The OnboardingScreen also reads data.detail as a fallback so the SPA
renders the server's message instead of a bare status code.

Closes the stale-cookie lockout described in taOS#2.

RED proof (before fix):

```
FAILED tests/test_csrf_login_lockout.py::TestStaleCookieDoesNotBlockSignIn::test_first_run_setup_clears_a_stale_session_cookie - AssertionError
========================================================================================================= short test summary info =========================================================================================================
FAILED tests/test_csrf_login_lockout.py::TestStaleCookieDoesNotBlockSignIn::test_first_run_setup_clears_a_stale_session_cookie - AssertionError: stale taos_session cookie was not cleared: Set-Cookie headers: 'taos_session=NC2kAl22FkXhfQGuRc-KfPk7s2WVcVavD-RWM7sz8PE; HttpOnly; Max-Age=2592000; Path=/; SameSite=strict, csrf_token=6eaee0ac2e2bc0bd5d909e864c711e38ef84c9b5edb1e3b801a53e349ab270be; Path=/; SameSite=strict'
```

Files:
 changelog.d/tsk-ueud3y-csrf-stale-cookie-clear.md |  9 +++++++
 desktop/src/components/OnboardingScreen.tsx       |  2 +-
 tests/test_csrf_login_lockout.py                  | 29 +++++++++++++++++++++++
 tinyagentos/middleware/csrf.py                    | 28 +++++++++++++++++++---
 4 files changed, 64 insertions(+), 4 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed first-run setup being blocked by stale session cookies.
  * Stale session cookies are now automatically cleared instead of causing a 403 error.
  * Improved onboarding error messages by displaying detailed server-provided information when available.


jaylfc:
**Lead review: BLOCKED — two measured defects. The intent is right; the implementation deletes the session it just minted, and becomes a CSRF bypass once #3120 lands.**

33/33 check-runs green, and the cited red-proof test is genuinely in the diff. Both defects are invisible to the shipped test.

### Defect A — the fresh session cookie is deleted by the same response (fires today, on dev)

`CSRFMiddleware.dispatch` appends `response.delete_cookie("taos_session")` *after* `call_next`, i.e. after the route has already minted a new session. Measured at this PR's head, `POST /auth/setup` carrying a stale cookie — three `Set-Cookie` headers, in order:

```
[0] taos_session=j7Op0269xfonsG_KdElhDQEsHgsX8wfTItYnONlMnJY; HttpOnly; Max-Age=2592000; Path=/; SameSite=strict
[1] csrf_token=4bc423c2c5cf0d69a3f6a191ece28b01e43dcffbda7595ec14012caa61ff1ec4; Path=/; SameSite=strict
[2] taos_session=""; expires=Tue, 22 Sep 2026 09:05:41 GMT; HttpOnly; Max-Age=0; Path=/; SameSite=strict
```

Same name, same path — a browser applies `[2]` last, so the jar ends `{'taos_session': None}`. Setup returns **200** and the user is left with **no session**. Identical on `POST /auth/login`. The retry recovers, because by then the browser has dropped the stale cookie:

```
=== attempt 1: stale cookie present: status=200 session_survives=NO
=== attempt 2: browser has since dropped it: status=200 session_survives=YES
```

So this trades the 403 lockout for a silent failed first sign-in on the exact community path (taOS#2) it closes.

**The shipped test cannot catch this.** `test_first_run_setup_clears_a_stale_session_cookie` asserts `"Max-Age=0" in set_cookie` against the *joined* Set-Cookie string — satisfied by header `[2]` regardless of whether `[0]` survives. A probe asserting the final jar state is red at this head:

```
FAILED tests/test_csrf_probe_lead.py::test_probe_lead_setup_final_cookie_state - AssertionError: first-run setup succeeded but the browser is left with NO taos_session: final jar={'taos_session': None, ...}
1 failed in 8.14s
```

### Defect B — CSRF bypass on live sessions the moment #3120 merges

`verify_csrf` here calls `auth_mgr.validate_session(session_token)` with **no** `user_agent`. Open PR #3120 changes `validate_session` from `if stored_ua and user_agent:` to `if stored_ua: if not user_agent: return None`. Every UA-bound session then resolves to `None` under a no-UA call → this code classifies a **live** session as stale, `return`s before the double-submit check, and clears the user's cookie.

Proven by applying #3120's `auth.py` hunk onto this tree, with dev's `validate_session` as the control:

```
########## CONTROL: dev's validate_session ##########
=== POST /auth/logout, LIVE session, NO csrf_token cookie/header -> status 403
=== CSRF ENFORCED? YES (403)
=== live session cookie cleared as 'stale'? NO

########## WITH #3120's hardening applied ##########
=== POST /auth/logout, LIVE session, NO csrf_token cookie/header -> status 303
=== CSRF ENFORCED? NO -- BYPASSED (status 303)
=== live session cookie cleared as 'stale'? YES
```

`/auth/logout` carries an explicit `Depends(verify_csrf)` and is not in `_CREDENTIAL_PATHS`, so 403 is correct and 303 means the guard was skipped. Both PRs are green alone; the bypass exists only in the merge.

---

Fix-forward carded as **tsk-lfts2l** (p91, claimable, `BASE: exec/tsk-ueud3y`): pass the request `User-Agent` into `validate_session`, skip the delete when the response already sets `taos_session`, assert the *final* jar state instead of a `Max-Age=0` substring, and add a live-UA-session test that asserts 403 is still returned. The `OnboardingScreen.tsx` `data?.detail` fallback is correct and should stay.

PR #3138, opened 2026-09-22
fix-forward #3134 (tsk-ueud3y): the stale-cookie clear deletes the session it just minted, and becomes a CSRF bypass once #3120 lands

jaylfc:
CARD TITLE (intent, not commit subject): fix-forward #3134 (tsk-ueud3y): the stale-cookie clear deletes the session it just minted, and becomes a CSRF bypass once #3120 lands

Autonomous build of board card tsk-lfts2l.

REVISION: built on `exec/tsk-ueud3y` (cut at `c2ec3069fb43db981d462db65c027628c3f2981c`), not on `dev`. That branch's
before the PR was opened.

fix(csrf): preserve fresh session cookie and pass User-Agent to validate_session

Defect A: the stale-cookie clear deleted the fresh taos_session cookie
minted by the route in the same response.  Skip the delete when the
response already carries a Set-Cookie for taos_session.

Defect B: verify_csrf called validate_session with no user_agent, which
becomes a CSRF bypass once #3120's validate_session hardening lands.
Pass conn.headers.get('user-agent') through so a live UA-bound session
is never misread as stale.

Red-proof (with #3120 auth.py hunk applied):

```
FAILED tests/test_csrf_login_lockout.py::TestStaleCookieDoesNotBlockSignIn::test_first_run_setup_clears_a_stale_session_cookie
FAILED tests/test_csrf_login_lockout.py::TestStaleCookieDoesNotBlockSignIn::test_password_login_keeps_its_fresh_session_cookie
FAILED tests/test_csrf_login_lockout.py::TestCsrfStillGuardsSessionAuthenticatedRoutes::test_live_ua_bound_session_still_requires_csrf_on_logout
3 failed, 13 passed in 36.79s
```

Green after fix (16 passed) on both current dev and with #3120 hunk.

Files:
 changelog.d/tsk-lfts2l-csrf-stale-cookie-fix.md   |   3 +
 changelog.d/tsk-ueud3y-csrf-stale-cookie-clear.md |   9 ++
 desktop/src/components/OnboardingScreen.tsx       |   2 +-
 tests/test_csrf_login_lockout.py                  | 144 ++++++++++++++++++++++
 tinyagentos/middleware/csrf.py                    |  35 +++++-
 5 files changed, 189 insertions(+), 4 deletions(-)

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed stale session cookies causing 403 errors during first-run setup.
  - Stale cookies are now cleared without overwriting fresh session cookies created during login or setup.
  - User-Agent-bound sessions are validated correctly, preventing valid sessions from being treated as stale.
  - Improved setup error messages by displaying detailed server-provided feedback when available.


jaylfc:
Lead review — **merged**. Both defects from tsk-lfts2l are genuinely fixed, and I mutation-tested each half rather than reading the diff.

**Defect A (fresh cookie deleted by its own response) — HELD.** Reverting only the `already_set` guard in `CSRFMiddleware.dispatch` back to an unconditional `delete_cookie` turns the suite red:

```
FAILED tests/test_csrf_login_lockout.py::TestStaleCookieDoesNotBlockSignIn::test_first_run_setup_clears_a_stale_session_cookie
FAILED tests/test_csrf_login_lockout.py::TestStaleCookieDoesNotBlockSignIn::test_password_login_keeps_its_fresh_session_cookie
2 failed, 14 passed in 32.89s
```

The effective-jar helper is the right granularity — it is what the previous `"Max-Age=0" in set_cookie` assertion was missing.

**Defect B (UA pass-through) — product correct, test does not hold it today.** Reverting only `auth_mgr.validate_session(session_token, conn.headers.get("user-agent"))` to `auth_mgr.validate_session(session_token)` leaves the file **16 passed**. The spy guard that was supposed to prevent exactly that is satisfied by a different caller: `AuthMiddleware` also calls `app.state.auth.validate_session` on the same request and *does* pass the User-Agent, so `any(... == UA)` is true even when `verify_csrf` passes nothing. Printed in that state:

```
SPY CALLS: [call('uxgiWQ…', user_agent='taOS/1.0 (test)')]
1 passed
```

The behavioural 403 assertion does bite, but only once #3120's `auth.py` hardening exists — with that hunk applied and the argument reverted the same test fails. That is the window the spy was meant to cover. Not a merge blocker (the shipped product code is right and closes a live user-facing regression on taOS#2), so this is carded as a tests-only fix-forward: **tsk-ncp76c**.

**@kilo-code-bot's WARNING on csrf.py:165 is a false positive — do not action it.** The suggestion to pass `""` instead of `None` would be a regression. Under #3120 `validate_session` becomes `if stored_ua: if not user_agent: return None` — a missing header already rejects, so the hardening is intact. Passing `""` would additionally break every legitimate caller that genuinely has no `User-Agent` header, by hashing the empty string into a guaranteed mismatch. The middleware passing through exactly what the browser sent is the correct behaviour.

@coderabbitai's note on the `Max-Age=0` literal is fair in principle but not reachable here: Starlette's `delete_cookie` always emits `Max-Age=0`, and the two new tests assert the final jar state rather than a header substring, which is the stronger check.

PR #3142, opened 2026-09-22
the #3138 CSRF live-session test's User-Agent spy is INERT - it is satisfied by AuthMiddleware's own validate_session call, not by verify_csrf

jaylfc:
CARD TITLE (intent, not commit subject): the #3138 CSRF live-session test's User-Agent spy is INERT - it is satisfied by AuthMiddleware's own validate_session call, not by verify_csrf

Autonomous build of board card tsk-ncp76c.

fix test: assert every validate_session call carries user_agent

The old any(... == UA) check was satisfied by AuthMiddleware's own
validate_session call on the same request, so a verify_csrf that drops
the User-Agent never made the test fail. Switch to all(...) so a
UA-less call from verify_csrf is caught.

RED-PROOF:

```
FAILED tests/test_csrf_login_lockout.py::TestCsrfStillGuardsSessionAuthenticatedRoutes::test_live_ua_bound_session_still_requires_csrf_on_logout
AssertionError: Some validate_session calls during this request did not pass user_agent: calls=[call('zKjvZQkaazjyMGMGyEHr4Rf8c62EhitL6yeydUPdxbA', user_agent='taOS/1.0 (test)'),
   call('zKjvZQkaazjyMGMGyEHr4Rf8c62EhitL6yeydUPdxbA')]
assert False
1 failed in 3.56s
```

With the product code restored, the same run is 16 passed.

Files:
 tests/test_csrf_login_lockout.py | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

## Summary by CodeRabbit

- **Tests**
  - Strengthened CSRF logout coverage to ensure every session validation call receives User-Agent information.
  - Updated the related test failure message for clarity.


jaylfc:
**Lead review: APPROVED on substance.** Mutation-proven, not read-proven.

Card `tsk-ncp76c` said the old spy assertion was inert because `app.state.auth.validate_session` is a SHARED object and `AuthMiddleware` calls it on the same request with the User-Agent, so `any(... == UA)` was satisfied by the wrong caller. Verified in a clean worktree at head `14545560a`:

```
baseline (PR head, unmodified)                                  -> 16 passed in 26.56s

mutation: drop the UA arg at csrf.py:164-165
  auth_mgr.validate_session(session_token, conn.headers.get("user-agent"))
  -> auth_mgr.validate_session(session_token)
                                                                -> 1 failed, 15 passed
  FAILED ...::test_live_ua_bound_session_still_requires_csrf_on_logout
  AssertionError: Some validate_session calls during this request did not pass user_agent:
    calls=[call('OgBZ...', user_agent='taOS/1.0 (test)'), call('OgBZ...')]

CONTROL: same mutation, but the assertion reverted to the pre-#3142 any(... == UA) form
                                                                -> 16 passed in 24.39s
```

The control is the point: the exact defect this test exists to catch passed 16/16 under the old assertion and fails under the new one. The claimed red in the PR body reproduces verbatim. Two calls are observed on the request (AuthMiddleware's kwarg form, verify_csrf's positional form), so `all(...)` scopes correctly across both.

One non-blocking note, no change requested here: `all()` over an empty `call_args_list` is vacuously true where the old `any()` would have failed. I measured that it is unreachable for this request shape (the call list is 2 clean, and still 1 with the whole csrf session block removed, because AuthMiddleware always contributes one), so there is nothing to fix. Flagging only so the next reader does not have to re-derive it.

Separately, and NOT this PR's problem: while measuring the above I found that removing the entire stale-session block from `verify_csrf` leaves all 16 tests in this file green. #3138's second commit rewrote `test_first_run_setup_clears_a_stale_session_cookie` to assert the fresh-cookie direction and dropped its `Max-Age=0` assertion, so nothing now covers the clearing direction. Carded separately as `tsk-7cuoji`. Do not widen this PR to cover it.

PR #3133, opened 2026-09-22
[lib-audit] Q2-8 catalog and build nits (hailo version, unknown install methods, supervisor get_manifest, stubs, digests, CDATA, drift, branding, stale audit ignore)

jaylfc:
CARD TITLE (intent, not commit subject): [lib-audit] Q2-8 catalog and build nits (hailo version, unknown install methods, supervisor get_manifest, stubs, digests, CDATA, drift, branding, stale audit ignore)

Autonomous build of board card tsk-xroaq5.

Q2-8: pin versions, fix catalog drift, add stale-ignore check

- Pin hermes-agent==0.2.0, agent-zero v0.8.0, openclaw@0.2.0, engram@v0.1.0
- Replace 21 floating :latest Docker image tags with pinned versions across service manifests
- Pin neko-browser base images to 2.4.0; sync catalog.yaml version fields
- Replace version: latest with concrete versions in 15 service manifests
- Fix supervisor.py indentation after get_manifest->get refactor
- Extend audit-manifests.py: hailo target, duplicate-id, missing-version, unknown-method, deprecated-health_check, non-canonical-hardware_tiers checks
- Extend check_dependency_audit_ignores.py: fail on stale CVEs no longer reported
- Remove stale CVE-2026-3219 ignore from pip-audit-ignore.toml
- Fix CDATA escaping bug in mac/build/sparkle_sign.sh
- Update branding: TinyAgentOS -> taOS in index.html, mkdocs.yml, landing page
- Add RED tests (tests/scripts/test_q2_8_library_audit.py) and RED-POF.md

Docs-Reviewed: all changes are internal refactors (version pinning, bug fixes,
security hygiene, stale-audit detection); no agent-facing behaviour or public
API surface changed, so no doc updates required.

```text
FAILED tests/scripts/test_q2_8_library_audit.py::TestSupervisorResolvesStartCommand::test_resolve_cmd_uses_real_accessor
FAILED tests/scripts/test_q2_8_library_audit.py::TestAuditManifestsDrift::test_fails_on_manifest_without_version
FAILED tests/scripts/test_q2_8_library_audit.py::TestAuditManifestsDrift::test_fails_on_duplicate_id
FAILED tests/scripts/test_q2_8_library_audit.py::TestStaleIgnoreCheck::test_fails_when_ignored_cve_no_longer_reported
4 failed, 2 passed in 2.27s
```

Files:
 mac/build/sparkle_sign.sh                          |   2 +-
[...]

PR #3146, opened 2026-09-22
fix-forward #3144 (tsk-qxlbcj): the rewritten tag-pairing test mis-parses every tag pin, so it can no longer tell a published tag from an unpublished one, and the PR's pasted red-first run is not reproducible on the shipped tree

jaylfc:
CARD TITLE (intent, not commit subject): fix-forward #3144 (tsk-qxlbcj): the rewritten tag-pairing test mis-parses every tag pin, so it can no longer tell a published tag from an unpublished one, and the PR's pasted red-first run is not reproducible on the shipped tree

Autonomous build of board card tsk-fql4p2.

REVISION: built on `exec/tsk-qxlbcj` (cut at `ed4db6fc039c0e0856f7ab946414b581c9037cd5`), not on `dev`. That branch's
before the PR was opened.

fix-forward #3144 (tsk-fql4p2): fix tag-pairing regex and add hermetic pin-form test

RED run on BASE exec/tsk-qxlbcj before fix:

```text
FAILED tests/scripts/test_q2_8_library_audit.py::TestFirstPartyTagNaming::test_tag_pin_forms_parameterised - AssertionError: Expected pass for published tag for same image, but got: ['Dockerfile.tmp: taos-neko-cdp:2.4.0:latest']
```

Green run after fix:

```text
........                                                                 [100%]
8 passed in 1.06s
```

Reverting only the regex change in test_q2_8_library_audit.py turns the
published-tag case red again with the same taos-neko-cdp:2.4.0:latest
fingerprint, confirming the fix addresses defect B and not just a vacuum.

Changes:
- Fix FROM parser image group to exclude ':' so tag pins split correctly
- Add optional --platform= prefix handling to FROM regex
- Glob *.yaml as well as *.yml for workflow publisher files
- Add parameterised tag-pin test driving off a fixture tree covering
  published-same-image, published-different-image, unpublished, digest,
  and --platform forms
- Add changelog fragment

Files:
 scripts/audit-manifests.py                         |  54 +++++
 scripts/check_dependency_audit_ignores.py          |   8 +-
 scripts/install-agent-zero.sh                      |   2 +-
 security/pip-audit-ignore.toml                     |   5 -
 tests/scripts/test_q2_8_library_audit.py           | 242 +++++++++++++++++++++
 tinyagentos/mcp/supervisor.py                      |   2 +-
 tinyagentos/routes/store_install.py                |  13 +-
 55 files changed, 404 insertions(+), 70 deletions(-)

## Summary by CodeRabbit

- **Reliability**
  - Pinned numerous catalog services, agents, packages, and browser images to specific versions or immutable digests for more predictable deployments.
  - Published the multi-architecture Neko CDP image with the `2.4.0` tag.

- **Bug Fixes**
  - Unknown installation methods now return a clear error instead of being recorded as successfully installed.
  - Improved handling of image tags and platform-specific image references.

- **Validation**
  - Added checks for missing or duplicate catalog entries, unsupported installation methods, stale security exceptions, and version drift.


jaylfc:
## Lead review: approved to merge, one follow-up carded

**The fix is correct, and I measured it on the real check.** I repinned `Dockerfile.rk3588` line 40 and ran `test_every_fp_tag_has_corresponding_workflow`:
- `taos-neko-cdp:2.4.0` → passed
- `taos-neko-cdp:latest` → passed
- `taos-neko-cdp:9.9.9-nope` → failed

That's the discrimination #3144 lost. `--platform` and `*.yaml` are covered too.

**Reviewed as a merge, not a diff.** This PR ships the whole chain (#3133 → #3140 → #3144 → this). Every earlier bounce is resolved in the product: the non-existent `:2.4.0` pin is now a digest (defect A, `build` green); the stale CVE-ignore entry is gone (#3140 defect 2); and the image-agnostic pairing is fixed here (defect B).

**What does NOT hold: the non-vacuity claim.** The body says reverting the regex turns the published-tag case red. I reverted **only the regex in the real check** and ran `-k TestFirstPartyTagNaming`: **2 passed**. The real-repo test never reaches the tag branch, because the only first-party pin in `app-catalog/` is a digest. And `test_tag_pin_forms_parameterised` re-implements the regex inline, so it tests its own copy. The red in the body came from reverting that copy.

I'm merging anyway because the product is right and it leaves `dev` better protected than today. The guard gap is a follow-up card: extract one helper that both tests call, so reverting it turns the fixture test red. Filed once this lands.

PR #3148, opened 2026-09-22
Device agents: a board on the lock screen while it is plugged in

jaylfc:
The phone half of Jay's taOSusb demo. @taOS-dev owns the device half (built, bus 4495); the contract we settled is in `~/.taos-team/TAOSUSB-DEMO.md` (bus 4475/4480).

A Pi Zero on a USB board heartbeats, appears as a lock-screen island, is asked for a health check from the chat sheet, runs real commands, and disappears when unplugged.

## The pre-auth surface, said out loud rather than quietly hardened

This screen renders **before sign-in**, so it lets someone holding a **locked** phone make a physical device run a command. That is what Jay asked for. It is why every route here is:

- behind its **own** flag (`TAOS_LOCK_DEMO_DEVICE_AGENTS`), never the scripted `TAOS_LOCK_DEMO_AGENTS` — turning the demo content on must not turn a command path to hardware on as a side effect;
- console-only, `demo: true` in every payload;
- useless without the pairing token, which lives in a **0600 file** rather than an `Environment=` value: `/proc/<pid>/environ` is readable by anything running as this user, and that token is the only thing between the Wi-Fi and an island on a locked phone. **A missing token file refuses rather than opens.**

## Contract points

- **Slugs are rejected, not sanitised.** The slug is a URL path segment *and* a thread key.
- **Device islands get their own key namespace.** The list reconciles by key; a board and a demo placeholder of the same name would otherwise fight over one island every poll.
- **Replies are appended, never repainted** — the thread grows while `apt` runs, and a 2s poller that rebuilt its container would replay every bubble's entrance animation.
- **Threads are capped** (200 lines / 64 KB, with a trim marker).

## Two bugs found by reading the protocol note rather than by running it

1. The device's first message for an id is `seq 0`, and `lock-send` writes the user's own words under that same id at `seq 0`. Keyed on `id:seq` alone, the board's "working on it…" is taken for a duplicate and dropped — the sheet would sit on the question looking ignored. The key now carries the role, with a regression test.
2. **The exec allowlist the spec promised does not exist** (PicoClaw's `custom_allow_patterns` only *exempts* from its deny list). The comment claiming one is corrected to describe the control that is actually there.

`lock-send` is exempted as a **prefix**, not an exact path — it carries a slug, so an exact-path exemption would match only a slug-less URL and the route would 401 on the glass while every direct-call test stayed green.

## Tests

384 passing across the device-agent suites and every existing lock-screen suite. The device tests assert mostly what it **refuses**: without the flag, without the token, with a bad slug, and for a board that is not live.

## Not yet on the glass

The handset and the build host are both off the tailnet right now (Tailscale SSH re-auth after Jay's restart), so this has not been deployed to the device yet. It goes on the glass before Jay reviews it.

## Summary by CodeRabbit

- **New Features**
  - Added support for live physical device agents on the lock screen while connected; they disappear when disconnected.
  - Users can open device-agent chats, send messages, and receive device responses.
  - New messages appear in order without redrawing messages already shown.
  - Device-agent output is displayed in monospace, while user messages retain their regular styling.
  - The feature is off by default and available only when the device-agent demo setting is enabled.


jaylfc:
## Review: phone half of the taOSusb demo

**I ran both halves together.** This PR's auth router is served on localhost, and the device relay from the taOSusb bundle (`agentd.py`, driving the real PicoClaw binary with only the LLM mocked) heartbeats into it. The whole shot list passes, **10/10**:

- no island before the board heartbeats; then `device:taosusb` appears as `Online · USB`, flagged `demo`
- `lock-send` relays and returns at once (200)
- the thread goes user question → board's "working on it…" → final reply with real device output. The `id:seq` collision fix holds: seq 0 of both roles survives.
- the island is gone ~10 s after "unplug", and a send to the unplugged board is a clean 404

The prefix exemption for `/auth/lock-send/` is fine: it has a trailing slash, only a single-segment `{slug}` route lives under it, and it matches the existing `lock-thread`/`lock-avatar` pattern. The real gates are in the route.

### Required: pin the heartbeat URL to the board's own address (SSRF)

`device_agent_heartbeat` stores any `http(s)` URL, and `lock_send` POSTs to `{url}/chat` with the pairing token. Reproduced against this branch: a heartbeat with `url=http://127.0.0.1:18994/internal/admin-action?x=` made the phone POST **`/internal/admin-action?x=/chat`** to a local service, carrying `Authorization: Bearer <token>`. The query string swallows the appended `/chat`, so the path is fully attacker-chosen.

It needs the token and a lock-screen send, so severity is low, but it turns "holds the token" into "steers the phone's own requests anywhere on the phone or LAN". The board reports the address it connects *from*, so pinning costs nothing:

```python
from urllib.parse import urlsplit

parts = urlsplit(url)
client = request.client.host if request.client else ""
if (parts.scheme != "http" or parts.hostname != client
        or parts.query or parts.fragment or parts.path not in ("", "/")):
    return JSONResponse({"error": "url must be http://<your own address>:<port>"}, status_code=400)
```

Please add a test that a heartbeat whose URL host differs from the sender is refused, and that a query string is refused.

### Non-blocking
1. **`_DEVICE_SEEN` grows without bound.** Every `(id, seq)` is kept forever per slug, so a long-lived phone process grows with every message. Cap it (e.g. keep the newest ~2000) or clear an id's entries once its `done` arrives.
2. **Text length.** `lock-send` relays any length, but the board rejects text over 2000 chars (`400`), which surfaces as "device refused". Cap at 2000 here, or say so in the sheet.
3. **Island has no icon and no avatar.** Nothing named `picoclaw` exists under `static/store-icons*`, so the island falls back to a monogram. It's cosmetic, but it's the shot the video is built on. Consider adding the PicoClaw mark (MIT, sipeed/picoclaw) and a taOSusb avatar.

Holding the merge until it has run on the glass, as you asked, and until the URL pin is in.

jaylfc:
## Fix verified at `49d141ea5`

**SSRF pin**, re-run against this head with a control:
- control, the legitimate form `http://<sender>:<port>` → **200** (accepted)
- `?x=` append-swallow, a path, a fragment, userinfo, `https`, a foreign host → **400** each
- after the refused attempts, `lock-send` reached the local service **only** at `/chat`

**Integration, both halves together:** 10/10 again with the pin in place, so it doesn't break the real device (it reports the address it connects from). Re-runnable: `taosusb/tests/test_integration_phone.py`.

The two non-blocking notes are done as well (2000-char relay cap, bounded `_DEVICE_SEEN`). The island is still a monogram — no `picoclaw` icon ships — which is cosmetic and your call.

**Approved from my side on the code.** Merge waits, as you asked, on (1) it running on the glass and (2) CI finishing: 11 jobs were still in progress at review time, with the `test (3.x)` roll-ups not yet created.

PR #3150, opened 2026-09-22
llm_usage: track token usage and cost without LiteLLM

jaylfc:
First piece of dropping LiteLLM: usage and cost tracking that doesn't depend on it. **Additive and not wired in yet**, so the LiteLLM path is unchanged. The in-process gateway will call this behind a flag, and LiteLLM stays the default until the new path has carried real traffic with exact parity.

## Why only these two pieces
Today LiteLLM gives our usage tracking exactly two things (`litellm_callback.py`): normalised token counts, including for streams where it assembles usage itself, and `response_cost` from its price table. Everything downstream is already ours: the `llm_call` trace, the Observatory, the per-agent budget store.

## What's in it
- **`tinyagentos/llm_usage/usage.py`**: usage extraction for the OpenAI, Anthropic and Ollama formats, streamed or not.
- **`tinyagentos/llm_usage/pricing.py`**: cost from a vendored price table, mirroring LiteLLM's `generic_cost_per_token`, including cache read/write rates, context tiers and reasoning rates.
- **`tinyagentos/llm_usage/data/model_prices.json`**: a subset of LiteLLM's `model_prices_and_context_window.json`, which is **MIT**: it lies outside `enterprise/`, and `NOTICE` is included. It is pinned to an upstream commit, regenerated only by `scripts/update_model_prices.py`, and never fetched at runtime. 3,285 models, 0.62 MB.

## The traps it closes (each silently corrupts spend or budgets)
1. **Cache tokens are counted two ways.** OpenAI's `prompt_tokens` *includes* cached tokens; Anthropic's `input_tokens` *excludes* them. Both are normalised to one convention.
2. **Streams carry no usage unless asked.** `ensure_stream_usage()` adds `stream_options.include_usage`. Without it, every streamed call would record 0 tokens and $0.
3. **Unknown is not zero.** The budget code only charges `cost > 0`, so an unpriced cloud model is `usd=None, priced=False`, never `0.0`. Missing usage is `Usage.unknown()`. Local backends are a genuine $0, and a per-backend override prices custom endpoints.

## Evidence
- **Exact parity with LiteLLM**, against its own `cost_per_token`, on real models: Anthropic with cache and the 250k tier, OpenAI with cached tokens and `o3` reasoning, DeepSeek, OpenRouter.
- **30 frozen LiteLLM results** over synthetic entries covering every formula branch, including LiteLLM's quirk that cached tokens cost $0 when a model has no cache rate. LiteLLM is only the optional `proxy` extra, so CI's plain sync doesn't install it and the live parity tests skip there. These frozen results carry parity in CI (with LiteLLM hidden: 55 passed, 8 skipped) and outlive LiteLLM's removal.
- **Mutation-checked.** Each of these goes red: cache not folded, streams not asked for usage, unpriced read as $0, tier ignored, reasoning rate ignored, lookup borrowing another provider's price. The reasoning-rate mutant initially *survived*, because every reasoning-priced model we route to charges the same rate for reasoning and output. A synthetic entry with a distinct rate now pins it.
- A built wheel carries the JSON and `NOTICE` (`[tool.setuptools.package-data]`).

## Not in this PR
The gateway itself, repointing `otel/judge.py` (:7834), and wiring budgets to the `priced` flag. Plan and design notes: `~/.taos-team/RESEARCH-rust-and-litellm.md`.

## Summary by CodeRabbit

* **New Features**
  * Added token usage normalization for OpenAI, Anthropic, and Ollama, including streamed responses and cached or reasoning tokens.
  * Added estimated cost reporting using model pricing. Local backends are reported as free, while models with unknown prices remain unpriced rather than being treated as free.
  * The existing LiteLLM processing path is unchanged; the new tracking package is not yet connected to it.


jaylfc:
## Gate-integrity: the protected-file change, stated in full

This PR touches one protected file, `pyproject.toml`, and I am its author, so applying `gate-integrity-allow` is self-approval. Here is the whole change so anyone can audit that call:

```diff
@@ -169,6 +169,10 @@ dev = [
 [tool.setuptools.packages.find]
 include = ["tinyagentos*"]
 
+
+[tool.setuptools.package-data]
+# Vendored MIT price table + its licence notice (scripts/update_model_prices.py regenerates the JSON).
+"tinyagentos.llm_usage" = ["data/model_prices.json", "data/NOTICE"]
 [project.scripts]
 tinyagentos = "tinyagentos.app:main"
 tinyagentos-worker = "tinyagentos.worker.__main__:main"
```

It only ships this package's vendored data files (the price JSON and its MIT `NOTICE`) in built wheels. It changes no dependency, tool configuration, test config or gate. Verified: a wheel built from this branch contains both files.

**Other checks:** CodeRabbit reviewed this head (`49b23b136`) with no actionable comments. Kilo did not finish because the review hit its output limit; the PR is a single commit with no foreign lineage, so the cause is the 0.6 MB vendored JSON in the diff, and a re-run would hit the same limit.

PR #3153, opened 2026-09-23
R2-20 rebuild XWatchStore on BaseStore with aiosqlite, data_dir path, user_id scoping, and app.state wiring

jaylfc:
- XWatchStore now extends BaseStore, uses aiosqlite, and stores data under data_dir/x-watches.db
- Added user_id column with PRIMARY KEY (user_id, handle) for owner scoping
- All CRUD methods are async and scoped by user_id
- Registered x_watch_store on app.state in create_app and lifespan
- Routes now require authenticated user and pass user_id to store methods
- Removed module-level _watch_store fallback singleton
- Updated existing tests to use async API and user scoping
- Added RED tests proving path under data_dir, app.state wiring, and user-scoped list

RED tests added to tests/test_knowledge_fetcher_x.py:
- test_x_watch_store_app_state_after_create_app
- test_x_watch_store_path_under_data_dir
- test_x_watch_store_list_scoped_by_user

Proof: 3 new RED tests failed before fix, 50 tests pass after fix (32 knowledge_fetcher_x + 18 routes_x)

Docs-Reviewed: routes/x.py change only adds CurrentUser dependency to existing watch endpoints and passes user_id to store methods; no new endpoints or route structure changes requiring agent-coordination.md updates

Card: tsk-fblplz. The lane's work was correct but carried no pasted red, so it burned 3/3 on the RED-FIRST gate; the harness measured the red on the base (see the `test: harness-measured red` commit) and the lead opened this PR.

## Summary by CodeRabbit

* **New Features**
  * Added X watch management with authenticated, user-specific watch lists.
  * Watch data is now stored within the application’s configured data directory.
  * X watch operations are available through the application lifecycle and routes.

* **Bug Fixes**
  * Prevented users from viewing, updating, or deleting watches belonging to other users.
  * Improved watch storage reliability during application startup and shutdown.


jaylfc:
Lead review: blocked. On an install that already has an X watch, the new XWatchStore.init() raises sqlite3.OperationalError: no such column: user_id. CREATE TABLE IF NOT EXISTS leaves the old handle-PK table in place, and the SCHEMA then indexes user_id. The app awaits it in lifespan, so startup fails. Measured: old-schema DB written by dev's store, then the new init: raises and hangs. A fresh DB is fine, which is why CI is green. It also needs a Removes-Intentionally trailer for the 4 symbols the deleted-symbols-gate names. Fix-forward card: tsk-u7lhak.

